PR #51, opened 2015-06-03
in RotateTool, make angle a multiple of _snap_angle when _snap_rotation is enabled

ad1217:
Fixes a problem where rotation would not snap to the right angles, and was therefore not terribly useful


nallath:
Nice find. Thanks for the contribution!

awhiemstra:
Thank you! It was on my list of things to do but just did not get around to it yet.

PR #53, opened 2015-06-04
Fix: when PYQT_VERSION_STR does not contain patch, the unpack will fail.

derekhe:
Use slice to get the only major and minor version.


awhiemstra:
Thank you. Didn't know it was possible for it to only contain major and minor.

PR #65, opened 2015-09-01
Fix translate tool moving objects with an offset

fieldOfView:
The translate tool moves objects with an offset on subsequent operations.

Steps to reproduce before fix:
* Load object
* Use translate tool to move object around (either one of the axes, or the object itself)
-> behaves as expected
* Use translate tool again
-> jumps to an offset position, moves correctly with object

### DIFF
--- a/plugins/Tools/TranslateTool/TranslateTool.py
+++ b/plugins/Tools/TranslateTool/TranslateTool.py
@@ -63,6 +63,8 @@ class TranslateTool(Tool):
             else:
                 self.setDragPlane(Plane(Vector(0, 1, 0), 0))
 
+            self.setDragStart(event.x, event.y)
+
         if event.type == Event.MouseMoveEvent:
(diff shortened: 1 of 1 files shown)


nallath:
@awhiemstra I believed you said this bug was caused by a fix you made. Does this break anything of your fix?

awhiemstra:
It basically reverts the change.

fieldOfView:
@awhiemstra: How about this fix? The reason https://github.com/fieldOfView/Uranium/commit/cd1b061070c4933594f941e08a3ac83cf4786211 does not work is dragStart does not get reset on mouseup. This new commit fixes that (instead of reverting your previous commit)

awhiemstra:
@fieldOfView I was considering merging it anyway since the current issue is far worse than what I tried to solve, but if this fixes it too I'm happy. :)

nallath:
Works like a charm for me.

PR #66, opened 2015-09-02
Store window position and maximized state

fieldOfView:
Currently Uranium only stores the window size. When the window is maximized, the size is also stored as is. This makes the "restore size" function of the window useless (because both the maximized and restored window size are the same). Also on Windows the window is actually a couple of pixels wider than fullscreen because the window has a border.

This PR only stores the window-size when the window is not maximized. Additionally it stores the window position and the window state.

The PR contains a workaround for a Qt bug in Windows, where the windowState is not updated until after the moveEvent and resizeEvent have fired:
https://bugreports.qt.io/browse/QTBUG-30085


fieldOfView:
See https://github.com/Ultimaker/Cura/issues/188

PR #67, opened 2015-09-02
Fix crash on exit when no machine is selected

fieldOfView:
Steps to reproduce:
* delete cura profile
* start cura
=> machine wizard pops up
* close machine wizard
=> no machine is added
* close cura
=> crash because _active_machine is None


awhiemstra:
While it is a good catch, ideally the machine wizard should not be closeable in the first place.

PR #70, opened 2015-09-22
Make sure window is not positioned off-screen

fieldOfView:
If the window is positioned on an secondary screen when Cura is closed, and Cura is subsequently opened without the secondary screen being available, the window is drawn fully off-screen. This PR checks if (part of) the window is on the currently available screen(s), and moves the window back to the primary screen.


awhiemstra:
Look fine, thank you.

PR #73, opened 2015-10-27
Fix scale snapping

ad1217:
Fixes the first issue described in Ultimaker/Cura#493, in which scaling would not snap to 5% increments correctly.


TMHogenhout:
@ad1217 We were actually working on the same thing at the same time. So I manually merged our commits. I love what you did by the way, especially your fix for the ScaleOperation. Thanks again for all the great work!!

PR #74, opened 2015-10-29
Disable dragging the advanced settings pane horizontally

fieldOfView:
The advanced settings pane can be dragged horizontally, which is weird.

Steps to reproduce:
* Switch to advanced settings
* Expand a couple of categories
* Drag the scrollview horizontally

Expected result:
Dragging horizontally has doesn't


awhiemstra:
Hmm delicious platform differences. Seems Windows has some different defaults than Linux. In any case, merged and thank you! :)

PR #75, opened 2015-11-10
Check stored window position against all available screens

fieldOfView:
In 37fe8028a57e6a3090da59ef7de9881c543405de I tried to make sure the stored position of the window fits within the currently available screen geometry by looking at self.screen().availableGeometry(). 

It turns out that - at least under Windows - self.screen() does not get properly populated until after the window is shown. This update to the patch iterates through all available screens instead of relying on self.screen().

### DIFF
--- a/UM/Qt/Bindings/MainWindow.py
+++ b/UM/Qt/Bindings/MainWindow.py
@@ -44,7 +44,12 @@ class MainWindow(QQuickWindow):
         self.setHeight(int(self._preferences.getValue("general/window_height")))
         self.setPosition(int(self._preferences.getValue("general/window_left")), int(self._preferences.getValue("general/window_top")))
         # Make sure restored geometry is not outside the currently available screens
-        if not self.geometry().intersects(self.screen().availableGeometry()):
+        screenFound = False
+        for screenNr in range(0, self._app.desktop().screenCount()):
+            if self.geometry().intersects(self._app.desktop().availableGeometry(screenNr)):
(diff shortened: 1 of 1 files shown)


UM/Qt/Bindings/MainWindow.py
```diff
@@ -44,7 +44,12 @@ def __init__(self, parent = None):
         self.setHeight(int(self._preferences.getValue("general/window_height")))
         self.setPosition(int(self._preferences.getValue("general/window_left")), int(self._preferences.getValue("general/window_top")))
         # Make sure restored geometry is not outside the currently available screens
-        if not self.geometry().intersects(self.screen().availableGeometry()):
+        screenFound = False
```
awhiemstra: Minor nitpicking, but our coding style uses no caps with underscores for local variables, so you should use screen_found and similar.

fieldOfView:
Thanks, I'll update the PR. Is there a coding style document somewhere?

awhiemstra:
Not strictly a document, but the script in https://github.com/Ultimaker/Meta/blob/master/pep8_check_python.py checks most of the rules we established.

fieldOfView:
Did you see the update?

nallath:
Seems like we missed this pull request. Should we close this?

fieldOfView:
AFAIK the issue is still there. I'll make a JIRA issue and will see if the PR still applies.

fieldOfView:
Merged manually

PR #83, opened 2015-12-16
Write and Read Profile in plug-ins

Ghostkeeper:
This feature introduces plug-in readers and writers.

It also creates two new types of plug-ins: Profile Reader and Profile Writer, which is desired for a new feature in Cura to write and load settings in g-code.

A similar pull request is made for Cura. They should be tested together.

### DIFF
--- a/UM/Qt/Bindings/ProfilesModel.py
+++ b/UM/Qt/Bindings/ProfilesModel.py
@@ -118,6 +118,7 @@ class ProfilesModel(ListModel):
             return
 
         #Parse the fileType to deduce what plugin can save the file format.
+        #TODO: This parsing can be made unnecessary by storing for each plugin what the fileType string is in complete (in addition to the {(description,extension)} dict).
         #fileType has the format "<description> (*.<extension>)"
         split = fileType.rfind(" (*.") #Find where the description ends and the extension starts.
         if split < 0: #Not found. Invalid format.
--- a/UM/Qt/Bindings/ProfilesModel.py
+++ b/UM/Qt/Bindings/ProfilesModel.py
@@ -136,7 +136,20 @@ class ProfilesModel(ListModel):
     @pyqtSlot(result = "QVariantList")
     def getFileNameFiltersRead(self):
         filters = []
-        for extension,description in self._manager.getProfileSupportedFileTypesRead().items(): #Format the file types that can be read correctly for the open file dialog.
+        for extension, description in self._manager.getProfileSupportedFileTypesRead().items(): #Format the file types that can be read correctly for the open file dialog.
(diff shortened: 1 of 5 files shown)


UM/Qt/Bindings/ProfilesModel.py
```diff
@@ -108,18 +117,72 @@ def exportProfile(self, name, url):
         if not profile:
             return
 
-        error = ""
+        #Parse the fileType to deduce what plugin can save the file format.
+        #fileType has the format "<description> (*.<extension>)"
+        split = fileType.rfind(" (*.") #Find where the description ends and the extension starts.
+        if split < 0: #Not found. Invalid format.
+            Logger.log("e", "Invalid file format identifier %s", fileType)
```
awhiemstra: Shouldn't you return after this? Now you have an invalid identifier but still try to open it.

UM/Qt/Bindings/ProfilesModel.py
```diff
@@ -108,18 +117,72 @@ def exportProfile(self, name, url):
         if not profile:
             return
 
-        error = ""
+        #Parse the fileType to deduce what plugin can save the file format.
```
awhiemstra: It would be nice to not have to deal with the parsing here and instead use a dict lookup or similar to retrieve the plugin that corresponds to a certain file type. This is what I do in LocalFileOutputDevice. For now, you can leave this as is but please add a TODO comment here about this.

UM/Settings/MachineManager.py
```diff
@@ -351,6 +357,65 @@ def saveProfiles(self):
...
+    #   \return All profile writers that are currently loaded.
+    def getProfileWriters(self):
+        return self._profile_writers.items()
+
+    ##  Returns a dictionary of the file types the profile readers can read.
+    #
+    #   Each file extension should have a description.
+    #
+    #   \return A dictionary of the file types the profile readers can read.
+    def getProfileSupportedFileTypesRead(self):
```
awhiemstra: getSupportedProfileTypesRead would sound less awkward to me.

UM/Settings/MachineManager.py
```diff
@@ -351,6 +357,65 @@ def saveProfiles(self):
...
+                        supported_types[extension] = description
+
+        return supported_types
+
+    ##  Returns a dictionary of the file types the profile writers can write.
+    #
+    #   Each file extension should have a description.
+    #
+    #   \return A dictionary of the file types the profile writers can write.
+    def getProfileSupportedFileTypesWrite(self):
```
awhiemstra: As above, getSupportedProfileTypesWrite

PR #85, opened 2015-12-18
Add uppercase extensions to File Open dialog

fieldOfView:
Fixes CURA-567
See http://stackoverflow.com/questions/26867595/qml-filedialog-namefilters-case-sensitive-on-linux

### DIFF
--- a/UM/Qt/Bindings/MeshFileHandlerProxy.py
+++ b/UM/Qt/Bindings/MeshFileHandlerProxy.py
@@ -16,6 +16,7 @@ from UM.Operations.AddSceneNodeOperation import AddSceneNodeOperation
 from UM.Message import Message
 
 import os.path
+import platform
 
 from UM.i18n import i18nCatalog
 i18n_catalog = i18nCatalog("uranium")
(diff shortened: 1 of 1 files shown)


UM/Qt/Bindings/MeshFileHandlerProxy.py
```diff
@@ -31,9 +32,14 @@ def supportedReadFileTypes(self):
         file_types = []
         all_types = []
 
-        for ext, desc in self._mesh_handler.getSupportedFileTypesRead().items():
-            file_types.append("{0} (*.{1})".format(desc, ext))
-            all_types.append("*.{0}".format(ext))
+        if platform.system() == "Linux":
```
daid: Why is this linux specific? Don't think it harms to do this on all OSes.

fieldOfView: It is linux specific because the other platforms do not require it (they care a bit less about case sensitivity), and it looks slightly weird to have all extensions listed twice in the file open dialog.

PR #87, opened 2015-12-30
Per-machineinstance profiles

fieldOfView:
There are "starter-profiles" (eg Low, Normal, High, Ulti) for each Machine Type (eg UMO+) and Machine Variant (eg UM2+ 0.25mm nozzle).

Fields can be added to profile files:
```
[general]
...
machine_type=Ultimaker 2
machine_variant=0.25mm Nozzle
```
Cura will filter the profiles to show just the profiles for the currently selected machine. Custom profiles (created by the user in Cura by modifiying a starter profile) add an additional field:
```
[general]
...
machine_instance=Aldo's Ultimaker 2
```
This is used to show that profile only for the machine instance for which the profile has been created. This mechanism can be extended to filter profiles based on materials ([general] material=PLA).

From a user's perspective not much has changed; the proper profiles are loaded automatically. 

The current active profile is removed from cura.cfg, and is now stored per machine instance.

CURA-568
CURA-569

### DIFF
--- a/UM/Qt/Bindings/MachineManagerProxy.py
+++ b/UM/Qt/Bindings/MachineManagerProxy.py
@@ -109,9 +109,9 @@ class MachineManagerProxy(QObject):
                 custom_profile = deepcopy(profile)
                 custom_profile.setReadOnly(False)
                 custom_profile.setName(custom_profile_name)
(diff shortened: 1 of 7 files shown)


fieldOfView:
Closing because further work is needed to support materials

PR #89, opened 2016-01-06
New plugin type: Backend output writers

Ghostkeeper:
To separate the writers for the output from the backend from the writers for saving meshes, a new plugin type is created: Backend output writers. This writes the output from the backend somewhere.

This addresses issue CURA-611, which is actually a bug but since this change is so invasive I decided to make a branch for it. This corresponds to the branch feature_outputwriterplugin in Cura (sorry, I didn't pay attention to the plurality).


Ghostkeeper:
Sorry, I meant to merge this into 2.1.

PR #91, opened 2016-01-15
Scanning i18n directory for languages

thopiekar:
Instead of having a list of languages hardcoded, we scan the i18n directory for available languages.
It was a TODO, which was left after fixing po-file installation. The reference to the added macro is there as an URL.

This has been already applied to Cura, but same is needed here (commit e1966a7ea59933f72e1a8fdf35ae06398c626f4b)


thopiekar:
@TMHogenhout Ready to be merged :smile:

nallath:
This is so much better :)

PR #92, opened 2016-01-19
[WIP] i18n: Adding new set of helper commands

thopiekar:
After @TMHogenhout added a documentation how to translate Uranium (and Cura), I decided to create some commands to CMake.

I think about calling the commands like this:
```i18n-create-po``` - to (re-)create po-file for all languages
```i18n-create-po-de``` - to (re-)create po-file for _de_ language
```i18n-create-po-[etc]``` - for every directory at resources/i18n/*

```i18n-update-po``` - to update po-file all languages
```i18n-update-po-de``` - to update po-file for _de_ language
```i18n-update-po-[etc]``` - for every directory at resources/i18n/*

```i18n-create-mo``` - to (re-)create mo-file for all languages
```i18n-create-mo-de``` - to (re-)create mo-file for _de_ language
```i18n-create-mo-[etc]``` - for every directory at resources/i18n/*

I already played a bit with it and it works so far. I also ran into an problem in FindGettext.cmake, which does not find "msginit", so all ```i18n-create-*``` do not work for me

What do you think?


thopiekar:
Oops, wrong branch. Will reopen a new PR for this...

PR #93, opened 2016-01-19
[WIP] i18n: Adding new set of helper commands

thopiekar:
After @TMHogenhout added a documentation how to translate Uranium (and Cura), I decided to create some commands to CMake.

I think about calling the commands like this:
```make i18n-create-po``` - to (re-)create po-file for all languages
```make i18n-create-po-de``` - to (re-)create po-file for _de_ language
```make i18n-create-po-[etc]``` - for every directory at resources/i18n/*

```make i18n-update-po``` - to update po-file all languages
```make i18n-update-po-de``` - to update po-file for _de_ language
```make i18n-update-po-[etc]``` - for every directory at resources/i18n/*

```make i18n-create-mo``` - to (re-)create mo-file for all languages
```make i18n-create-mo-de``` - to (re-)create mo-file for _de_ language
```make i18n-create-mo-[etc]``` - for every directory at resources/i18n/*

I already played a bit with it and it works so far. I also ran into an problem in FindGettext.cmake, which does not find "msginit", so all ```i18n-create-*``` do not work for me
However, in case it works, it is easy then to add new language support. Just ```mkdir resources/i18n/ru```, rerun ```cmake .``` and run ```make i18n-create-po-ru```, for example.

----

TODOs:
* Add version to CMakeLists and only add "-f" when URANIUM_VERSION == "master"
- https://github.com/Ultimaker/Cura/commit/2300410c2ef29362f0cbf8480e6e7b702674cdf9

----

What do you think?

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -61,7 +61,9 @@ if(GETTEXT_FOUND)
         file(GLOB po_files ${CMAKE_SOURCE_DIR}/resources/i18n/${lang}/*.po)
         foreach(po_file ${po_files})
             string(REGEX REPLACE ".*/(.*).po" "${CMAKE_BINARY_DIR}/resources/i18n/${lang}/LC_MESSAGES/\\1.mo" mo_file ${po_file})
-            add_custom_command(TARGET translations POST_BUILD COMMAND mkdir ARGS -p ${CMAKE_BINARY_DIR}/resources/i18n/${lang}/LC_MESSAGES/ COMMAND ${GETTEXT_MSGFMT_EXECUTABLE} ARGS ${po_file} -o ${mo_file})
+            add_custom_command(TARGET translations POST_BUILD
+                               COMMAND mkdir ARGS -p ${CMAKE_BINARY_DIR}/resources/i18n/${lang}/LC_MESSAGES/
+                               COMMAND ${GETTEXT_MSGFMT_EXECUTABLE} ARGS ${po_file} -o ${mo_file})
         endforeach()
     endforeach()
     install(DIRECTORY ${CMAKE_BINARY_DIR}/resources DESTINATION ${CMAKE_INSTALL_DATADIR}/uranium/)
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -50,11 +50,6 @@ if(GETTEXT_FOUND)
     # The files are checked for a _qt suffix and if it is found, converted to
     # qm, otherwise they are converted to .po.
     add_custom_target(translations ALL)
-    # copy-translations can be used to copy the built translation files from the
-    # build directory to the source resources directory. This is mostly a convenience
-    # during development, normally you want to simply use the install target to install
-    # the files along side the rest of the application.
-    add_custom_target(copy-translations)
 
     SUBDIRLIST(languages ${CMAKE_SOURCE_DIR}/resources/i18n/)
     foreach(lang ${languages})
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -70,7 +70,6 @@ if(GETTEXT_FOUND)
             else()
                 message("GETTEXT_MSGINIT_EXECUTABLE not found!")
             endif()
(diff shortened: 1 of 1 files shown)

### CODE REVIEW

**Comment by thopiekar** [comment]
Just uploaded now my code. This is how it looks like, when executing the new targets.
```
thopiekar@home:[...]/Uranium$ make i18n-create-mo
Scanning dependencies of target i18n-create-mo-fr
Built target i18n-create-mo-fr
Scanning dependencies of target i18n-create-mo-fi
Built target i18n-create-mo-fi
Scanning dependencies of target i18n-create-mo-en
Built target i18n-create-mo-en
Scanning dependencies of target i18n-create-mo-de
Built target i18n-create-mo-de
Scanning dependencies of target i18n-create-mo
Built target i18n-create-mo
thopiekar@home:[...]/Uranium$ make i18n-update-po
.................... fertig.
Built target i18n-update-po-fr
................... fertig.
Built target i18n-update-po-fi
.................... fertig.
Built target i18n-update-po-en
.................... fertig.
Built target i18n-update-po-de
Built target i18n-update-po
thopiekar@home:[...]/Uranium$ make i18n-create-po
Built target i18n-create-po-fr
Built target i18n-create-po-fi
Built target i18n-create-po-en
Built target i18n-create-po-de
Built target i18n-create-po
```

**Comment by nallath** [comment]
I like your method of doing it. I don't have all that much experience with build scripts though, so I don't know if this is the best way of doing it. Maybe @awhiemstra can say more about that.

**Comment by thopiekar** [comment]
Do you want this rebased on master?

(If so, please do not close this PR. I'll close it, when the new PR is ready.)

**Comment by nallath** [comment]
I think it's a good idea to rebase this to master. @Ghostkeeper @awhiemstra   What do you guys think?

**Comment by Ghostkeeper** [comment]
From a feature standpoint it'll be nice to have I think, though the current system is not very hard to use either.

From a technical standpoint it's hard to say without testing it ;)

**Comment by thopiekar** [comment]
@Ghostkeeper doesn't sound like being enthusiastic about this PR :stuck_out_tongue_winking_eye: 

If I imagine I would start with Cura and want to add new language support it would be hard for me to jump in. I never expect that other people are experienced like me, so I created these commands just to make it easier for other native speaking people to jump in.

PS: Don't worry, I'm preparing some other easter eggs, which (hopefully) will be more helpful for you :smile:

**Comment by Ghostkeeper** [comment]
Don't get me wrong. I'm okay with it.

I just don't trust myself with reviewing this code since I have not enough experience with CMake. I can sort of just make out what it does...

**Comment by nallath** [comment]
I hardly have any experience with CMake as well. @awhiemstra should probably review this.

File: CMakeLists.txt [hunk 1]
```diff
@@ -49,19 +49,41 @@ if(GETTEXT_FOUND)
...
+    add_custom_target(i18n-update-po)
+    add_custom_target(i18n-create-mo ALL)
 
     SUBDIRLIST(languages ${CMAKE_SOURCE_DIR}/resources/i18n/)
-    foreach(lang ${languages})
-        file(GLOB po_files ${CMAKE_SOURCE_DIR}/resources/i18n/${lang}/*.po)
+    file(GLOB pot_files ${CMAKE_SOURCE_DIR}/resources/i18n/*.pot)
+    foreach(language ${languages})
+        # *.po files
+        add_custom_target(i18n-create-po-${language})
```
awhiemstra: It would be nice if we can move all this code into its own macro, since we now create individual targets for each translation.

awhiemstra:
Apart from the inline comment that would be nice to have, this looks pretty good. If we can somehow generalise this to something that would also be usable in Cura it would be even better. Right now we have duplicated code there...

thopiekar:
I guess porting it to Cura wouldn't be difficult, as this section looks almost copy-pasted.
But I have no idea now to tell CMake to make a module of it, so we could "import" it in Cura when executing "cmake".

I've created a new PR on this rebased on master here: https://github.com/Ultimaker/Uranium/pull/119

CMakeLists.txt
thopiekar: Ok, I'll give it a try :+1:

PR #95, opened 2016-01-20
Setting dependancy

nallath:


UM/Settings/SettingOverrideDecorator.py
```diff
@@ -27,14 +27,17 @@ def getAllSettings(self):
     def getAllSettingValues(self):
         self._temp_values = {}
         instance = Application.getInstance().getMachineManager().getActiveMachineInstance()
-
+        #instance.getMachineDefinition().updateRequiredBy()
```
awhiemstra: Please just remove the entire line instead of commenting it.

awhiemstra:
Apart from the minor note everything seems fine and it can be merged.

PR #98, opened 2016-01-29
Profiles rework

fieldOfView:
This is a reimagining of how profiles work in Cura. Instead of automatically creating new profiles when the user tries to change a setting on a "readonly" profile and having multiple changed profiles, there is only one "working profile" per machine type. This is more akin to how Legacy Cura worked.

This PR also implements different profiles for different printer types, variants (nozzle sizes in the UI) and materials. Finally it contains numerous little UX improvements related to profiles.

Also see Cura PR: https://github.com/Ultimaker/Cura/pull/618

Fixes or contributes to:
CURA-547
CURA-430 (hopefully)
CURA-805
CURA-199
CURA-402
CURA-583
CURA-626
CURA-24
CURA-751
CURA-569
CURA-568
CURA-432

### DIFF
--- a/UM/Qt/Bindings/ProfilesModel.py
+++ b/UM/Qt/Bindings/ProfilesModel.py
@@ -197,14 +197,8 @@ class ProfilesModel(ListModel):
         self._onProfilesChanged()
 
         #Restore active profile for this machine_instance.
-        active_profile = None
         active_instance_name = self._manager.getActiveMachineInstance().getActiveProfileName()
-        #We can't use findProfile(active_instance_name) because there may be multiple profiles with the same name on different machine instances.
-        profiles = self._manager.getProfiles()
-        for profile in profiles:
-            if profile.getName() == active_instance_name:
-                active_profile = profile
-                break
+        active_profile = self._manager.findProfile(name)
 
         self._manager.setActiveProfile(active_profile)
 
--- a/UM/Qt/Bindings/ProfilesModel.py
+++ b/UM/Qt/Bindings/ProfilesModel.py
@@ -26,6 +26,7 @@ class ProfilesModel(ListModel):
         super().__init__(parent)
 
         self._add_use_global = False
+        self._add_working_profile = False
 
         self.addRoleName(self.IdRole, "id")
         self.addRoleName(self.NameRole, "name")
@@ -54,6 +55,19 @@ class ProfilesModel(ListModel):
     @pyqtProperty(bool, fset = setAddUseGlobal, notify = addUseGlobalChanged)
     def addUseGlobal(self):
         return self._add_use_global
+
+    addWorkingProfileChanged = pyqtSignal()
+
+    def setAddWorkingProfile(self, add):
+        if add != self._add_working_profile:
+            self._add_working_profile = add
+            self._onProfilesChanged()
+            self.addWorkingProfileChanged.emit()
+
+    ##  Whether to add a "Current Settings;8" entry.
+    @pyqtProperty(bool, fset = setAddWorkingProfile, notify = addWorkingProfileChanged)
+    def addWorkingProfile(self):
+        return self._add_working_profile
 
     @pyqtSlot(str)
     def removeProfile(self, name):
@@ -215,6 +229,24 @@ class ProfilesModel(ListModel):
                 "value": "global",
                 "settings": None
             })
+
+        if self._add_working_profile:
+            profile = self._manager.getWorkingProfile()
+            settings_dict = profile.getChangedSettings()
+            settings_list = []
+            for key, value in settings_dict.items():
+                setting = self._manager.getActiveMachineInstance().getMachineDefinition().getSetting(key)
+                settings_list.append({"name": setting.getLabel(), "value": value})
+            settings_list = sorted(settings_list, key = lambda setting:setting["name"])
+            self.appendItem({
+                "id": 1,
+                "name": catalog.i18nc("@item:inlistbox", "Current settings"),
+                "active": False,
+                "readOnly": True,
+                "value": "working",
+                "settings": settings_list
+            })
+
 
         profiles = self._manager.getProfiles()
         profiles.sort(key = lambda k: k.getName())
--- a/UM/Qt/Bindings/ProfilesModel.py
+++ b/UM/Qt/Bindings/ProfilesModel.py
@@ -200,6 +200,13 @@ class ProfilesModel(ListModel):
         active_instance_name = self._manager.getActiveMachineInstance().getActiveProfileName()
         active_profile = self._manager.findProfile(active_instance_name)
 
+        if not active_profile:
+            #A profile by this name is no longer in the filtered list of profiles.
+            profiles = self._manager.getProfiles()
+            for profile in profiles:
+                active_profile = profile #Default to first profile you can find.
+                break
+
         self._manager.setActiveProfile(active_profile)
 
     def _onProfilesChanged(self):
--- a/UM/Qt/Bindings/ProfilesModel.py
+++ b/UM/Qt/Bindings/ProfilesModel.py
@@ -237,14 +237,17 @@ class ProfilesModel(ListModel):
                 "settings": None
             })
 
-        if self._add_working_profile:
+        active_machine = self._manager.getActiveMachineInstance()
+
+        if self._add_working_profile and active_machine:
             profile = self._manager.getWorkingProfile()
             settings_dict = profile.getChangedSettings()
             settings_list = []
-            for key, value in settings_dict.items():
-                setting = self._manager.getActiveMachineInstance().getMachineDefinition().getSetting(key)
-                settings_list.append({"name": setting.getLabel(), "value": value})
-            settings_list = sorted(settings_list, key = lambda setting:setting["name"])
+            if active_machine:
+                for key, value in settings_dict.items():
+                    setting = active_machine.getMachineDefinition().getSetting(key)
+                    settings_list.append({"name": setting.getLabel(), "value": value})
+                settings_list = sorted(settings_list, key = lambda setting:setting["name"])
             self.appendItem({
                 "id": 0,
                 "name": catalog.i18nc("@item:inlistbox", "Current settings"),
@@ -260,9 +263,10 @@ class ProfilesModel(ListModel):
         for profile in profiles:
             settings_dict = profile.getChangedSettings()
             settings_list = []
-            for key, value in settings_dict.items():
-                setting = self._manager.getActiveMachineInstance().getMachineDefinition().getSetting(key)
-                settings_list.append({"name": setting.getLabel(), "value": value})
+            if active_machine:
+                for key, value in settings_dict.items():
+                    setting = self._manager.getActiveMachineInstance().getMachineDefinition().getSetting(key)
+                    settings_list.append({"name": setting.getLabel(), "value": value})
             settings_list = sorted(settings_list, key = lambda setting:setting["name"])
             self.appendItem({
                 "id": id(profile),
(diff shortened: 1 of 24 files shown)


UM/Qt/Bindings/ProfilesModel.py
```diff
@@ -54,9 +61,22 @@ def setAddUseGlobal(self, add):
     def addUseGlobal(self):
         return self._add_use_global
 
+    addWorkingProfileChanged = pyqtSignal()
+
+    def setAddWorkingProfile(self, add):
+        if add != self._add_working_profile:
+            self._add_working_profile = add
+            self._onProfilesChanged()
+            self.addWorkingProfileChanged.emit()
```
Ghostkeeper: I'm not sure what this function and the above pyqtSignal does. Perhaps you could document at least what the signal is? Or what an "add working profile" is as a noun?

But it seems that `addWorkingProfileChanged` and `setAddWorkingProfile` have some sub-par names in any case.

UM/Qt/Bindings/ProfilesModel.py
```diff
@@ -93,31 +114,29 @@ def importProfile(self, url):
...
-                    self._manager.addProfile(profile)
-                    return { "status": "duplicate", "message": catalog.i18nc("@info:status", "Profile was imported as {0}", name) }
-                else:
-                    return { "status": "ok", "message": catalog.i18nc("@info:status", "Successfully imported profile {0}", profile.getName()) }
+                profile.setName(self._manager.makeUniqueProfileName(profile.getName())) #Ensure a unique name
+                if profile.getMachineTypeId():
+                    #Make sure the profile is available for the currently selected printer
+                    profile.setMachineTypeId(self._manager.getActiveMachineInstance().getMachineDefinition().getProfilesMachineId())
+                self._manager.addProfile(profile) #Add the new profile to the list of profiles.
+                return { "status": "ok", "message": catalog.i18nc("@info:status", "Successfully imported profile {0}", profile.getName()) }
```
Ghostkeeper: Is it a good idea to also make the newly imported profile the active profile? Maybe that isn't something that should be done by this function though (but rather at the code that calls this function: the button event).

UM/Qt/Bindings/ProfilesModel.py
```diff
@@ -93,31 +114,29 @@ def importProfile(self, url):
...
-    def exportProfile(self, name, url, fileType):
+    @pyqtSlot(int, str, QUrl, str)
+    def exportProfile(self, id, name, url, fileType):
         #Input checking.
         path = url.toLocalFile()
         if not path:
             return
 
-        profile = self._manager.findProfile(name)
+        if id==-1:
```
Ghostkeeper: Spaces around `==`.

Ghostkeeper: Also, be sure to document cases like this well! Otherwise, the caller of the function will probably have no idea that -1 can be used to export the working profile.

UM/Qt/Bindings/ProfilesModel.py
```diff
@@ -192,6 +211,31 @@ def getFileNameFiltersWrite(self):
...
+        #Restore active profile for this machine_instance.
+        active_instance_name = self._manager.getActiveMachineInstance().getActiveProfileName()
+        active_profile = self._manager.findProfile(active_instance_name, instance = self._manager.getActiveMachineInstance())
+
+        if not active_profile:
+            #A profile by this name is no longer in the filtered list of profiles.
+            profiles = self._manager.getProfiles(instance = self._manager.getActiveMachineInstance())
+            for profile in profiles:
+                active_profile = profile #Default to first profile you can find.
+                break
```
Ghostkeeper: Could it happen that a machine instance has no profiles at all? Or do we require that for every filter there is at least one profile?

UM/Settings/MachineManager.py
```diff
@@ -67,6 +71,46 @@ def getAllMachineVariants(self, machine_id):
...
+            if not material or material in machine_materials:
+                continue
+
+            profile_machine_type = profile.getMachineTypeId()
+            profile_machine_variant = profile.getMachineVariantName()
+            profile_machine_instance = profile.getMachineInstanceName()
+
+            if (profile_machine_instance and profile_machine_instance == instance_id) or \
+                    (profile_machine_variant and profile_machine_variant == machine_variant and profile_machine_type == machine_type) or \
+                    (profile_machine_type == machine_type):
```
Ghostkeeper: This warrants some documentation... Fairly ugly construct itself, too, but I wouldn't know a better one.

UM/Settings/MachineManager.py
```diff
@@ -67,6 +71,46 @@ def getAllMachineVariants(self, machine_id):
...
+                machine_materials.append(material)
+
+        if len(machine_materials) > 0:
+            return machine_materials
+        elif machine.getMachineDefinition().getSetting("machine_gcode_flavor").getDefaultValue() == "UltiGCode":
+            #UltiGCode printers don't use generic materials in Cura
+            return []
```
Ghostkeeper: Then shouldn't this check be before the `len(machine_materials)` check?

UM/Settings/MachineManager.py
```diff
@@ -109,8 +156,11 @@ def removeMachineInstance(self, instance):
         self._machine_instances.remove(instance)
         instance.nameChanged.disconnect(self._onInstanceNameChanged)
 
+        file_name = urllib.parse.quote_plus(instance.getName())
         try:
-            path = Resources.getStoragePath(Resources.MachineInstances, urllib.parse.quote_plus(instance.getName()) + ".cfg")
+            path = Resources.getStoragePath(Resources.MachineInstances, file_name + ".cfg")
+            os.remove(path)
+            path = Resources.getStoragePath(Resources.MachineInstanceProfiles, file_name + ".cfg")
```
Ghostkeeper: There could be more than one profile associated with one machine instance though?

UM/Settings/MachineManager.py
```diff
@@ -150,7 +200,42 @@ def setActiveMachineInstance(self, machine):
...
         self._updateSettingVisibility(setting_visibility)
 
+        self._protect_working_profile = True
+        profile = self.findProfile(machine.getActiveProfileName(), instance = self._active_machine)
+        if profile:
+            self.setActiveProfile(profile)
+        else:
+            for profile in self._profiles:
+                self.setActiveProfile(profile) #default to first profile you can find
+                break
```
Ghostkeeper: What happens if `self._profiles` is empty?

UM/Settings/MachineManager.py
```diff
@@ -150,7 +200,42 @@ def setActiveMachineInstance(self, machine):
...
+            return
+
+        emit = False
+        if material != self._active_machine.getMaterialName():
+            self._active_machine.setMaterialName(material)
+            emit = True
+
+        material_profile = self.findProfile(material, type = "material", instance = self._active_machine)
+        if material_profile:
+            self._active_machine.getWorkingProfile().mergeSettingsFrom(material_profile, reset = False)
```
Ghostkeeper: I think `emit` should be set to `True` here too, no? Since it changes the active machine.

UM/Settings/Profile.py
```diff
@@ -73,9 +122,15 @@ def isReadOnly(self):
     #
     #   \note If the setting is not a user-settable setting, this method will do nothing.
     def setSettingValue(self, key, value):
-        Logger.log('d' , "Setting value of setting %s to %s",key,value)
+        Logger.log('d' , "Setting value of %s to %s on profile %s",key,value,self._name)
```
Ghostkeeper: Code style: spaces after commas.

UM/Settings/Profile.py
```diff
@@ -30,7 +33,13 @@ def __init__(self, machine_manager, read_only = False):
         super().__init__()
         self._machine_manager = machine_manager
         self._changed_settings = {}
-        self._name = "Unknown Profile"
+        self._changed_settings_defaults = {}
```
Ghostkeeper: Why does the profile now need to carry a copy of the default values? Is it a copy, actually?

UM/Qt/Bindings/ProfilesModel.py
fieldOfView: The ```WorkingProfile``` (working as in "Working Title", not as in "Working Man") is the profile each instance has attached to it, which the user changes by changing settings in the UI.

```addWorkingProfile``` is a flag in the profilesmodel whether or not to list this Working Profile to the listmodel, analogous from the ```addUseGlobal``` flag.

```addWorkingProfileChanged``` is the signal the fires when the ```addWorkingProfile``` flag is changed, analogous to the existing ```addUseGlobalChanged```.

Though I agree that the naming reads like a random concatenation of verbs (and a noun), it is in line with the existing ````AddUseGlobal```` naming.

UM/Qt/Bindings/ProfilesModel.py
fieldOfView: I don't necessarily agree. It probably would be good if it were possible to "activate" a profile from the Manage Profiles dialog. Perhaps that should be done in a different issue unless we feel very strongly about it now (but it would be another button, with another untranslated string).

UM/Qt/Bindings/ProfilesModel.py
fieldOfView: Will do (2x)

UM/Qt/Bindings/ProfilesModel.py
fieldOfView: If a machine instance has no profiles at all, it gets the "generic" profiles (Low, Normal, High, Ulti). If those 4 are not available, then something is wrong.

UM/Settings/MachineManager.py
fieldOfView: I'm open to suggestions. I guess the variable names could use some work to make it more self-explanatory too.

UM/Settings/MachineManager.py
fieldOfView: No, because there are UltiGCode printers that will have material profiles (UM2+, UM3) (defeating the whole purpose of ultigcode, I know). I'll add more/better documentation.

UM/Settings/MachineManager.py
fieldOfView: This is the "WorkingProfile", and each instance has exactly 1 of those (no less, no more). The normal profiles are stored in a different folder.

UM/Settings/MachineManager.py
fieldOfView: Nothing.
I copied this approach from somewhere. I guess it could be rewritten as ```if len(self._profiles) > 0: self.setActiveProfile(self._profiles[0])```
Note that the case where ```self._profiles``` is empty should be very rare (if ever) because there will be either machine-specific or generic profiles. The generic profiles are basically always available.

UM/Settings/MachineManager.py
fieldOfView: > Since it changes the active machine.
It doesn't. It changes the working profile. The "material_profile" here is a partial profile, that gets merged into the current settings in the working profile. 

Needs more documentation.

UM/Settings/Profile.py
fieldOfView: The working profile carries a "copy" of the profile it was based on, so the reset value widget can reset to the value set in the profile the user selected. Otherwise we can only reset back to the machine default. See CURA-626

UM/Settings/Profile.py
fieldOfView: And no comma after 'd'. I copied someone's bad habbits, thanks for spotting.

UM/Qt/Bindings/ProfilesModel.py
Ghostkeeper: Yeah, I don't feel very strongly about it. I was just asking whether you think that's better. If it's ever deemed better to activate the imported profile immediately we can make an issue about it then.

PR #102, opened 2016-01-29
Introduce UM.Qt.Bindings.ContainerProxy class

awhiemstra:
This is a way to handle Python containers in QML without the performance penalties of constantly converting to QVariantMap/QVariantList.


nallath:
Looks good. Can be merged.

PR #106, opened 2016-02-04
Cmake files qtcreator

filushka:
Please, merge code to master. Additional CMakeLists.txt files allow to open Uranium project into Qtcreator.


fieldOfView:
Is there a JIRA issue for this with more information? If not, please give some more details about the usecase. What does this fix? Who will this benefit? Why does this need to be in the repository as opposed to on your machine? 

The main reason I ask is that it looks like a LOT of files that we will have to update each time files are added or removed. AFAIK none of us use Qtcreator, but perhaps you can convince us of its merrits, and that it is worth doing this additional maintenance work when code changes.

filushka:
It helps browse through the files inside native editor for qml and python files. This is a feature, which speed up developer' sjob.

PR #108, opened 2016-02-07
i18n: de: Another set of translation improvements

thopiekar:
Needed for PR: https://github.com/Ultimaker/Cura/pull/629

### DIFF
--- a/resources/i18n/de/uranium.po
+++ b/resources/i18n/de/uranium.po
@@ -5,16 +5,18 @@
 #
 msgid ""
 msgstr ""
-"Project-Id-Version: PACKAGE VERSION\n"
+"Project-Id-Version: \n"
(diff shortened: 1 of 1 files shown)


Ghostkeeper:
I'll redo the update for the .mo files shortly.

thopiekar:
Ah, OK, but you mean the po files I guess. mo are the compiled files :)

Ghostkeeper:
No, I meant the compiled files. We haven't set up yet that these are compiled automatically by https://github.com/Ultimaker/cura-build, so they still have to get updated manually into the https://github.com/Ultimaker/cura-binary-data repository. The reason for that is that we didn't want to depend on the extra programs that compile the .po files, which are platform-specific again and such.

Ghostkeeper:
.mo files have been updated.

PR #112, opened 2016-02-16
Fix performance issues

awhiemstra:
This makes profile switiching, variant switching and machine switching react in <500ms, with one minor visual regression: Setting items will pop-in relatively randomly when switching, instead of directly appearing. In my opinion, it is a trade-off we can live with.

Fixes CURA-472


awhiemstra:
Wrong branch

PR #113, opened 2016-02-16
Performance fixes

awhiemstra:
This makes profile switiching, variant switching and machine switching react in <500ms, with one minor visual regression: Setting items will pop-in relatively randomly when switching, instead of directly appearing. In my opinion, it is a trade-off we can live with.

Fixes CURA-472

### DIFF
--- a/UM/Qt/Bindings/SettingCategoriesModel.py
+++ b/UM/Qt/Bindings/SettingCategoriesModel.py
@@ -63,6 +63,8 @@ class SettingCategoriesModel(ListModel):
     def _onActiveMachineChanged(self):
         self._beginReset()
 
+        self._items = []
+
         if self._machine_instance:
             for category in self._machine_instance.getMachineDefinition().getAllCategories():
                 category.visibleChanged.disconnect(self._onCategoryVisibleChanged)
--- a/UM/Qt/Bindings/SettingCategoriesModel.py
+++ b/UM/Qt/Bindings/SettingCategoriesModel.py
@@ -1,7 +1,7 @@
 # Copyright (c) 2015 Ultimaker B.V.
 # Uranium is released under the terms of the AGPLv3 or higher.
 
-from PyQt5.QtCore import Qt, QCoreApplication, pyqtSlot
+from PyQt5.QtCore import Qt, QCoreApplication, pyqtSlot, pyqtProperty, pyqtSignal
 
 from UM.Qt.ListModel import ListModel
 from UM.Resources import Resources
@@ -32,17 +32,25 @@ class SettingCategoriesModel(ListModel):
         self.addRoleName(self.SettingsRole, "settings")
         self.addRoleName(self.HiddenValuesCountRole, "hiddenValuesCount") # Probably need a better name for this
 
+        self._resetting = False
+
     @pyqtSlot(str)
     def filter(self, text):
         for item in self.items:
             item["settings"].filter(text)
 
+    resettingChanged = pyqtSignal()
+    @pyqtProperty(bool, notify = resettingChanged)
+    def resetting(self):
+        return self._resetting;
+
     @pyqtSlot()
     def reload(self):
-        self.clear()
+        self._beginReset()
+        self._items = []
         if self._machine_instance:
             for category in self._machine_instance.getMachineDefinition().getAllCategories():
-                self.appendItem({
+                self._items.append({
                     "id": category.getKey(),
                     "name": category.getLabel(),
                     "icon": category.getIcon(),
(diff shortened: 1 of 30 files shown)


UM/Qt/Bindings/SettingCategoriesModel.py
```diff
@@ -1,7 +1,7 @@
 # Copyright (c) 2015 Ultimaker B.V.
 # Uranium is released under the terms of the AGPLv3 or higher.
 
-from PyQt5.QtCore import Qt, QCoreApplication, pyqtSlot
+from PyQt5.QtCore import Qt, QCoreApplication, pyqtSlot, pyqtProperty, pyqtSignal
```
Ghostkeeper: I don't see pyqtSignal being used any more.

UM/Qt/Bindings/Theme.py
```diff
@@ -32,36 +35,90 @@ def __init__(self, engine, parent = None):
...
         self._em_width = self._em_height;
 
-        self._initializeDefaults()
+        Preferences.getInstance().addPreference("general/theme", Application.getInstance().getApplicationName())
+
+        try:
+            theme_path = Resources.getPath(Resources.Themes, Preferences.getInstance().getValue("general/theme"))
+            self.load(theme_path)
+        except FileNotFoundError:
+            pass
```
Ghostkeeper: If the theme file is not found, `self._fonts`, `self._colors` and `self._sizes` will be left uninitialised. I think it'd be good to call `_initializeDefaults` nonetheless. Also, you might want to move the win32 font workaround into `_initializeDefaults`. If the font gets reset by `load` it might not get set back properly on win32.

UM/Qt/qml/UM/Settings/SettingCheckBox.qml
```diff
@@ -38,37 +103,6 @@ CheckBox
...
-                color: UM.Theme.colors.checkbox_mark
-                source: UM.Theme.icons.check
-                opacity: control.checked
-                Behavior on opacity { NumberAnimation { duration: 100; } }
-            }
-        }
+        indicator:
     }
-}
+}*/
```
Ghostkeeper: If you remove this comment, I think you'll see that git finds that a lot less code was changed than it reports now. Remove those old bits and pieces!

UM/Qt/qml/UM/Settings/SettingView.qml
```diff
@@ -129,66 +131,75 @@ ScrollView
...
-                                        preferences.getCurrentItem().scrollToSection(categoryId);
-                                    }
-                                }
+                            Connections
+                            {
+                                target: item;
+                                onItemValueChanged: delegateItem.settingsModel.setSettingValue(model.key, value);
+                                onContextMenuRequested: { contextMenu.key = delegateItem.categoryId; contextMenu.popup(); }
+                                onResetRequested: delegateItem.settingsModel.resetSettingValue(model.key);
+                                onShowTooltip: base.showTooltip(settingLoader, Qt.point(0, settingLoader.height / 2), "<b>"+model.name+"</b><br/>"+model.description)
```
Ghostkeeper: Code style: Whitespace around operators.

UM/Qt/qml/UM/Settings/SettingView.qml
```diff
@@ -227,5 +238,26 @@ ScrollView
...
+        Menu
+        {
+            id: contextMenu;
+
+            property string key;
+
+            MenuItem
+            {
+                //: Settings context menu action
+                text: catalog.i18nc("@action:menu","Hide this setting");
```
Ghostkeeper: Codestyle: Space after comma.

UM/Qt/qml/UM/Settings/SettingView.qml
```diff
@@ -227,5 +238,26 @@ ScrollView
...
+                onTriggered: delegateItem.settingsModel.hideSetting(key);
+            }
+            MenuItem
+            {
+                //: Settings context menu action
+                text: catalog.i18nc("@action:menu","Configure setting visiblity...");
```
Ghostkeeper: Codestyle: Space after comma.

UM/Qt/qml/UM/SimpleButton.qml
```diff
@@ -0,0 +1,36 @@
...
+    property alias color: image.color;
+    property alias backgroundColor: background.color;
+    property alias iconSource: image.source;
+
+    property alias hovered: base.containsMouse;
+
+    Rectangle {
+        id: background;
+        anchors.fill: parent;
+        color: "transparent";
```
Ghostkeeper: Shouldn't this (or the one below in RecolorImage) be something like `backgroundColor`?

awhiemstra: No, I alias the property in base, that way if you set base.backgroundColor it automatically sets the rectangle's color.

UM/Qt/Bindings/Theme.py
awhiemstra: What do you mean with "you might want to move the win32 font workaround into _initializeDefaults."? The system font that is initialized by _initializeDefaults is already the default system font and will be correct on Windows. The workaround when loading fonts is to replace the requested font family with the system font family, since other fonts really do not render well on Windows. Since _initializeDefaults does not load any other fonts, I do not see what I need to change there.

Ghostkeeper: I see. I guess that the `QCoreApplication.instance().setFont(default_font)` call would change the font in the QCoreApplication which then lets `_initializeDefaults` initialize the system font differently.

PR #115, opened 2016-03-07
Duplicate profile

fieldOfView:
Rename "Add" button on Profiles manager to "Duplicate", and add the functionality to duplicate profiles instead of just the current settings.

Updated from @aheinous' work.

Fixes CURA-801


UM/Qt/Bindings/MachineManagerProxy.py
```diff
@@ -166,6 +166,15 @@ def createProfile(self):
         self._manager.setActiveProfile(profile)
         return profile.getName()
 
+    @pyqtSlot(str, result = str)
+    def duplicateProfile(self, profile_name):
+        original_profile = self._manager.findProfile(profile_name, instance = self._manager.getActiveMachineInstance())
+        if not original_profile:
+            return
```
Ghostkeeper: Will this work? I think it would sometimes set the current profile to index -1 via the `selectProfile` signal in ProfilesPage.qml line 27.

UM/Qt/qml/UM/Preferences/ProfilesPage.qml
```diff
@@ -102,7 +118,13 @@ ManagementPage
         {
             id: renameDialog;
             object: base.currentItem != null ? base.currentItem.name : "";
+            property bool removeWhenRejected: false;
             onAccepted: base.model.renameProfile(base.currentItem.name, newName.trim());
+            onRejected: {
+                if(removeWhenRejected) {
+                    base.model.removeProfile(base.currentItem.name)
+                }
```
Ghostkeeper: This `if` condition will never be true?

fieldOfView: It will be true if the renameDialog is opened from ```onAddObject```

UM/Qt/Bindings/MachineManagerProxy.py
fieldOfView: Yes, that will work. Setting the currentIndex to -1 will deselect everything.

PR #116, opened 2016-03-10
[2.1] GLESv2: Just some little things

thopiekar:
Here are two commits that make Cura at least run on Ubuntu armhf.
There are still problems left... For example the preview widget (where you can see the build plate) is black and when choosing the layer view the log says about missing classes or functions in PyQt5.

I will give more details on that later, but before losing these changes I just want to share them with you.

* Related to an old conversation at https://github.com/Ultimaker/Cura/issues/573

### DIFF
--- a/UM/Qt/GL/QtOpenGL.py
+++ b/UM/Qt/GL/QtOpenGL.py
@@ -38,7 +38,10 @@ class QtOpenGL(OpenGL):
         self._gl.initializeOpenGLFunctions()
 
         self._gpu_vendor = OpenGL.Vendor.Other
-        vendor_string = self._gl.glGetString(self._gl.GL_VENDOR).lower()
+        vendor_string = self._gl.glGetString(self._gl.GL_VENDOR)
+        if vendor_string is None:
+            vendor_string = ""
+        vendor_string = vendor_string.lower()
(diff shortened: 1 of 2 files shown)


Ghostkeeper:
You're very brave to try this :stuck_out_tongue: 

Two things:
* We'll want this committed to master rather than 2.1. Only risk-free bugfixes are allowed on 2.1 now. The vendor string is risk-free, but the float precision is not. Could you make a new PR to master?
* The only developer who knows enough about this (what defines are defined on what sort of machines) or at least is experienced enough to find information about it, is @awhiemstra and he's enjoying his weekend early this week, so it'll have to wait until Monday.

Thanks for your efforts. I'm really impressed by your enthusiasm. We should totally invite you over to Ultimaker HQ sometime.

thopiekar:
OK, will resend two PRs with both commits cherry-picked.

I really love to hear you like my work :dancer: Often I don't get such great feedback and sharing my work is completely in my sense of opensource development. I know that slicing software (like other CAx software) can be very expensive. So as Cura is free and opensource it is an honor to help out wherever I can.

thopiekar:
I guess I can close this now :smile:

PR #117, opened 2016-03-11
[WIP] GLESv2: Defining precision for float

thopiekar:
When running Cura with closed-source Mali drivers it is needed to define the precision for float. I'm not sure (as I wrote this some time ago), but this was not needed if you are using software rendering via LLVM pipe.
The same effect can happen, when using other GLESv2 closed-source drivers.
- Related to https://github.com/Ultimaker/Uranium/pull/116

### DIFF
--- a/resources/shaders/composite.shader
+++ b/resources/shaders/composite.shader
@@ -13,6 +13,14 @@ vertex =
     }
 
 fragment =
+    #ifdef GL_ES
+        #ifdef GL_FRAGMENT_PRECISION_HIGH
+            precision highp float;
+        #else
+            precision mediump float;
+        #endif // GL_FRAGMENT_PRECISION_HIGH
+    #endif // GL_ES
+
     uniform sampler2D u_layer0;
     uniform sampler2D u_layer1;
     uniform sampler2D u_layer2;


nallath:
@awhiemstra Could you have a look at this? Seems a shame to let this pull request go to waste.

awhiemstra:
Hmm, actually, all the float values (including vec2, vec3 and vec4) should have a proper precision qualifier like is done in https://github.com/Ultimaker/Uranium/blob/master/resources/shaders/default.shader . QOpenGLShader includes a #define for the precision qualifiers on desktop GL, so those are not a problem for desktop GL.

thopiekar:
That's correct, but they are not set in Ubuntu (armhf) and probably on any
other Linux distribution using ARM. There is more work needed to get Cura
working there, but setting these precisions is a good beginning, I guess.

awhiemstra:
What I mean is that rather than put in a global define, add highp, mediump, lowp precision qualifiers to the variable declarations in the shaders. They are required for GLES but are ignored on Desktop GL. Otherwise you need to add the #defines on all shaders, which gets ugly rather quickly.

thopiekar:
Honestly I have no experiences with shaders. I moved my lines to resources/shaders/default.shader and it doesn't work :disappointed:

ChrisTerBeke:
Closing because old. Shaders also have changed since this PR was created. Still needed?

thopiekar:
Dunno, never tested again.

jwalt:
Please reopen. I can confirm that this is still required (and still works) on an ODROID XU4 board with Mali GPU.

Ghostkeeper:
The fix also needs to be applied to the `fragment41core` shader in that file, nowadays.

Ghostkeeper:
Developers, see issue CURA-4800.

thopiekar:
@Ghostkeeper Is it possible to import a section of code in these files? I would like to prevent copy/pasting..

Ghostkeeper:
Not that I know of. Maybe `#import` works?
See also https://stackoverflow.com/questions/20175340/how-to-import-libraries-in-opengl-glsl

alekseisasin:
@thopiekar Then this changes should be applied for xray_composite.shader, composite.shader, simulationview.shader.
Do we need to apply this change only for old Opengl version or also for #version 410 core ?

Ghostkeeper:
It would also be for the modern core. It's the Mali drivers that require this, not a specific version of OpenGL.

thopiekar:
Yep, the Mali drivers are somehow special... Will con it as soon as I have managed to get an armhf AppImage running, so I can easily distribute it, so people can test it.

PR #118, opened 2016-03-11
GLESv2: Make GL_VENDOR check more failsave

thopiekar:
When executing glGetString(self._gl.GL_VENDOR) it returns here on Ubuntu Xenial (armhf) using closed-source mali drivers "None".
Therefore .lower() fails. I think for now replacing the None-Type with "" is ok for now.

* Related to https://github.com/Ultimaker/Uranium/pull/116


Ghostkeeper:
Thanks!

PR #120, opened 2016-03-18
[master] tracing for logger

thopiekar:
This will add logger outputs like:

```
2016-03-18 15:51:06,332 - INFO - <PATH>/UM/PluginRegistry.py (loadPlugin [61]): Loaded plugin ChangeLogPlugin
2016-03-18 15:51:06,334 - INFO - <PATH>/UM/Qt/QtApplication.py (__init__ [36]): Command line arguments: {'file': [], 'disable-textures': False, 'external-backend': False, 'debug-mode': False}
2016-03-18 15:51:06,481 - WARNING - <PATH>/UM/Settings/MachineManager.py (saveVisibility [707]): No active machine found when trying to save setting visibility
2016-03-18 15:51:06,588 - WARNING - <PATH>/UM/Settings/Profile.py (setSettingValue [148]): Tried to set value of non-user setting platform_adhesion
2016-03-18 15:51:06,590 - WARNING - <PATH>/UM/Settings/Profile.py (setSettingValue [148]): Tried to set value of non-user setting platform_adhesion
2016-03-18 15:51:06,613 - INFO - <PATH>/UM/Backend/Backend.py (startEngine [45]): Started engine process: /usr/bin/CuraEngine
2016-03-18 15:51:06,653 - DEBUG - <PATH>/UM/Backend/Backend.py (_onSocketStateChanged [122]): Backend connected on port 49674
2016-03-18 15:51:06,654 - DEBUG - <PATH>/UM/Controller.py (setActiveView [83]): Setting active view to SolidView
```

### DIFF
--- a/UM/Logger.py
+++ b/UM/Logger.py
@@ -29,8 +29,8 @@ class Logger:
     #   \param *args \type{list} List of variables to be added to the message.
     @classmethod
     def log(cls, log_type, message, *args):
-        func = inspect.currentframe().f_back.f_code
-        address = "%s (%s [%s]): " %(func.co_filename, func.co_name, func.co_firstlineno)
+        function = inspect.currentframe().f_back.f_code
+        address = "%s (%s [%s]): " %(function.co_filename, function.co_name, function.co_firstlineno)
         for logger in cls.__loggers:
             filled_message = address + message % args # Replace all the %s with the variables. Python formating is magic.
             logger.log(log_type, filled_message)
--- a/UM/Logger.py
+++ b/UM/Logger.py
@@ -34,7 +34,7 @@ class Logger:
         filename = function.co_filename
         for path in sys.path:
             if filename.startswith(path):
-                filename = filename.replace(path, "<PATH>")
+                filename = filename.replace(path, "...")
                 continue
         address = "%s (%s [%s]): " %(filename, function.co_name, function.co_firstlineno)
         for logger in cls.__loggers:
--- a/UM/Logger.py
+++ b/UM/Logger.py
@@ -2,6 +2,7 @@
 # Uranium is released under the terms of the AGPLv3 or higher.
 
 import traceback
+import inspect
 
 from UM.PluginObject import PluginObject
 
--- a/UM/Logger.py
+++ b/UM/Logger.py
@@ -1,6 +1,7 @@
 # Copyright (c) 2015 Ultimaker B.V.
 # Uranium is released under the terms of the AGPLv3 or higher.
 
+import sys
 import traceback
 import inspect
 
--- a/UM/Logger.py
+++ b/UM/Logger.py
@@ -30,7 +30,12 @@ class Logger:
     @classmethod
     def log(cls, log_type, message, *args):
         function = inspect.currentframe().f_back.f_code
-        address = "%s (%s [%s]): " %(function.co_filename, function.co_name, function.co_firstlineno)
+        filename = function.co_filename
+        for path in sys.path:
+            if filename.startswith(path):
+                filename = filename.replace(path, "<PATH>")
+                continue
+        address = "%s (%s [%s]): " %(filename, function.co_name, function.co_firstlineno)
         for logger in cls.__loggers:
             filled_message = address + message % args # Replace all the %s with the variables. Python formating is magic.
             logger.log(log_type, filled_message)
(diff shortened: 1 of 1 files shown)


nallath:
I like the extra info, but not sure if we want to have this at all times (without any option to disable this).

thopiekar:
```
2016-03-18 16:21:07,195 - INFO - <PATH>/UM/Qt/QtApplication.py (__init__ [36]): Command line arguments: {'external-backend': False, 'debug-mode': False, 'file': [], 'disable-textures': False}
```

This line brings me to an idea: What about enabling it only when ```debug-mode == True```?

nallath:
I like that option.

thopiekar:
Ok, I'll need to take a look how to get ```debug_mode```. Hope I'll find
some lines where it is already used.

thopiekar:
Hmm, ```debug_mode``` is nothing that is available in Uranium. Cura is getting it directly from PyQt5.QtCore.

I always understood Uranium as layer between Cura and PyQt5, so I'm not sure whether it needs to be moved to Uranium, too.

Or do you think it will be save to extract it from https://github.com/Ultimaker/Cura/blob/master/cura/CrashHandler.py ?

thopiekar:
Btw, @nallath what do you think about https://github.com/Ultimaker/Uranium/pull/120/commits/c9e28c9dbf35771ee0d357eabed0e90f4471b36c?

Ghostkeeper:
I've tested it and it looks useful. Especially for users sending their logs over to us after a crash.

This brings up the matter a bit of how verbose the logging should be. Ideally the file logging would be more verbose than the console logging, but for that we'd have to implement different levels / types of logging per logger plug-in.

We'll merge this in any case, but is that something you'd like to help us with, @thopiekar?

UM/Logger.py
```diff
@@ -28,8 +30,15 @@ def getLoggers(cls):
     #   \param *args \type{list} List of variables to be added to the message.
     @classmethod
     def log(cls, log_type, message, *args):
+        function = inspect.currentframe().f_back.f_code
+        filename = function.co_filename
+        for path in sys.path:
+            if filename.startswith(path):
+                filename = filename.replace(path, "<PATH>")
```
awhiemstra: Minor nitpick: I find the <PATH> rather distracting, how about replacing with a simple : or ~ so you get ":/UM/Application.py" or "~/UM/Application.py"?

awhiemstra:
Debug mode is a switch I introduced to Cura to hide the crash dialog by default. We could definitely introduce something similar in Uranium to handle this in a more graceful way. Though I would start with a log-level switch to set the maximum log level to output.

UM/Logger.py
thopiekar: Using ```~``` is not much useful. I always thing about ```$HOME``` when I seeing it. But ```:```, or any other like ```#``` is ok.
Just decide which one you like the most.

thopiekar:
I even thought about a new logger target "dd" for a lot of messages. For example it should be ok to have all the messages, created by the 5s-Timers in the console, but not in the logs.

Btw.: What about the logfiles itself? Are they cutted in any way?

awhiemstra:
> Btw.: What about the logfiles itself? Are they cutted in any way?

Not at the moment. We should introduce some kind of log rotation there. Also, we should indicate when a new instance starts while we're at it.

thopiekar:
I even think more about storing them in a directory, eg.
```.local/share/cura/logs/```​ and give them timestamps in iso format. And
when there are files which don't fit a pattern (old logs), just remove them.

UM/Logger.py
thopiekar: So, how do you want it replaced now? Should I better take a ```#```? It would make it less confusing then ```~```. ```:``` looks somehow ugly :sweat_smile: 
What do you think, @nallath @Ghostkeeper ?

fieldOfView: Another option would be to remove the path entirely (so you end up with /UM/Application.py)

Ghostkeeper: How about `.../UM/Application.py`?

awhiemstra: ... is a very common approach. Sounds like a good suggestion to me. :)

awhiemstra:
Sooo... I was wondering the other day what happened with this and now see this is still open. I say we should merge this as it improves the usefulness of logging.

thopiekar:
@Ghostkeeper Could you be so kind and write me a mail. There is just something offtopic I want to talk about and I couldn't find any email address here.

Thanks :wink:

PR #123, opened 2016-03-30
WIP: [CURA-1696] i18n: Bringing back copy target(s)

thopiekar:
* In the past there was "copy-translations" which I removed as it wasn't in use. This commit brings it back as "i18n-copy-mo".
* In case no mo files are generated it will generate them before copying.
* If no directory is set the targets won't appear and a warning will be shown.
* CURA_BINARY_DATA_DIRECTORY can be set as a global varibale (in .profile for example) or as option ("cmake -D<PATH>").


Ghostkeeper:
@awhiemstra could you take a look at this? I don't understand the intended use of these targets myself, but maybe you have an idea what they're for. I'd hate to keep thopiekar waiting.

CMakeLists.txt
```diff
@@ -4,6 +4,13 @@ cmake_minimum_required(VERSION 2.8.12)
 
 include(GNUInstallDirs)
 
+option(CURA_BINARY_DATA_DIRECTORY "Directory to the cura-binary-data repository")
+
+if(NOT CURA_BINARY_DATA_DIRECTORY AND NOT $ENV{CURA_BINARY_DATA_DIRECTORY} EQUAL "")
```
awhiemstra: It is probably better to use DEFINED $ENV{CURA_BINARY_DATA_DIRECTORY}. In addition, EQUAL does not do string comparison, you need to use STREQUAL for that.

awhiemstra:
The original purpose of the copy target was to make development easier. Since we now have proper install targets for the translations, that is not much of an issue anymore. The new purpose seems to be to make it easier to update the cura-binary-data repo, which seems a reasonable thing. Though I am wondering if make install with DESTDIR set to the cura-binary-data repo dir would not also fix that.

thopiekar:
Done. Well, if you merge this no more experiments are needed :wink: 
I think it is also very useful to use autocompletation now, eg. ```make i18n-```[TAB][TAB]. It will list all generated targets for ```make```.

thopiekar:
Old PR from me, but I'll merge it before working on CURA-1696.

fieldOfView:
@thopiekar, it is not customary to merge your own branch.

Due to holidays we're understaffed, and I for one have no idea of the impact of these changes to our (fragile) build system. The issue is not in our current sprint, so even though you put it in "Code Review" we don't see it on our boards to review (and there is a LOT of work on our boards).

thopiekar:
True, but at the moment I'm also very busy and if things are blocking or there are bugs I probably couldn't fix I just go ahead with future issues. This PR was made a long time a go and was well tested.
Together with CURA-1696 I'll also add instructions to the docs to explain how to use the additional "Make" targets.
So merging this earlier (while there are only changes/fixes in Uranium during the last months) shouldn't be a problem in my point of view..

PR #126, opened 2016-04-17
[master] Mesh: Converting floating point faces into int()

thopiekar:
It seems that some CAD applications export the number of faces as floating point numbers. Our STL loader seems not to care about this and passes num_faces as float().
This commit converts the floating point value into int() as needed. Of course, we should check about this much earlier.

This might be useful to backport to a bugfix version of 2.1.


nallath:
Slight code nitpicking:
There is no need to use () in a single if statement in python (my IDE actually complains about it as being a style violation)

thopiekar:
Yes, this is correct! Often it is a bad habit :sweat_smile:

PR #129, opened 2016-04-21
Fix several pylint reported issues

awhiemstra:
This fixes a bunch of issues reported by pylint in files in the root of UM/. Most of them are fairly trivial but some things are a little more involved.


.pylint-ignore.dict
```diff
@@ -0,0 +1,21 @@
...
+bool
+reimplemented
+gil
+osx
+traceback
+shaders
+whitespace
+gettext
+i18n
+catalog
```
Ghostkeeper: I find it weird that it doesn't know `catalog`. What dialect of English was it using, I wonder?

UM/ColorGenerator.py
```diff
@@ -1,6 +1,7 @@
 # Copyright (c) 2015 Ultimaker B.V.
 # Uranium is released under the terms of the AGPLv3 or higher.
 
+#pylint: disable=bad-whitespace
```
Ghostkeeper: I think this one isn't necessary any more, right, since https://github.com/Ultimaker/Uranium/pull/129/commits/bdc41ac5bb6fc2c105360c639341d27bd42eceb9?

.pylint-ignore.dict
awhiemstra: Just "english". Could be that it is simply a bug in the aspell english dictionary.

UM/ColorGenerator.py
awhiemstra: Unfortunately, pylint only ignores this style of whitespace for dicts, not lists. So it is still needed.

PR #131, opened 2016-05-03
[master] ConsoleLogger: Optional colored output

thopiekar:
In case the colorlog module is installed, it's Formatter will be used to colorize the output.
It improves the possibility to find ERRORs and WARNINGs.
It should be no problem to make that available on Windows and OSX, too, as colorlog is crossplatform and handles the different ways to pass the colors to the terminals.

cura12

### DIFF
--- a/plugins/ConsoleLogger/ConsoleLogger.py
+++ b/plugins/ConsoleLogger/ConsoleLogger.py
@@ -5,13 +5,27 @@ from UM.Logger import LogOutput
 
 import logging
 
+try:
+    from colorlog import ColoredFormatter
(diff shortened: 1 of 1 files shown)


nallath:
Hmm. Doesn't seem to work for me. Could be that it's Konsole that is messing things up.

thopiekar:
I'm using Konsole as well. You probably need  to install python3-colorlog.
As stated in the commit message the code automatically falls back to the
regular uncolored output if that module is missing.
I just added it like that in case we don't want to ship 'colorlog' in the
cura bundles.

thopiekar:
I guess it is working now? :smile:

nallath:
I used color stuff with it the konsole before and that worked right away (so I assumed i didn't need any extra stuff). 
And yes. Works now. See you tonight!

thopiekar:
Ok, great!

See you :smiley:

PR #132, opened 2016-05-04
Optionally scale meshes that are so tiny they could not otherwise be selected

fieldOfView:
When objects have been modeled with meters as units, they become so tiny that the cannot be selected in Cura to be scaled up.

See https://github.com/Ultimaker/Cura/issues/697#issuecomment-209518061

Also see https://github.com/Ultimaker/Cura/pull/733

### DIFF
--- a/UM/Mesh/ReadMeshJob.py
+++ b/UM/Mesh/ReadMeshJob.py
@@ -13,6 +13,7 @@ from UM.Mesh.MeshReader import MeshReader
 
 import time
 import os.path
+import math
 
 from UM.i18n import i18nCatalog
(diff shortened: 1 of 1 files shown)


BagelOrb:
Why is this a pull request?

I think it's good we have this now!
I'd been arguing for this quite some time ago...

fieldOfView:
There is no CCB-approved jira issue for this functionality. And today is not a working day for me, so this should be regarded as a "community contribution".

BagelOrb:
I just do it on a separate branch and I do create a JIRA issue for it, so that you can easily see from JIRA which commits are involved...

PR #136, opened 2016-06-08
[master] CURA-1597: Fixing wrong characters read from JSON

thopiekar:
Python3 handles all strings internally as unicode, but when it comes to
read data from files it seems that Python still decides to use the
default encoding by the OS (here Windows). At least, this is the only
explaination I have on this behaviour.

This fix forces open(XY) to use UTF-8 as encoding and therefore the data
goes correctly to the JSON encoding algorithm and finally got correctly
displayed in the UI.

Again: This is for CURA-1597.


awhiemstra:
*slaps forehead* Duh!

I fixed it for the containers loaded by ContainerRegistry::load() but completely forgot that DefinitionContainer also loads file.

PR #138, opened 2016-06-10
temp fix: make machine settings visible again (CURA-1645)

BagelOrb:
There might be changes needed to the definition files.


fieldOfView:
Hmm, we agreed with Paul that there would only be profiles for 0.4mm nozzles. 
Showing the machine settings breaks things.

BagelOrb:
Breaks what?
I see no harm in giving our internal 'users' more flexibility.

fieldOfView:
CURA-1627
If paul wants to make profiles for 0.6mm nozzles, we can provide a 0.6mm variant. 

Also if we give Paul the option to include machine settings in quality profiles, he will (like he set all the nozzle sizes to 7/8 times the specified nozzle size). Making ~25 settings that should not be used visible just so 1 setting can be edited is not a good idea.

PR #140, opened 2016-06-14
Added convenience functions to better build meshes

Johan3DV:
Adds convenience functions to MeshData used in commits for Cura and CuraEngine to speed-up layerview

### DIFF
--- a/UM/Mesh/MeshData.py
+++ b/UM/Mesh/MeshData.py
@@ -244,6 +244,30 @@ class MeshData():
         self._vertex_count = 0
         self._face_count = 0
     
+    ##  Set the amount of faces and vertices before loading data to the mesh.
+    #
+    #   This way we can create the array before we fill it. This method will reserve
+    #   `num_vertices` amount of space for vertices, `num_vertices` amount of space
+    #   for colors and `num_faces` amount of space for indices.
+    #
+    #   \param num_faces Number of faces for which memory must be reserved.
+    #   \param num_vertices Number of vertices for which memory must be reserved.
+    def reserveFaceAndVerticeCount(self, num_faces, num_vertices):
+        if type(num_faces) == float:
+            Logger.log("w", "Had to convert 'num_faces' with int(): %s -> %s ", num_faces, int(num_faces))
+            num_faces = int(num_faces)
+        if type(num_vertices) == float:
+            Logger.log("w", "Had to convert 'num_vertices' with int(): %s -> %s ", num_vertices, int(num_vertices))
+            num_vertices = int(num_numvertices)
+
+        self._vertices = numpy.zeros((num_vertices, 3), dtype=numpy.float32)
+        self._colors = numpy.zeros((num_vertices, 4), dtype=numpy.float32)
+        self._indices = numpy.zeros((num_faces, 3), dtype=numpy.int32)
+
+        self._vertex_count = 0
+        self._face_count = 0
+
+    
     ##  Add a vertex to the mesh.
     #   \param x x coordinate of vertex.
     #   \param y y coordinate of vertex.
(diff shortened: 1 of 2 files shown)


nallath:
Note that the MeshData will be changed. Simeon made a ton of changes (so the MeshData is now imutable). I'd recommend waiting until that work is merged before having alook at merging this. See the simple_convex_hull branch.

Johan3DV:
For the simple_convex_hull the difference appears to be these functions should be part of MeshBuilder instead.

Thinking about the need for these functions can easily be removed from the related Cura PR.

UM/Mesh/MeshData.py
```diff
@@ -398,6 +422,24 @@ def addColors(self, colors):
...
+    # Add faces defined by indices into vertices with vetex colors defined by colors
+    # Assumes vertices and colors have the same length.
+    #
+    def addFacesWithColor(self, vertices, indices, colors):
+        #If badly used, crash and burn
+        if len(self._indices) <  self._face_count + len(indices) or len(self._colors) < self._vertex_count + len(colors) or len(self._vertices) < self._vertex_count + len(vertices):
+            Logger.log( "w", "Insufficient size of mesh_data: f_c: %s, v_c: %s, _in_l: %s, in_l: %s, _co_l: %s, co_l: %s, _ve_l: %s, ve_l: %s", self._face_count, self._vertex_count,len(self._indices),len(indices),len(self._colors),len(colors),len(self._vertices),len(vertices))
+            return
+
+        self._indices[self._face_count:(self._face_count+len(indices)),:] = self._vertex_count + indices
```
awhiemstra: Codestyle: space after ,

UM/Mesh/MeshData.py
```diff
@@ -398,6 +422,24 @@ def addColors(self, colors):
         else:
             self._colors = numpy.concatenate((self._colors[0:self._vertex_count], colors))
     
+    # Add faces defined by indices into vertices with vetex colors defined by colors
+    # Assumes vertices and colors have the same length.
+    #
+    def addFacesWithColor(self, vertices, indices, colors):
```
awhiemstra: For this entire method, please put a space after ,

UM/Mesh/MeshData.py
```diff
@@ -398,6 +422,24 @@ def addColors(self, colors):
...
+        #If badly used, crash and burn
+        if len(self._indices) <  self._face_count + len(indices) or len(self._colors) < self._vertex_count + len(colors) or len(self._vertices) < self._vertex_count + len(vertices):
+            Logger.log( "w", "Insufficient size of mesh_data: f_c: %s, v_c: %s, _in_l: %s, in_l: %s, _co_l: %s, co_l: %s, _ve_l: %s, ve_l: %s", self._face_count, self._vertex_count,len(self._indices),len(indices),len(self._colors),len(colors),len(self._vertices),len(vertices))
+            return
+
+        self._indices[self._face_count:(self._face_count+len(indices)),:] = self._vertex_count + indices 
+        self._face_count += len(indices)
+        
+        end_idx = self._vertex_count + len(vertices)
```
awhiemstra: Codestyle: Please do not abbreviate and instead just use end_index.

UM/Mesh/MeshData.py
```diff
@@ -244,6 +244,30 @@ def reserveVertexCount(self, num_vertices):
...
+    #
+    #   \param num_faces Number of faces for which memory must be reserved.
+    #   \param num_vertices Number of vertices for which memory must be reserved.
+    def reserveFaceAndVerticeCount(self, num_faces, num_vertices):
+        if type(num_faces) == float:
+            Logger.log("w", "Had to convert 'num_faces' with int(): %s -> %s ", num_faces, int(num_faces))
+            num_faces = int(num_faces)
+        if type(num_vertices) == float:
+            Logger.log("w", "Had to convert 'num_vertices' with int(): %s -> %s ", num_vertices, int(num_vertices))
+            num_vertices = int(num_numvertices)
```
awhiemstra: Probably a typo here, but you are using num_numvertices which is undefined.

UM/Mesh/MeshData.py
```diff
@@ -244,6 +244,30 @@ def reserveVertexCount(self, num_vertices):
...
+    #   `num_vertices` amount of space for vertices, `num_vertices` amount of space
+    #   for colors and `num_faces` amount of space for indices.
+    #
+    #   \param num_faces Number of faces for which memory must be reserved.
+    #   \param num_vertices Number of vertices for which memory must be reserved.
+    def reserveFaceAndVerticeCount(self, num_faces, num_vertices):
+        if type(num_faces) == float:
+            Logger.log("w", "Had to convert 'num_faces' with int(): %s -> %s ", num_faces, int(num_faces))
+            num_faces = int(num_faces)
+        if type(num_vertices) == float:
```
awhiemstra: For type checking, it is better to use isinstance(num_vertices, float) since that supports subclasses. However, in general it is better not to typecheck in python. If you want to ensure they are ints, just call int(num_vertices) when using it. That way anything that can be converted to an int is supported instead of special casing only for float.

awhiemstra:
The simple_convex_hull branch was merged, so you can now also move the related methods to the right place.

UM/Mesh/MeshData.py
Johan3DV: I agree type checking is not preferable but neither is allowing to call with a floating point number. I'm not a fan of dynamic typing.

I've rewritten the code to give a warning when this isn't an integer type. If you feel there is no need for this check it's easy to remove. I originally got the pattern from another function in this class.

Johan3DV:
I've moved the function to MeshBuilder and hope I've taken all comments into consideration.

awhiemstra:
Since this has no dependencies and the code looks good I am merging it. Thank you!

PR #141, opened 2016-06-16
Improvements to ContainerRegistry::load

awhiemstra:
This contains two major changes:

* Sort the list of files to load by the container type. This ensures we load DefinitionContainers before InstanceContainers and InstanceContainers before ContainerStacks.

* Use pickle to cache a binary version of the DefinitionContainer that is much faster to load. This reduces startup time on my machine from ~5 seconds to ~3 seconds.

### DIFF
--- a/UM/Settings/ContainerRegistry.py
+++ b/UM/Settings/ContainerRegistry.py
@@ -166,31 +166,43 @@ class ContainerRegistry:
     #   If this function is called again, it will clear the old data and reload.
     def load(self):
         files = []
-        files_resource_type = []
         for resource_type in self._resource_types:
             resources = Resources.getAllResourcesOfType(resource_type)
-            files.extend(resources)
-            files_resource_type.extend([resource_type]*len(resources))
 
-        for file_path, resource_type in zip(files, files_resource_type):
             try:
-                mime = MimeTypeDatabase.getMimeTypeForFile(file_path)
+                resource_storage_path = Resources.getStoragePathForType(resource_type)
+            except UnsupportedStorageTypeError:
+                resource_storage_path = ""
+
+            # Pre-process the list of files to insert relevant data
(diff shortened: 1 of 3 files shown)


fieldOfView:
This should also fix the issue where the machine definition is loaded before its extruders have loaded/deserialised? I was just filing an issue about that.

awhiemstra:
That is the exact issue this fixes yes. :)

UM/Settings/ContainerRegistry.py
```diff
@@ -166,36 +167,58 @@ def addContainerType(cls, container):
     #   If this function is called again, it will clear the old data and reload.
```
Ghostkeeper: You might want to expand the documentation of this function a bit. I think the last sentence is not entirely correct any more. It will reload, but it will reload from the cache anyway...

awhiemstra: It is not even correct anymore because we actually do not clear the existing list of containers.

UM/Settings/ContainerRegistry.py
```diff
@@ -4,6 +4,7 @@
 import os
 import re #For finding containers with asterisks in the constraints.
 import urllib
+import pickle
```
nallath: I really like Ghostkeepers style of adding a comment why we use certain libs. In the case of pickle, that might be a good one (as I never heard of it before seeing it here)

awhiemstra: Fixed.

UM/Settings/ContainerRegistry.py
awhiemstra: Fixed the comment and made sure we actually do not re-add already existing containers. (Since that was rather bad behaviour.)

UM/Settings/ContainerRegistry.py
nickthetait: I agree, it is incredibly valuable to know why an import was added.

PR #143, opened 2016-06-20
Feature stl speed improvement

nallath:
Improve speed of .stl loading by ~20x. 

CURA-1749

### DIFF
--- a/UM/Mesh/MeshData.py
+++ b/UM/Mesh/MeshData.py
@@ -228,6 +228,15 @@ class MeshData():
 
         self._vertex_count = 0
         self._face_count = 0
+
+
+    def setIndices(self, indices):
+        self._indices = indices
+        self._face_count = int(self._indices.size / 3)
+
+    def setVertices(self, vertices):
+        self._vertices = vertices
+        self._vertex_count = int(self._vertices.size / 3)
     
     ##  Set the amount of verts before loading data to the mesh.
     #
--- a/plugins/FileHandlers/STLReader/STLReader.py
+++ b/plugins/FileHandlers/STLReader/STLReader.py
@@ -5,15 +5,14 @@
 from UM.Mesh.MeshReader import MeshReader
 from UM.Mesh.MeshData import MeshData
 from UM.Logger import Logger
-from UM.Math.Matrix import Matrix
-from UM.Math.Vector import Vector
 from UM.Scene.SceneNode import SceneNode
-from UM.Job import Job
 
-import os
-import struct
-import math
-import time
+import numpy
+import stl  # numpy-stl lib
+import stl.mesh
+
+# Increase max count. (10 million should be okay-ish)
+stl.stl.MAX_COUNT = 10000000
 
 class STLReader(MeshReader):
     def __init__(self):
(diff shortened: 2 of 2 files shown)


thopiekar:
```
2016-06-20 16:20:41,934 - ERROR -.../UM/PluginRegistry.py (_findPlugin [261]): Import error loading module STLReader: No module named 'stl'
2016-06-20 16:20:41,935 - ERROR - .../UM/PluginRegistry.py (_populateMetaData [231]): Could not find plugin STLReader
```

I also couldn't find "stl" in the numpy package or somewhere else. Is it possible to keep the old code as a fallback?

nallath:
You need install numpy-stl (there is a wheel for this)

nallath:
I've re-added the old code, which is now used as a fallback.

thopiekar:
Great, thanks. No need for pip here 👌

nickthetait:
x20 performance increase? :open_mouth:

nallath:
Yep. I've had files that took ~140 seconds take ~7 seconds to load. The new implementation isn't the most elegant either, but it's soo much faster (and I think that the improvement is already so big, that spending a lot of time in getting it down even further is not reaaaly worth it)

UM/Mesh/MeshData.py
```diff
@@ -228,6 +228,15 @@ def reserveFaceCount(self, num_faces):
...
         self._face_count = 0
+
+
+    def setIndices(self, indices):
+        self._indices = indices
+        self._face_count = int(self._indices.size / 3)
+
+    def setVertices(self, vertices):
+        self._vertices = vertices
+        self._vertex_count = int(self._vertices.size / 3)
```
Ghostkeeper: I don't think you should divide by 3 here.

UM/Mesh/MeshData.py
```diff
@@ -228,6 +228,15 @@ def reserveFaceCount(self, num_faces):
 
         self._vertex_count = 0
         self._face_count = 0
+
+
+    def setIndices(self, indices):
+        self._indices = indices
+        self._face_count = int(self._indices.size / 3)
```
Ghostkeeper: You might want to consider resetting the normals in this function.

UM/Mesh/MeshData.py
Ghostkeeper: And you might want to consider resetting the normals and indices in this function. Up to your discretion though, because that requires you to first call `setVertices` and then `setIndices` in that order (as we are doing now).

nallath: I think we should. 
We have a matrix that looks like this: [[x0, y0, z0], [x1, y1, z0],..., [xn, yn, zn]]
The size then reports n * 3. In order to get n we need to divide by three.

thopiekar:
First we should add ```stl-numpy``` to cura-build...

thopiekar:
Can I try to add it to cura-build?

nallath:
@thopiekar; Yes please. I tried to do it, but got stuck when trying to build QT.

thopiekar:
That's strange. Yesterday I made a full build without any errors. But note
that I made these builds on Ubuntu Wily, which is known to work with
cura-build. AFAIK, we don't ship our own Python in cura-build on Linux. 😉

nallath:
Oh no i mean, when i tried to change the build script, it wouldn't let me build at all. I'm happy that you did it :)

PR #144, opened 2016-06-21
New SettingDefinitionsModel

awhiemstra:
This is a reimplementation of SettingDefinitionsModel that is both simpler and faster. It is designed to never trigger a complete model reset apart from containerId or root property changes. In addition, it fixes a number of bugs currently present in the model:

* Children-of-children are visible even if their parents are not
* Empty categories are hidden
* Filtering removes rows instead of hiding the data
* Allow setting the expanded settings so we can re-enable saving and loading of expanded categories.
* Row changes are done through insertRows/removeRows which prevents a complete model reset and only changes things that actually changed.

Additionally this paves the way for creating an "EnabledVisibilityHandler" that modifies visibility based on the enabled property.

### DIFF
--- a/UM/Qt/qml/UM/Preferences/SettingVisibilityPage.qml
+++ b/UM/Qt/qml/UM/Preferences/SettingVisibilityPage.qml
@@ -25,36 +25,6 @@ PreferencesPage
     }
     resetEnabled: true;
 
-    function updateToggleVisibleSettings()
-    {
-        var typeRoleId = settingsListView.model.roleId("type");
-        var visibleRoleId = settingsListView.model.roleId("visible");
-        var all_visible = true;
-        var all_hidden = true;
-        for(var i = 0; i < settingsListView.model.rowCount(); i++) {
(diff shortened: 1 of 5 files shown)


sedwards2009:
@nickthetait  The model referred to here is not the 3D model. The model here is part of the QML UI and the settings sidebar.

awhiemstra:
Yeah, it is the data model class that provides the structure for the sidebar view. Terminology gets kind of confusing with all these different concepts. :)

nickthetait:
Ah, thanks for explaining that.
*starts thinking about how to improve the terminology*

fieldOfView:
@nickthetait: In the code we talk in terms of "meshes" most of the time when we refer to 3d models, to reduce this confusion. In the UI, they are referred to as "objects" (which would also be confusing in the code).

nallath:
This breaks the per-object settings.

awhiemstra:
The actual change needed is really minor: add "expanded: [ "*" ]" to line 156 in PerObjectSettingsPanel.qml. I will push that after this PR is merged since it is rather silly to make an entire branch for just that change.

PR #146, opened 2016-06-24
Ignore Mylyn on Eclipse

thopiekar:


.gitignore
```diff
@@ -29,8 +29,9 @@ _eric6project/Uranium.e4t
 .cache
 *.kate-swp
 
-## Eclipse+PyDev
+## Eclipse+PyDev incl. Mylyn
 .project
+.settings/org.eclipse.mylyn.tasks.ui.prefs
```
fieldOfView: Not knowing anything about Eclipse or Mylyn, how about ignoring .settings alltogether?

thopiekar: And I don't know how other IDE's are storing their settings. It is possible that they also use .settings for their files.
To be sure it is only ignoring Mylyn I only added the new file here.

PR #147, opened 2016-06-27
Logger: fixes for logException

thopiekar:
* Don't format lines from traceback
* Directly pass args to self.log()
* Format message only once
* Only format messages when args are passed

Contributes to CURA-1445 and CURA-1771

### DIFF
--- a/UM/Logger.py
+++ b/UM/Logger.py
@@ -38,22 +38,25 @@ class Logger:
                 filename = filename.replace(path, "...")
                 continue
         address = "%s (%s [%s]): " %(filename, function.co_name, function.co_firstlineno)
+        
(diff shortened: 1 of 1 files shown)


Ghostkeeper:
Why was the "Exception: " part of logException removed?

thopiekar:
Good question. I think I've removed it to debug the problem that a %s was left. Would you mind if I write "Exception: " + message, instead of "Exception: %s" %(message)? If we do any modifications on the code in the future it would be easier to find the correct position of incorrect formatting.

thopiekar:
Done

thopiekar:
This PR was missed while working on https://github.com/Ultimaker/Cura/pull/824, which got splitted up.

PR #150, opened 2016-07-04
Support work for Material Editing

awhiemstra:
This includes a bunch of changes needed to make it possible to edit materials.

### DIFF
--- a/UM/Settings/ContainerRegistry.py
+++ b/UM/Settings/ContainerRegistry.py
@@ -345,7 +345,6 @@ class ContainerRegistry:
     def getMimeTypeForContainer(cls, container_type):
         mime_type_name = UM.Dictionary.findKey(cls.__mime_type_map, container_type)
         if mime_type_name:
-            print(mime_type_name)
             return MimeTypeDatabase.getMimeType(mime_type_name)
 
         return None
--- a/UM/Settings/ContainerRegistry.py
+++ b/UM/Settings/ContainerRegistry.py
@@ -148,6 +148,13 @@ class ContainerRegistry:
                             continue
                         except AttributeError:
                             pass
+                    if key == "read_only":
+                        try:
+                            if value != container.isReadOnly():
+                                matches_container = False
+                            continue
+                        except AttributeError:
+                            pass
                     if value != container.getMetaDataEntry(key):
                         matches_container = False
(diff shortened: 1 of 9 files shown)


nallath:
This still has merge conflics. Could you try and fix those?

awhiemstra:
Fixed

UM/Settings/ContainerRegistry.py
```diff
@@ -1,4 +1,4 @@
-# Copyright (c) 2016 Ultimaker B.V.
+    # Copyright (c) 2016 Ultimaker B.V.
```
Ghostkeeper: I tend to think you didn't intend to indent this impending entente.

UM/Settings/ContainerRegistry.py
```diff
@@ -146,6 +148,13 @@ def findContainers(self, container_type = None, **kwargs):
                             continue
                         except AttributeError:
                             pass
+                    if key == "read_only":
+                        try:
+                            if value != container.isReadOnly():
+                                matches_container = False
+                            continue
+                        except AttributeError:
+                            pass
```
Ghostkeeper: I'm starting to think that instead of using `continue` everywhere, we should just use `elif`s.

UM/Settings/Models/ContainerPropertyProvider.py
```diff
@@ -0,0 +1,150 @@
...
+        if property_name not in self._watched_properties:
+            return
+
+        value = self._getPropertyValue(property_name)
+
+        if self._property_values[property_name] != value:
+            self._property_values[property_name] = value
+            self.propertiesChanged.emit()
+
+    def _update(self, container = None):
```
Ghostkeeper: Container parameter is unused.

UM/Settings/ContainerRegistry.py
awhiemstra: We probably should clean this up at some point, it is pretty ugly right now.

sedwards2009: I'm personally already developing PTSD just thinking about all of the nasty bugs which will silently escaping logging thanks to the use of `except AttributeError` and `except TypeError`. (This is **terrible** use of `except`.)

PR #151, opened 2016-07-07
[CURA-1502] Better implementation of Undo bug.

jackha:
CURA-1502

### DIFF
--- a/UM/Operations/OperationStack.py
+++ b/UM/Operations/OperationStack.py
@@ -17,8 +17,8 @@ class OperationStack():
         self._current_index = -1 #Index of the most recently executed operation.
         self._lock = threading.Lock() #Lock to make sure only one thread can modify the operation stack at a time.
 
-        # Only merge operations when we're in the same operation. So, whenever an operation is started
-        # or ended, we have to set _merge_operations to False so the original behaviour is in place.
+        # The merge behaviour must only occur when an operation is in the middle of a user action.
+        # So, whenever an operation is started or ended, we do not want this auto-merge.
         self._merge_operations = False
         self._controller = UM.Application.getInstance().getController()  # Circular dependency blah
         self._controller.toolOperationStarted.connect(self._onToolOperationStarted)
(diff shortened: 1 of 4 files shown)


UM/Operations/OperationStack.py
```diff
@@ -16,6 +17,17 @@ def __init__(self):
...
+        self._merge_operations = False  #Only merge operations when we're in the same operation.
+        self._controller = UM.Application.getInstance().getController()  # Circular dependency blah
+        self._controller.toolOperationStarted.connect(self._onToolOperationStarted)
+        self._controller.toolOperationStopped.connect(self._onToolOperationStopped)
+
+    def _onToolOperationStarted(self, tool):
+        self._merge_operations = False
+
+    def _onToolOperationStopped(self, tool):
+        self._merge_operations = False
```
nallath: Soo, never merge? Why are they set to false in both cases?

fieldOfView: The first one should be True, I guess.
Personally I would name the variable ```self._operation_active``` or something similar, so when you use it in line 116 you know _why_ the operation is being merged.

nickthetait:
Could you describe a little bit what the undo bug was?

fieldOfView:
The undo function remembers all rotate positions from the model, even when the mouse is not click-released.

Reproduce steps:
1) Load any model.
2) Rotate the model 120°
3) Hit CTRL+Z.
4) Rotate the model without snap rotation, for 34°.
4) Hit CTRL+Z.

Was the bug reproducible?
100%

Expected result:
After hitting CTRL+Z, the model is reset to the original rotation, 0°. After rotating 34°, the model should again be reset to 0°.

Actual result:
After hitting CTRL+Z, the model is rotated to the previous snap rotation, 105°. When using non-snap rotation, the model is rotated back 1° at a time.

fieldOfView:
Actually, it was a remaining aspect of the above; a rotation (or indeed scale or move) operation would be chopped into parts if the user paused moving the mouse while still holding the mouse button.

jackha:
Added some more comments. The merge behaviour must only occur when an operation is in the middle of a user action.

nickthetait:
Thanks for explaining fieldofview. Sounds like a seriously annoying bug!

PR #152, opened 2016-07-08
CURA-1445: Adding getDurationInISOFormat

thopiekar:
Currently we are not passing the print duration in ISO format.

This will make it possible to get it directly from our print_information variable in SliceInfo plugin.

### DIFF
--- a/UM/Qt/Duration.py
+++ b/UM/Qt/Duration.py
@@ -100,4 +100,8 @@ class Duration(QObject):
                 return i18n_catalog.i18nc("@label Minutes only duration format, {0} is minutes", "{0} minutes", self._minutes)
 
         return ""
+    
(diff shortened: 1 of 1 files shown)


fieldOfView:
Couldn't this be done by adding a format to Duration.getDisplayString()?

thopiekar:
Good idea. I thought Duration.getDisplayString() is meant to be used for the UX only.
Shouldn't we call it better DurationFormat.Format.ISO8601
- https://en.wikipedia.org/wiki/ISO_8601

As this format is an standard I would move it on top of "short" and "long". Any reason why I shouldn't call it "ISO8601 = 0"? Does it have to start with "= 1"?

fieldOfView:
Why was this merged when there was ongoing discussion about the implementation?

PR #155, opened 2016-07-30
Convenience functions to convert a UM.Math.Color to/from a hex-string

fieldOfView:
RGB-colors are commonly represented as "web-friendly" #RRGGBB strings (eg #ffffff for white). This PR adds convenience functions to the UM.Math.Color class to convert to/from this format. Usefull in cases where colors are specified in configurations.

### DIFF
--- a/UM/Math/Color.py
+++ b/UM/Math/Color.py
@@ -67,3 +67,25 @@ class Color:
             (value & 0x000000ff) >> 0,
             (value & 0xff000000) >> 24
         )
+
+    ##  Returns a new Color constructed from a 7-character string "#RRGGBB" format.
+    #
+    #   \param value A 7-character string representing a color in "#RRGGBB" format.
+    #   \return A Color constructed from the components of value.
+    @staticmethod
+    def fromRGBString(value):
+        return Color(
+            int(value[1:3], 16) / 255,
+            int(value[3:5], 16) / 255,
+            int(value[5:7], 16) / 255,
+            1.0
+        )
+
+    ##  Returns a 7-character string in "#RRGGBB" format representing the color.
+    #
+    #   \return A 7-character string representing a color in "#RRGGBB" format.
+    def toRGBString(self):
+        value = ((int(self._r * 255) & 255) << 16) + \
+                ((int(self._g * 255) & 255) << 8) + \
+                (int(self._b * 255) & 255)
+        return "#%s" % hex(value)[2:]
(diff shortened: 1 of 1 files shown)


Ghostkeeper:
Seems useful. Passes my code review but I haven't the resources to test it at the moment.

fieldOfView:
It is low priority for now, dont spend too much time on it

UM/Math/Color.py
```diff
@@ -67,3 +67,25 @@ def fromARGB(value):
             (value & 0x000000ff) >> 0,
             (value & 0xff000000) >> 24
         )
+
+    ##  Returns a new Color constructed from a 7-character string "#RRGGBB" format.
+    #
+    #   \param value A 7-character string representing a color in "#RRGGBB" format.
+    #   \return A Color constructed from the components of value.
+    @staticmethod
+    def fromRGBString(value):
```
awhiemstra: It would be nice if you could turn this into a "fromRGBAString" method that detects whether the string is RGB or RGBA. That way we do not have to add an extra function for that later.

Ghostkeeper: How about `fromHexString`?

fieldOfView: Sure, I could do that. I could also make it robust against 3 or 4 digit colors (eg #fff instead of #ffffff for white). I just did not want to add unnecessary complexity.

awhiemstra: The thing is, UM.Math.Color is meant as an RGB_A_ color value. The API should reflect that, which is why I would like to make sure factory functions like this support the full class, not just a special case.

fieldOfView: Argh! QML supports #aarrggbb (http://doc.qt.io/qt-5/qml-color.html), but CSS is going towards #rrggbbaa (https://drafts.csswg.org/css-color/#hex-notation). I think it makes most sense to opt to support QML over CSS4 here.

fieldOfView:
This now takes alpha into account, using the suggested fromHexString/toHexString. toHexString has an optional argument to include the alpha (defaulting to False because rrggbb colors are that much more prevalent). I have chosen to go the qml-route of #aarrggbb instead of the css4-route of #rrggbbaa.

awhiemstra:
To be fair, I have always disliked that Qt uses ARGB. OpenGL and any other drawing API uses RGBA. I once asked the lead developer of Qt if they were finally going to switch to RGBA and he said "no because it is a simple transformation to convert to RGBA" which seems silly to me since no conversion is better than a conversion. But well, the main reason for it is legacy. Using ARGB in this case is fine since it is mainly meant for interaction with QML.

nallath:
Soo, should we merge this then? I see no reason not to.

awhiemstra:
Yes this can be merged.

PR #157, opened 2016-08-10
Update UI texts to be compliant to Ultimaker dictionary

fieldOfView:
A number of issues were found where the UI uses terms that do not comply to the terms set in the "Ultimaker dictionary".

CURA-2053


fieldOfView:
On second (third?) thought, the changes in this PR are so minor that they don't merit a PR.

PR #160, opened 2016-08-14
[CURA-1696] Move translation targets to a separate module

thopiekar:
Moves all macros and also the if clause, which looks for Gettext.


thopiekar:
Also belongs here: https://github.com/Ultimaker/Uranium/commit/cff519a897611b7b578d260af50e7b46d8434262

PR #164, opened 2016-09-10
Added Polish translation

MarPiRK:
Also changed formatting of long representation of print time – not it uses plural forms. I'm not sure if it's the best way to do it. Feel free to change it. Unfortunately it required updating `pot` file, so it's a little bit messy.

### DIFF
--- a/UM/Qt/Duration.py
+++ b/UM/Qt/Duration.py
@@ -97,11 +97,11 @@ class Duration(QObject):
                 return i18n_catalog.i18nc("@label Short hours-minutes format. {0} is hours, {1} is minutes", "{0:0>2}h {1:0>2}min", self._hours, self._minutes)
         elif display_format == DurationFormat.Format.Long:
             if self._days > 0:
-                return i18n_catalog.i18nc("@label Days-hours-minutes duration format. {0} is days, {1} is hours, {2} is minutes", "{0} days {1} hours {2} minutes", self._days, self._hours, self._minutes)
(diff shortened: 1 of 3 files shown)


Ghostkeeper:
It's almost as if you knew that we're entering a string freeze on the next version this weekend... Lucky!

I understand that the changes are because you re-generated the strings from our code. But that makes me wonder why it still says that the files originate from my home folder :stuck_out_tongue:

MarPiRK:
Hah, got that timing 😂

About paths I just wanted to change least lines as possible. Turns out that it didn't really help, so whatever.

Ghostkeeper:
What I do need to ask though is what application you intended to write these translations for. This is only Uranium, and the translations will be incomplete if you intend to translate Cura or NinjaKittens2.

MarPiRK:
Sure enough, I'm planning to translate entire Cura and I'm on the way, but as you might know it's sth like 1000 sentences. Uranium is the first step.

Ghostkeeper:
I know. Don't underestimate that FDMPrinter file...

MarPiRK:
So will this be merged?

thopiekar:
So the answer is yes :wink:

MarPiRK:
Thanks :smile:

Ghostkeeper:
It's not active yet in any application though.

thopiekar:
@MarPiRK, according to https://translations.launchpad.net/uranium/trunk/+pots/uranium/pl/+translate there are missing translations.
So if you like you can continue translating Uranium here or please update your translation via Git.
Additionally if you like to use launchpad to translate Cura, just create a PR with empty strings.

Regards

PR #165, opened 2016-10-04
Improve Performance of SettingPropertyProvider and property changes

awhiemstra:
This contains two changes: Replacing the dict in SettingPropertyProvider with QQmlPropertyMap, which is a dict-like object that allows QML bindings on individual keys in the dictionary so we do not constantly need to update the entire dict when a single property changes.

The other change is to collect all property change signals that should be handled by ContainerStack and emit them together. This reduces the amount of duplicate signals emitted by ContainerStack and allows us to emit a single signal that references all properties that have changed.

These changes combined reduce the time needed to handle toggling "Enable Support" from 140ms down to 40ms, according to the QML profiler.

### DIFF
--- a/UM/Settings/ContainerStack.py
+++ b/UM/Settings/ContainerStack.py
@@ -503,7 +503,7 @@ class ContainerStack(ContainerInterface.ContainerInterface, PluginObject):
         self._property_changes[key].add(property_name)
 
         if not self._emit_property_changed_queued:
-            Signal._app.callLater(self._emitCollectedPropertyChanges)
+            UM.Settings.ContainerRegistry.getApplication().callLater(self._emitCollectedPropertyChanges)
             self._emit_property_changed_queued = True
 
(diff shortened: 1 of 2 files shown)


Ghostkeeper:
After being reviewed by Nallath and me, we've determined in discussion with the test team that the performance issues are currently not pressing enough to risk this merge. The risk involved in the merge is primarily in how the propertieschanged signals are not immediately emitted, while some components might expect that. If some components expect that, we'd currently have a race condition on our hands (since we make no guarantees as to which order the events are handled in), but still...

On our slowest computer (Carlo's laptop), we eyeballed that it currently takes about 1.5s to enable or disable support. This change is expected to drop that to about 0.4s. However, 1.5s for the worst change is not unacceptable. We're probably going to merge this in the beginning of the development for v2.4.

Ghostkeeper:
I've made an issue in our internal tracker for the future sprint of 2.4 to make sure we merge this: CURA-2547.

PR #168, opened 2016-10-18
Some useful decorators to use when debugging.

sedwards2009:
Over the last few months I've accumulated a bunch of decorators which I use for debugging. Time to share.

* timeIt - Simply times and logs the execute duration of a method.
* logThread - Logs the name of current thread.
* spy - Logs the arguments and return value of a method.
* logStack - Logs the execution stack when the method is called.

### DIFF
--- a/UM/Logger.py
+++ b/UM/Logger.py
@@ -2,8 +2,11 @@
 # Uranium is released under the terms of the AGPLv3 or higher.
 
 import sys
-import traceback
 import inspect
+from functools import wraps
+import time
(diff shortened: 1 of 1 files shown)


ChrisTerBeke:
This seems useful, why was it never merged?

thopiekar:
I agree. Took the time and resolved the conflicts in the imports and sorted them by alphabet.

ChrisTerBeke:
Devs: CURA-4664.

PR #170, opened 2016-10-20
CMake: Using add_test for registering tests

thopiekar:
On JIRA it was discussed to use an alias instead of renaming our target called "tests" to "test".
When trying to define a target called "test", CMake warns about calling a custom target like this. I read in CMake docs how they expect to have test registered and found add_test for that.
CMake then provides a target called "test", but adding a custom target called "tests" and make it depend on "test" doesn't work for a reason. The error message says that "test" is not defined (even if I have the statement called after "enable_testing()"!).
Finally I kept it like that as we are already using "test" in CuraEngine.

Contributes to CURA-1511

PS: This one is for a future release. Just coded around it for fun while I'm on vacation..

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -1,6 +1,6 @@
 project(uranium NONE)
 
-cmake_minimum_required(VERSION 2.8.12)
+cmake_minimum_required(VERSION 3.2.0)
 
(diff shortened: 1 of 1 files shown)


thopiekar:
Ooopss.. Wrong JIRA ticket..

PR #173, opened 2016-10-23
CMake: Remove hardcoded projectname

thopiekar:
While updating Cura I noticed that translation files were still generated for Uranium (in a clean Cura build).
The reason for this is a hardcoded path, which needs to be replaced with CMake's PROJECT_NAME.

I remember there were discussions on JIRA about odd behaviour while checking our translations.
This might be one of the reasons...

Not sure which JIRA issue it really belongs to, but I set it as:
Contributes to CURA-570


thopiekar:
https://launchpadlibrarian.net/290538262/buildlog_ubuntu-xenial-amd64.cura_1%3A2.3.99-master~201610231132~rev1460~pkg68~ubuntu16.04.1_BUILDING.txt.gz

Ghostkeeper:
Frightening how many references to Cura are in that file, actually.

Wouldn't PROJECT_NAME be equal to Cura/Argus/Ninjakittens while this script is being used to define targets for their translations?

thopiekar:
@Ghostkeeper, correct it will use the project name as it was set in the CMakeLists.txt.
E.g. https://github.com/Ultimaker/Uranium/blob/b93fe0c7a7e1ec6d7c610fc8f0701697b0a8ab21/CMakeLists.txt#L1

cmake/UraniumTranslationTools.cmake
```diff
@@ -108,6 +108,6 @@ if(GETTEXT_FOUND)
         TARGETS_FOR_PO_FILES(${language})
         TARGETS_FOR_MO_FILES(${language})
     endforeach()
-    install(DIRECTORY ${CMAKE_BINARY_DIR}/resources DESTINATION ${CMAKE_INSTALL_DATADIR}/uranium/)
+    install(DIRECTORY ${CMAKE_BINARY_DIR}/resources DESTINATION ${CMAKE_INSTALL_DATADIR}/${PROJECT_NAME}/)
```
awhiemstra: Shouldn't this be CMAKE_PROJECT_NAME ?

Additionally, it might be better to use macro arguments for things like this.

thopiekar: No.

1. I'm already using it above and it works fine
2. As far as I understand https://cmake.org/Wiki/CMake_Useful_Variables "CMAKE_PROJECT_NAME" would be "cura-build" if we build Uranium inside cura-build, as cura-build would be the "the top level project". Isn't it?

awhiemstra: No, since Uranium in cura-build is built by calling an external CMake, not included in the main project. But I see now that CMake also defines PROJECT_NAME... Yay consistency!

thopiekar: Hmm, ok, if you like I can also rename all variables to CMAKE_PROJECT_NAME of course 😉

thopiekar: Ok, let it unmerged then. I'll add a TODO somewhere and report it on JIRA, so we can plan there, when we will be able to improve the macro with an option.
When finished I'll merge it on my own :wink: 

Thanks!

PR #176, opened 2016-10-30
MeshFileHandler: Care about type of "mesh_handler"

thopiekar:
Misunderstood the syntax how to declare multiple data formats in my
plugin's `__init__.py`.
However this is what I've left at the end, which could be helpful if
someone mixes up a list with a dictionary, for example.

### DIFF
--- a/UM/Mesh/MeshFileHandler.py
+++ b/UM/Mesh/MeshFileHandler.py
@@ -119,11 +119,15 @@ class MeshFileHandler(object):
         meta_data = PluginRegistry.getInstance().getAllMetaData(filter = {"mesh_reader": {}}, active_only = True)
         for entry in meta_data:
             if "mesh_reader" in entry:
-                for input_type in entry["mesh_reader"]:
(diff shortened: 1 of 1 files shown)


Ghostkeeper:
I sort of disagree with this change. It doesn't need to be a list. We might make it a set at some point. Or a generator. Anything that's iterable could go in there.

thopiekar:
What do you think about the solution now?

Ghostkeeper:
The current solution makes no real change. The `for` loop already checks for `__iter__` and gives an appropriate error when it's not iterable. `dict`s also have `__iter__`.

Checking for `__next__` will always be `False`, since the `__next__` method is not a method of the list or dictionary itself, but of the iterator returned by `__iter__`.

I think that no change is necessary here; this part already has proper ducktyping.

thopiekar:
Ok, so let's close it then..

Mit freundlichen Grüßen/With kind regards
- Thomas Karl Pietrowski

PR #179, opened 2016-10-31
Add "All Supported Types" option to profile import dialog

fieldOfView:
This PR adds a "All Supported Types" filter to the import profile dialog as the default option. This way the functionality to import gcode and legacy profiles are no longer hidden from the user; as it is now the user must manually switch filetypes to even see gcode or legacy profiles. Especially on OSX this is quite hidden.

### DIFF
--- a/UM/Settings/Models/InstanceContainersModel.py
+++ b/UM/Settings/Models/InstanceContainersModel.py
@@ -4,7 +4,7 @@ from UM.Qt.ListModel import ListModel
 
 from PyQt5.QtCore import pyqtProperty, Qt, pyqtSignal, pyqtSlot, QUrl
 
-from UM.PluginRegistry import PluginRegistry #For getting the possible profile readers/writers.
+from UM.PluginRegistry import PluginRegistry #For getting the possible profile readers and writers.
 from UM.Settings.ContainerRegistry import ContainerRegistry
 from UM.Settings.InstanceContainer import InstanceContainer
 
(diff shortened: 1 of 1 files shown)


nickthetait:
Small change that should make a huge quality of life improvement :v:

Ghostkeeper:
Looks nice! But the profile export now also has an "All supported types" option for some reason. And when I try to use it, I get the following error:

    2016-11-01 11:09:18,549 - ERROR - .../cura/Settings/CuraContainerRegistry.py (exportProfile [67]): Failed to export profile to /home/ruben/test.curaprofile: 'NoneType' object has no attribute 'write'

I think we should change the export dialogue for this. It should not be using the types we use for importing.

awhiemstra:
Propbably because getFileNameFilters is used both by import and export. Probably best to split the function in two to handle that.

fieldOfView:
I'll add a bool argument "single_filters_only" to getFileNameFilters later today.

fieldOfView:
How about this approach, where instead of adding a new method or argument, the current method and argument are used? 
81afe61a1ed5ea0fcb9492aa3f8c131f42129ef8
Or do we foresee a future change that may break this?

awhiemstra:
Yeah, that works too. Better than using a boolean trap. Please do expand the method documentation though.

Ghostkeeper:
I actually like the boolean trap better than opaquely checking if there is the string "_reader" in the input type name. Or instead of a boolean trap a "string trap" by way of a simplified enum. But it's all down to opinion, of course.

fieldOfView:
Does this need changes, or can it be merged?

nallath:
This fell of the radar. I will have a look at it right away. Sorry bout that!

PR #180, opened 2016-11-02
Fix readonly state of profiles on Windows

fieldOfView:
Change config storage path to use back-slashes on Windows

Python magically fixes mixed slashes and backslashes in paths on Windows. However the storage path gets compared to the full path of files when loading configs to determine readonly state, and this comparison fails if the config storage path uses forward-slashes. The result is that all config files are treated as readonly, including quality_changes and user profiles. This is only really visible in the Profiles preference pane; after restarting Cura all profiles are shown as protected profiles.

Note that this only seems to be necessary in master; I have not yet found out when the readonly state broke in master compared to 2.3.

### DIFF
--- a/UM/Resources.py
+++ b/UM/Resources.py
@@ -264,7 +264,7 @@ class Resources:
     @classmethod
     def __initializeStoragePaths(cls):
         if platform.system() == "Windows":
-            cls.__config_storage_path = os.path.join(os.path.expanduser("~/AppData/Local/"), cls.ApplicationIdentifier)
+            cls.__config_storage_path = os.path.join(os.path.expanduser("~\\AppData\\Local\\"), cls.ApplicationIdentifier)
         elif platform.system() == "Darwin":
             cls.__config_storage_path = os.path.join(os.path.expanduser("~/Library/Application Support"), cls.ApplicationIdentifier)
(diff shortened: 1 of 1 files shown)


Ghostkeeper:
I'm assigning sedwards2009 for this, since he's the only developer on Windows who is in this week.

fieldOfView:
Is there anything I can do/provide to speed this along?

Ghostkeeper:
Not really. I'll e-mail him, because I think he doesn't have his notifications enabled.

sedwards2009:
@fieldOfView  Where can I find the code which compares config paths? I suspect that a `os.path.normpath()` needs to be applied in a few places.

The pull request doesn't seem to break anything from my little bit of testing. So I'll merge it.

fieldOfView:
See https://github.com/Ultimaker/Uranium/blob/master/UM/Settings/ContainerRegistry.py#L223

sedwards2009:
@fieldOfView thanks. I can put a `normpath()` or two in there and then the comparison should be ok. The paths probably need to be made absolute too before comparison.

sedwards2009:
Actually `os.path.realpath()` is the one I'm looking for.

PR #181, opened 2016-11-06
CURA-1764: Appending environment variables with set_tests_properties

thopiekar:
As discussed on JIRA the best way to add our environment variables is to pass them via "set_tests_properties". Of course, this depends on #171, which begins to use the set of commands provided by CMake to declare tests.

However, there seems to be a problem in the test suite, therefore it isn't working  completely now.

Fixes: #171 
(Includes: #171 )
Contributes to CURA-1697
Contributes to CURA-1764

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -12,16 +12,21 @@ include(GNUInstallDirs)
 
 find_package(PythonInterp 3.4.0 REQUIRED)
 
-# Checks using pylint
+# # Checks using pylint
 # Note that we use exit 0 here to not mark the build as a failure on check failure
 # In addition, the specified pylint configuration uses the spellchecker plugin. This required python-enchant to be installed.
 add_custom_target(check)
 add_custom_command(TARGET check POST_BUILD COMMAND "PYTHONPATH=${CMAKE_SOURCE_DIR}" ${PYTHON_EXECUTABLE} -m pylint --rcfile=${CMAKE_SOURCE_DIR}/pylint.cfg UM --msg-template=\"{path}:{line}: [{msg_id}({symbol}), {obj}] {msg}\" > ${CMAKE_BINARY_DIR}/pylint.log || exit 0 WORKING_DIRECTORY ${CMAKE_SOURCE_DIR})
 
-# Tests
-# Note that we use exit 0 here to not mark the build as a failure on test failure
-add_test(NAME pytest COMMAND "LANG=C" "PYTHONPATH=${CMAKE_SOURCE_DIR}" ${PYTHON_EXECUTABLE} -m pytest -r a --junitxml=${CMAKE_BINARY_DIR}/junit.xml ${CMAKE_SOURCE_DIR} || exit 0)
+# # Tests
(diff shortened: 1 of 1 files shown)


CMakeLists.txt
```diff
@@ -1,6 +1,6 @@
 project(uranium NONE)
 
-cmake_minimum_required(VERSION 2.8.12)
+cmake_minimum_required(VERSION 3.2.0)
```
awhiemstra: Our CentOS build VM is using CMake 2.8.12. This change would break that VM. As far as I can tell, there is no need to bump the version requirement.

CMakeLists.txt
```diff
@@ -12,16 +12,21 @@ include(GNUInstallDirs)
...
-# Tests
-# Note that we use exit 0 here to not mark the build as a failure on test failure
-add_custom_target(tests)
-add_custom_command(TARGET tests POST_BUILD COMMAND "LANG=C" "PYTHONPATH=${CMAKE_SOURCE_DIR}" ${PYTHON_EXECUTABLE} -m pytest -r a --junitxml=${CMAKE_BINARY_DIR}/junit.xml ${CMAKE_SOURCE_DIR} || exit 0)
+# # Tests
+enable_testing()
+add_test(NAME pytest COMMAND ${PYTHON_EXECUTABLE} -m pytest -r a --junitxml=${CMAKE_BINARY_DIR}/junit.xml ${CMAKE_SOURCE_DIR})
+# Using untranslated strings on Linux
+if(UNIX AND NOT APPLE) 
+    set_tests_properties(pytest PROPERTIES ENVIRONMENT LANG=de_DE.UTF-8)
```
awhiemstra: What is this change for? Specifically, why the change from "C" to "de_DE.UTF-8"?

CMakeLists.txt
```diff
@@ -12,16 +12,21 @@ include(GNUInstallDirs)
...
+endif()
+# Adding PYTHONPATH on every OS
+set_tests_properties(pytest PROPERTIES ENVIRONMENT PYTHONPATH=${CMAKE_SOURCE_DIR})
```
awhiemstra: Please add the following to provide backward compatibility:

`add_custom_target(tests COMMAND ${CMAKE_COMMAND} --build ${CMAKE_BINARY_DIR} --target test COMMENT "Notice: tests target is deprecated, please use the test target")`

This will run the test target when calling "make tests".

CMakeLists.txt
```diff
@@ -12,16 +12,21 @@ include(GNUInstallDirs)
 
 find_package(PythonInterp 3.4.0 REQUIRED)
 
-# Checks using pylint
+# # Checks using pylint
```
awhiemstra: Nitpick: Double #

CMakeLists.txt
thopiekar: Good question 👀 
I think I forgot to reset it before pushing the PR here.

thopiekar:
@awhiemstra Done

PR #183, opened 2016-11-08
Allow adding a container at a specified index in the stack

fieldOfView:
This PR adds an optional parameter to ContainerStack.addContainer that makes it possible to add a container at the specified index in an existing stack, instead of always at the top. If the parameter is not specified, the container is added to the top of the stack as before the PR.

### DIFF
--- a/UM/Settings/ContainerStack.py
+++ b/UM/Settings/ContainerStack.py
@@ -401,7 +401,6 @@ class ContainerStack(ContainerInterface.ContainerInterface, PluginObject):
     def insertContainer(self, index, container):
         if container is self:
             raise Exception("Unable to add stack to itself.")
-            return
 
         container.propertyChanged.connect(self._collectPropertyChanges)
         self._containers.insert(index, container)
--- a/UM/Settings/ContainerStack.py
+++ b/UM/Settings/ContainerStack.py
@@ -390,14 +390,22 @@ class ContainerStack(ContainerInterface.ContainerInterface, PluginObject):
     ##  Add a container to the top of the stack.
     #
     #   \param container The container to add to the stack.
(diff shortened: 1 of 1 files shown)


awhiemstra:
Rather than add an extra parameter, please create an extra method called insertContainer that expects an index and a container. This makes the API more straightforward.

fieldOfView:
Good point. Shall I then change addContainer to a stub that uses insertContainer with index=0?
I'll also add error checking for illegal indexes.

UM/Settings/ContainerStack.py
```diff
@@ -391,12 +391,21 @@ def findContainer(self, criteria = None, container_type = None, **kwargs):
...
+
+    ##  Insert a container into the stack.
+    #
+    #   \param index \type{int} The index of to insert the container at.
+    #          A negative index counts from the bottom
+    #   \param container The container to add to the stack.
+    def insertContainer(self, index, container):
+        if container is self:
             raise Exception("Unable to add stack to itself.")
+            return
```
Ghostkeeper: This return statement is dead code.

fieldOfView: Is it, really? I am not sure, because uncaught exceptions are handled by CrashHandler.py, and CrashHandler.py for most exceptions just logs a "Critical" error and continues: https://github.com/Ultimaker/Cura/blob/master/cura/CrashHandler.py#L34
Note that I think that that behavior is not right; a Critical error should stop program execution. But that's a whole different issue ;-)

awhiemstra: An exception will still exit the running code. What it does not do, is abort the main loop. So yes, the return is superfluous.

fieldOfView: Fixed it here: ff1d0e5

PR #184, opened 2016-11-08
Add a system-defined monospaced font to the theme

fieldOfView:
A fixed width font is useful for displaying code, eg start/end-gcode in Cura.


Ghostkeeper:
This works. If I add `font: UM.Theme.getFont("fixed")` to the properties of the start and end g-code fields of the Custom FDM Printer, it displays the g-code in a fixed-width font. Only that font itself can't be added by this pull request of course, since that is in Cura.

Ghostkeeper:
I wonder how this holds up in other operating systems. But I guess we'll find out.

PR #185, opened 2016-11-17
Fix logging filenames

awhiemstra:
This PR changes logging to log the full classpath of the class that wants to print the logging, rather than using the file name. This avoids printing filenames of build systems and fixes logging when using the AppImage or other tools that modify paths.

In addition, this adds support for using str.format() for logging messages.

### DIFF
--- a/UM/Logger.py
+++ b/UM/Logger.py
@@ -26,9 +26,18 @@ class Logger:
         return cls.__loggers
 
     ##  Send a message of certain type to all loggers to be handled.
+    #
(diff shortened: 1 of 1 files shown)


UM/Logger.py
```diff
@@ -30,21 +30,22 @@ def getLoggers(cls):
     #   \param message \type{string} containing message to be logged
     #   \param *args \type{list} List of variables to be added to the message.
```
Ghostkeeper: You forgot to update the documentation.

PR #188, opened 2016-11-26
UM.Application: Remove Unity workaround

thopiekar:
Just found this workaround between regular code.
I would prefer to move the workaround to all the other workarounds at cura_app.py.

### DIFF
--- a/UM/Application.py
+++ b/UM/Application.py
@@ -44,8 +44,6 @@ class Application():
         self._version = version
         self._buildtype = buildtype
 
-        os.putenv("UBUNTU_MENUPROXY", "0")  # For Ubuntu Unity this makes Qt use its own menu bar rather than pass it on to Unity.
-
         Signal._app = self
         Resources.ApplicationIdentifier = name
         i18nCatalog.setApplication(self)


Ghostkeeper:
I need this workaround :(
If it is preferred to move the workaround to similar code, don't just delete it but add it back among the other code too!

Ghostkeeper:
Ah, I see it was added back in a different pull request. Sorry!

It would've made sense to make this one pull request though. One should not be accepted without the other. I'll test both changes and see if the behaviour is still correct.

Ghostkeeper:
Ah, I see that the code was added back in the Cura repository rather than the Uranium repository.

This would break the other applications that depend on Uranium though. This workaround is not specific to Cura.

thopiekar:
But I would like to collect all work at one place.
I personally don't like to have all the hacks flying around. 😒
This one for example is not marked or whatsoever.
Maybe we should decide to move all workarounds to Uranium then? The NVidia workaround should also be useful for Uranium, too.

Ghostkeeper:
I agree that the NVidia workaround should've been put in Uranium.

We may still have workarounds that are specific to Cura though. For instance if there's anything necessary to make USB printing work. I think we should strive to put the workarounds in the modules that concern them: USB printing workarounds in the USB printing plug-in, rendering workarounds in the rendering code, Unity GUI workarounds in the Qt code, etc.

PR #189, opened 2016-11-27
CURA-1921: Adding command-line option to remove user data

thopiekar:
Adds command-line options: "--mrproper" and "--all-users".
"--all-users" can be optionally used for other tasks in the future.
E.g. shutting down all running Cura/Uranium instances.

The prints inside the code are needed since the _mrproper function is
called very early, when UM.Logger is not set up.
I'm personally not very happy about the solution in _mrproper to use
hardcoded pattern, but I guess this is the only way to remove
everything.
Alternatively those patterns can be moved into UM.Resources, but that
needs to be discussed.

Contributes to CURA-1921

### DIFF
--- a/UM/Application.py
+++ b/UM/Application.py
@@ -20,6 +20,8 @@ from UM.i18n import i18nCatalog
 from UM.Workspace.WorkspaceFileHandler import WorkspaceFileHandler
 
 import UM.Settings
+import shutil
+from UM.Platform import Platform
 
 ##  Central object responsible for running the main event loop and creating other central objects.
 #
@@ -58,6 +60,13 @@ class Application():
             Resources.addSearchPath(os.path.join(os.path.abspath(os.path.dirname(__file__)), "..", "resources"))
 
         self._main_thread = threading.current_thread()
+
+        self._parsed_command_line = None
+        self.parseCommandLine()
+
+        if self.getCommandLineOption("mrproper", False):
+            self._mrproper(not self.getCommandLineOption("all-users", False))
+            sys.exit(0)
 
         super().__init__()  # Call super to make multiple inheritance work.
 
@@ -110,9 +119,6 @@ class Application():
         self._plugin_registry.setApplication(self)
 
         UM.Settings.ContainerRegistry.setApplication(self)
-
-        self._parsed_command_line = None
-        self.parseCommandLine()
 
         self._visible_messages = []
         self._message_lock = threading.Lock()
@@ -320,7 +326,9 @@ class Application():
 
     def parseCommandLine(self):
         parser = argparse.ArgumentParser(prog = self.getApplicationName()) #pylint: disable=bad-whitespace
-        parser.add_argument("--version", action="version", version="%(prog)s {0}".format(self.getVersion()))
+        parser.add_argument("--version",
+                            action="version",
+                            version="%(prog)s {0}".format(self.getVersion()))
         parser.add_argument("--external-backend",
                             dest="external-backend",
                             action="store_true", default=False,
@@ -333,7 +341,8 @@ class Application():
     #
     #   \param parser \type{argparse.ArgumentParser} The parser that will parse the command line.
     def addCommandLineOptions(self, parser):
-        pass
+        parser.add_argument("--mrproper", action="store_true", dest="mrproper", default=False, help="Remove user data")
+        parser.add_argument("--all-users", action="store_true", dest="all-users", default=False, help="Allow to remove user data from all users")
 
     def addExtension(self, extension):
         self._extensions.append(extension)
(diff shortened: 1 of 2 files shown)


Ghostkeeper:
What does "mrproper" stand for?

fieldOfView:
I'm guessing:
mr _clean_logo
'Mister Proper' in german.

thopiekar:
Correct, but I had something different in mind.
Don't know whether you ever build the Linux kernel from source, but there is also a "mrproper" command to clean your sources. Compared to "make clean", "make mrproper" removes also your kernel configuration.
So I reserved "clean" here, which could be used in the future to remove files, which are e.g. not "deserializable", etc.

ChrisTerBeke:
Do we still find this useful enough to fix the conflicts and merge it?

thopiekar:
Afaik, it was some WIP work and never working fluently. Since we want to core-ify Cura, I guess we want it back or implement it in a different way.

Ghostkeeper:
It would be kind of useful for developing and also for helping people that have such bugs that they must do a factory reset...

Ghostkeeper:
We'll discuss it internally. Developers, see issue CURA-4711.

LipuFei:
I think a command line argument for setting the cura directory to use would be more useful. Then, you can run multiple Curas in parallel (e.g., running multiple Curas to accelerate testing). A shortcut for removing the directories is not necessarily that convenient.

alekseisasin:
@LipuFei , I think removing from all cura user data at once is fine. I see that this branch has conflicts @thopiekar  can you check this conflicts. I see that you define in code "\\AppData\\Local\\", maybe it is better to use Resources folder to retrieve folders, then we will be sure that folders are identical if later we diced to make some changes.

fieldOfView:
I still have an issue with the obscure "meister proper" reference (but I still like you @thopiekar). 
Also the "--all-user" does not really work for all users; it only removes the data for the current user and some local caches (but still only for the current user.

Appesteijn:
Is this functionality now covered by the crash dialog?

LipuFei:
@Appesteijn This is more like a dev/debug tool to make start cura clean easier.

thopiekar:
Yes, it was meant to be used like that. Once this would have worked (but was not finished) it should also be integrated in the uninstaller of Cura.
So whenever you uninstall Cura in Windows there should be something like a checkbox, whether the user data should also be removed or not.
In the days of this ticket we haven't used plugins that much, but I guess nowadays it would be nice to be able to clean up everything, which has been installed and downloaded while using Cura.

UM/Application.py
```diff
@@ -348,4 +357,57 @@ def getInstallPrefix():
...
+
+        else:
+            if Platform.isLinux() or Platform.isOSX() or Platform.isWindows():
+                path_list += [Resources.getConfigStoragePath(),
+                              Resources.getDataStoragePath(),
+                              Resources.getStoragePathForType(Resources.Cache),
+                              ]
+            else:
+                print("Unknown platform. Nothing to delete!")
+                return
```
Ghostkeeper: Please only use `Resources.getConfigStoragePath` and such to get these paths. Also don't forget `Resources.getSearchPaths`. This should automatically work whenever we feel like changing the path locations on a rainy afternoon.

UM/Application.py
```diff
@@ -348,4 +357,57 @@ def getInstallPrefix():
...
+                                ]
+                file_patterns = []
+            elif Platform.isOSX():
+                dir_patterns = ["Library/Application Support/%s" %self._application_name,
+                                ]
+                file_patterns = []
+            elif Platform.isLinux():
+                dir_patterns = ['.config/%s' %self._application_name,
+                                '.cache/%s' %self._application_name,
+                                '.local/share/%s' %self._application_name,
```
Ghostkeeper: I hope you're going to remove this, but we use double quotes for strings and `os.path.join` instead of path separators :wink:

UM/Application.py
```diff
@@ -348,4 +357,57 @@ def getInstallPrefix():
...
+        for path in path_list:
+            try:
+                if os.path.isdir(path):
+                    print("Removing dir: ", path)
+                    shutil.rmtree(path)
+                if os.path.isfile(path):
+                    print("Removing file: ", path)
+                    os.remove(path)
+            except Exception:
+                print("Failed to remove: %s" %path)
```
Ghostkeeper: Maybe also put the actual exception there so that you can see why it failed (e.g. user rights or something else).

UM/Application.py
```diff
@@ -348,4 +357,57 @@ def getInstallPrefix():
...
+            user_basedir = os.path.split(os.path.expanduser("~"))[0]
+            user_paths = [os.path.join(user_basedir, entry) for entry in os.listdir(user_basedir)]
+            for user_path in user_paths:
+                for dir_pattern in dir_patterns:
+                    path_list.append(os.path.join(user_path, dir_pattern))
+                for file_pattern in file_patterns:
+                    path_list.append(os.path.join(user_path, file_pattern))
+
+        else:
+            if Platform.isLinux() or Platform.isOSX() or Platform.isWindows():
```
Ghostkeeper: Unnecessary checking.

thopiekar:
Damn, this PR is from 2016? Time passes fast.
However, about your list:
 * This should be not difficult to do. I was not that much familiar with Cura these days, so I guessed I missed all the knowledge, which is inside this module.
* This is not what I'm going to do. The option is meant to be used, when running the Windows NSIS deinstaller or running a post-remove script in a deb package. There you are able to gain admin rights. I see no reason to remove it. Only thing I would suggest is adding a check for these rights or try to remove whatever it possible, eg. removing all group members' files.
 * The merge conflicts should be hopefully easy to solve.

UM/Application.py
thopiekar: Hope this was/is a rhetorical statement 😛

UM/Application.py
thopiekar: Hmm, when looking at that I guess I had in mind to catch the case when Cura is running on none of our known OSs. But as soon as we are using `Resources` I think this is nothing that needs to get checked.

UM/Application.py
thopiekar: Yep, catching only what is worrying here actually.

( And by the way: All the debug prints 🙄 )

UM/Application.py
thopiekar: I can only imagine it won't be able to catch other users' config directories. If I'm not wrong we are expanding the home variable inside `Resources`, therefore the info is lost.

UM/Application.py
```diff
@@ -333,7 +341,8 @@ def parseCommandLine(self):
     #
     #   \param parser \type{argparse.ArgumentParser} The parser that will parse the command line.
     def addCommandLineOptions(self, parser):
-        pass
+        parser.add_argument("--mrproper", action="store_true", dest="mrproper", default=False, help="Remove user data")
+        parser.add_argument("--all-users", action="store_true", dest="all-users", default=False, help="Allow to remove user data from all users")
```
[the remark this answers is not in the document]
thopiekar: Will work when you are gaining admin rights. Mainly this is the condition where this option should allow a full cleanup of the computer.
Only very risky, if we are making mistakes. Eg.: Ooops, $HOME is gone 😬

thopiekar:
Conclusion:
This PR needs definitely a rework. The debug messages need to get removed and the paths need to get obtained from `Resources`. `--all-users` as an option is the main purpose of the PR and should be used inside the NSIS uninstaller and optionally on other platforms.
However, I made this working during my freelancing time for Ultimaker. Don't expect that I'm going to work on this and an implementation into NSIS in a short term.
I simply don't have the time to work on this right now.

If you think this PR is not important for you anymore, then you can also feel free to close it.

ianpaschal:
^ In that case we're going to close this. Would be nice to have in the future but a full rework can get its own, shiny, new, antibacterial PR.

UM/Application.py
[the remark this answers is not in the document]
Ghostkeeper: Reminiscent of this one: https://github.com/MrMEEE/bumblebee-Old-and-abbandoned/issues/123

UM/Application.py
Ghostkeeper: Yes, we are expanding the home variable there. This won't be a problem any more when the feature for removing other users' configurations is removed.

PR #193, opened 2016-12-07
Change ContainerStack serialization format to handle special characters properly

awhiemstra:
This changes how the containers stored in a ContainerStack are serialized, from a single entry with a comma-separated list to a single section with index, container_id entries. This simplifies the code and allows ConfigParser to deal with special characters and parsing. A bonus is that the resulting config file is easier to read.

The old format is still supported when deserializing to make sure this is properly backward compatible. In addition, I updated the unit tests and created two additional tests, on for backward compatibility and one for special characters.

Contributes to issue CURA-3098.

### DIFF
--- a/tests/Settings/TestContainerStack.py
+++ b/tests/Settings/TestContainerStack.py
@@ -664,6 +664,79 @@ def test_backwardCompatibility(container_stack, container_registry):
     container_stack.deserialize(serialised)
     assert container_stack.getContainers() == [container_a, container_a, container_a]
 
+##  Test serialization and deserialization of a stack with containers with special characters in their ID
+#
+def test_idSpecialCharacters(container_stack, container_registry):
+    container_ab = MockContainer("a,b") # Comma used to break deserialize
+    container_registry.addContainer(container_ab)
+
+    serialized = """
+    [general]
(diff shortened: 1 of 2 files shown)


nallath:
Note; Merging this will break project loading (pre-read makes some assumptions about how that string looks).

tests/Settings/TestContainerStack.py
```diff
@@ -597,6 +645,98 @@ def test_setNextStack(container_stack):
...
+    0 = ☂℮﹩⊥ ḉ◎η☂αїη℮ґ
+    """.format(version = UM.Settings.ContainerStack.Version)
+
+    container_stack.deserialize(serialized)
+    assert container_stack.getContainers() == [test_container_1]
+
+    serialized = container_stack.serialize()
+
+    # Unfortunately, we cannot check that serialized == container_stack.serialized() due to dict
+    # having a random order.
```
Ghostkeeper: We recently made instance containers use an OrderedDict in order to sort the settings in the profiles. Perhaps that's a good idea to use for container stacks, too?

jackha:
Project loading is fixed in Cura.

PR #195, opened 2016-12-08
Revert "Revert "i18n: Remove creation of en translations""

thopiekar:
Since we have already i18n-*-po-* targets, we can remove it from the script.
I'm just don't know much about about the x-test. So I left it in the script.

Reverted this commit, because we've not discussed about it, but when writing the commit message of the first commit I hit the enter key instead if the tabulator and tada!.. commited.
Sorry for this.

This reverts commit a9f30ee12a2be9fcdf1ae42bff62bad66d3cda30.

### DIFF
--- a/scripts/extract-messages
+++ b/scripts/extract-messages
@@ -20,11 +20,6 @@ msgconv --to-code=UTF-8 $basedir/resources/i18n/$catalogname.pot -o $basedir/res
 for pot in $basedir/resources/i18n/*.pot; do
     filename=$(basename $pot)
 
-    dir=$basedir/resources/i18n/en
-    mkdir -p $dir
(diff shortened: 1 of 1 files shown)


ChrisTerBeke:
@Ghostkeeper interesting for you?

Ghostkeeper:
I agree that it would be better to remove them from the automated script. I haven't updated the English .mo files in a while now. Uranium just falls back to the original string if it can't find the string in the English catalogue, but since the English catalogue is exactly the same as the original string, there is no difference to the user.

I wouldn't want to remove the .po and .mo files just yet though, since it may be useful again in the future to be able to change an English string after the string freeze (it's been used for that in the past).

PR #196, opened 2016-12-09
UpdateChecker: Don't tell user about an exception

thopiekar:
.. but about an error. I think when we are talking about exceptions it is clear what we are talking about.
In my point of view it is more than enough to tell the user that there was an error. Don't think he is neither interested in more details nor able to fix the problem on her/his own.

### DIFF
--- a/plugins/UpdateChecker/UpdateChecker.py
+++ b/plugins/UpdateChecker/UpdateChecker.py
@@ -84,7 +84,7 @@ class UpdateCheckerJob(Job):
                 Logger.log("w", "Did not find any version information for %s." % application_name)
         except Exception:
             Logger.logException("e", "Exception in update checker while parsing the JSON file.")
-            Message(i18n_catalog.i18nc("@info", "An exception occured while parsing the JSON file.")).show()
+            Message(i18n_catalog.i18nc("@info", "An error occured while parsing the JSON file.")).show()
             no_new_version = False # Just to suppress the message below.
 
         if no_new_version and not self.silent:


Ghostkeeper:
Agreed.

The "JSON" is also a bit technical though. Perhaps we should make that something like "...when processing application update information"?

thopiekar:
Yep, technical things should really only go into the logs. At least I think so.

I would also avoid using "JSON" here and "parsing".
I don't think the user is interested in what's happened in the background. Obviously UpdateChecker doesn't work and this only counts. Additionally it isn't fatal. So I would prefer to add an auto-hide here ( if there is no at the moment).

Mailed Menno about it. I heard he is not online here so often.
I think it is a good opportunity to get him involved here on GitHub 😉

mennoberg:
I agree. Avoid these developer terminology and try to write a message that communicates in user end/touch points.

fieldOfView:
Even worse than "JSON" being a developer term, the user has no context here, so he/she does not know what "the" JSON file is. 

Suggested text:
"The server gave an invalid response while checking for a new version."

thopiekar:
This text sounds fine to me. What about making it start with "The update server .. " or something like this. I still see the problem the user won't know at the end, which component of Cura has got this problem. Moreover it might be a bit tricky for our people at the support to understand where the problem is from.
If it was allowed we could use "UpdateChecker: Received unexpected content from our servers."
But I don't think we ever shown messages with the plugin name in it, isn't it?

Idea: If it was not much work I would even prefer to have message boxes with titles where we could add the plugin name, if it is relevant.

fieldOfView:
Euhm, no. To quote Menno: "Avoid these developer terminology and try to write a message that communicates in user end/touch points.". Plugin names are definitely developer terminology.

thopiekar:
Hmm, yes you are right 🤔 It is hard to decide here, because we as developers love to get the most detailed info.

I think we should let @mennoberg decide here 🙂

ChrisTerBeke:
Not relevant anymore since changes in master.

Ghostkeeper:
No, that's not true. The changes are still relevant. That string is still in there.

PR #198, opened 2016-12-22
Camera control bugs

Ando120:
1. Zoom in bug.

When you continue zoom in after you don't see visual difference, then you must zoom out same amount after you will see visual zoom out.
In CameraTool.py "self._min_zoom = 0" but mathematically it will move endlessly closer to 0.
"self._min_zoom = 1" is fixing that bug

2. space + mouse zoom jump
Space + mouse zoom "jump"(zoom in or out) sometimes when start new zoom. ( only with "space + mouse" zoom)
Script don't reset sometimes "self._start_y" value before next zoom.
added reset when "space" is not active.
Maybe not best option but it will fix that bug

### DIFF
--- a/plugins/Tools/CameraTool/CameraTool.py
+++ b/plugins/Tools/CameraTool/CameraTool.py
@@ -107,7 +107,7 @@ class CameraTool(Tool):
     def initiateZoom(self, event):
         if event.type is event.MousePressEvent:
             return False
-        elif event.type is Event.MouseMoveEvent and self._ctrl_is_active is False: #space -> mousemove
+        elif event.type is Event.MouseMoveEvent and self._space_is_active is False: #space -> mousemove
             self._start_y = None
         elif event.type is Event.MouseMoveEvent and self._space_is_active is True:  # space -> mousemove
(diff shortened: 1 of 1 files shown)


Ghostkeeper:
The space bar + mouse zoom doesn't work at all any more with these changes. I didn't see it during my brief review, but shouldn't it reset the `_start_y` when space is not pressed, instead of when Ctrl is not pressed?

The limit to zooming in works properly.

Ando120:
Sorry, fixed it. There was "_ctrl_is_active" and replaced it with "_space_is_active".
I am Blender user and changed control for zoom to myself.

Ghostkeeper:
Looking good!

Ando120:
When that fix is added to new version? It wasnt in 2.4beta and not in 2.4beta2 too.

Ghostkeeper:
It was merged into master. We split off the 2.4 version around the beginning of December, to allow development of new features to continue while 2.4 remains stable (new features don't get added to 2.4 but master). Because this was merged into master rather than 2.4, it'll only show up in public builds at the publication of 2.5.

PR #199, opened 2016-12-23
Fix size of dialogs on high-ish DPI screens

fieldOfView:
The calculation of screenScaleFactor (QtApplication.py) is fairly naive and has an arbitrary cutoff. It fails on my new 1080p 13" laptop (which has a ~160 DPI resolution): dialog windows (which don't set their own sizes) become a factor 2 too large.

It also does not take into account future screens that might have a higher screen scale factor (new iPads already have a scale factor of 3, so it is not unlikely that we will get 3x screens in computers).

Screen.devicePixelRatio works everywhere else in Cura and Uranium, so I don't know why this screenScaleFactor was introduced in the first place. Perhaps the analysis of CURA-422 has an answer, but I don't have access to that anymore.

### DIFF
--- a/UM/Qt/qml/UM/Dialog.qml
+++ b/UM/Qt/qml/UM/Dialog.qml
@@ -13,8 +13,8 @@ Window {
     modality: Qt.ApplicationModal;
     flags: Qt.Dialog | Qt.CustomizeWindowHint | Qt.WindowTitleHint | Qt.WindowCloseButtonHint;
 
-    width: screenScaleFactor * 640;
-    height: screenScaleFactor * 480;
+    width: Screen.devicePixelRatio * 640;
+    height: Screen.devicePixelRatio * 480;
 
     property int margin: Screen.devicePixelRatio * 8;
     property bool closeOnAccept: true;  // Automatically close the window when the window is "accepted" (eg using the return key)


Ghostkeeper:
We need to have this tested on Windows 10, because seeing CURA-422's comments that has been a problem all along.

sedwards2009:
Here is the analysis from JIRA:

"""
The button size problem was already fixed by the time I got to this issue. I've focused on fixing the size of the dialog itself.
Windows 10 on a laptop with a high DPI screen (retina screen MacBook for me) will choose its standard screen DPI used for font sizing based on what the output screen(s) is. If the laptop screen is used then the DPI will be high (like 220 DPI), if instead a HD external monitor is connected and the lid closed, then Windows will choose something closer to 96 DPI.
Cura uses physical pixels/resolution and Qt isn't set to do any UI scaling. The dialog for the "Check Printer" feature has a pixel based default width/height. Dialog.qml in Uranium makes an attempt to scale the size of the dialog based on the DPI. But it doesn't work. It uses 'Screen.devicePixelRatio' as a scale factor, but this number is only set by Qt in the situation where Qt is doing the UI scaling for you. Cura/Uranium don't use this feature, so the scale factor is always 1. (logical pixels == physical pixels).
I've added a variable 'screenScaleFactor' to the root QML context which should be used as a scale factor. It is either 1 or 2 and is chosen depending on the current screen DPI. QML's screen object doesn't expose the DPI information we need so we have to jam it into the QML context manually. Dialog.qml uses this value now to compute its default width/height.
I've tested this on Windows 10, MacOS, but no Linux.
"""

fieldOfView:
Thanks for posting the analysis. 

The analysis does not explain why ```Screen.devicePixelRatio``` cannot be used instead of the introduced ```screenScaleFactor```. The former is native, the latter naive ;-). 

The native property is used in a lot of places in the QML, and correctly handles "hi-dpi-but-not-pixeldoubling/retina", and will correctly handle a 3x factor when it becomes relevant. Why would dialog.qml need a different method of dealing with this than all the other qml files using ```Screen.devicePixelRatio```?

sedwards2009:
@fieldOfView Read the 3rd paragraph of the analysis I posted.

fieldOfView:
The 3rd paragraph does not explain why everything but dialog.qml is not currently broken though. Only dialogs that don't specify their own size currently use the screenScaleFactor. The only one that I can find in the current codebase is the MachineAction dialog. 

Could we at least switch to using ```primaryScreen().devicePixelRatio()``` instead of doing a calculation based on ```primaryScreen().physicalDotsPerInch```?
http://doc.qt.io/qt-5/qscreen.html#devicePixelRatio-prop

awhiemstra:
Note that Qt changed the way it handles high-DPI in 5.6 . This will probably have an impact here. See also https://blog.qt.io/blog/2016/01/26/high-dpi-support-in-qt-5-6/

Ghostkeeper:
I'm not entirely sure what to do with this pull request. I can't test this properly since I'm not on Windows, and I feel that it should be tested on Windows.

fieldOfView:
Specifically it needs to be tested on Windows 10 with a high DPI ("retina") screen, preferably in a build so we know for sure that the test uses the same version of Qt that the released version will use.

awhiemstra:
Sounds like something @jackha should do.

jackha:
dpi-hi-aldo
dpi-hi-master
dpi-lo-aldo
dpi-lo-master

I've been comparing this branch with the master.

The good news is... on a low dpi screen the result is the same. On the downside, both the high dpi screens are not what you would want. With this PR the window size is relative too small. The master has relative too big window sizes.

fieldOfView:
Thanks for testing.

From the screenshots, I would argue that the "this PR" versions are closer to the desirable sizes. I agree it is not perfect though.

I don't have a windows system with hidpi screen (only "mediumdpi", 1080p 13") so I cannot take this further.

Note there's also this issue:
https://github.com/Ultimaker/Cura/issues/1457

Ghostkeeper:
I think that the fix is acceptable, though it is not exactly the same.

jackha:
Ok I'm merging this.

PR #206, opened 2017-01-10
Big flame graph profiler support

sedwards2009:
The Uranium support needed for the Big Flame Graph profiler. It currently always inserts the shims to add profiling, there is no option to turn that on/off at start up.

The plugin itself is: https://github.com/sedwards2009/cura-big-flame-graph

### DIFF
--- a/UM/FlameProfiler.py
+++ b/UM/FlameProfiler.py
@@ -1,0 +1,149 @@
+# Uranium is released under the terms of the AGPLv3 or higher.
+
+import time
+import math
+from contextlib import contextmanager
+
+from UM.Logger import Logger
+
+###########################################################################
+SIGNAL_PROFILE = True
+record_profile = False
+
+class ProfileCallNode:
+    def __init__(self, name, line_number, start_time, end_time, children):
+        self.__name = name
+        self.__line_number = line_number
+        self.__start_time = start_time
+        self.__end_time = end_time
+        self.__children = children if children is not None else [] # type: List[ProfileCallNode]
+
+    def getStartTime(self):
+        return self.__start_time
+
+    def getEndTime(self):
+        return self.__end_time
+
+    def getDuration(self):
+        return self.__end_time - self.__start_time
+
+    def toJSON(self, root=False):
+        if root:
+            return """
+{
+  "c": {
+    "callStats": """ + self._plainToJSON() + """,
+    "sampleIterval": 1,
+    "objectName": "Cura",
+    "runTime": """ + str(self.getDuration()) + """,
+    "totalSamples": """ + str(self.getDuration()) + """
+  },
+  "version": "0.34"
+}
+"""
+        else:
+            return self._plainToJSON()
+
+    def _plainToJSON(self):
+        return '''{
+"stack": [
+  "''' + self.__name + '''",
+  "Code: ''' + self.__name + '''",
+  ''' + str(self.__line_number) + ''',
+  ''' + str(self.getDuration()) + '''
+],
+"sampleCount": '''+ str(self.getDuration()) + ''',
+"children": [
+    ''' + ",\n".join( [kid.toJSON() for kid in self.__children]) + '''
+]
+}
+'''
+child_accu_stack = [ [] ]
+clear_profile_requested = False
+record_profile_requested = False
+stop_record_profile_requested = False
+
+def getProfileData():
+    raw_profile_calls = child_accu_stack[0]
+    if len(raw_profile_calls) == 0:
+        return None
+
+    start_time = raw_profile_calls[0].getStartTime()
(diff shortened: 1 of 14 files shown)


UM/FlameProfiler.py
```diff
@@ -0,0 +1,187 @@
...
+            with profileCall(function.__qualname__):
+                return function(*args, ** kwargs)
+        else:
+            return function(*args, **kwargs)
+    return runIt
+
+##  Drop in replacement for PyQt5's pyqtSlot decorator which records profiling information.
+#
+#   See the PyQt5 documentation for information about pyqtSlot.
+def pyqtSlot(*args, **kwargs):
```
awhiemstra: I would like to have a switch that can disable this behaviour completely, preferably with an environment variable like I have done for Signal tracing. That way we can enable this only when we actually need it.

UM/Qt/Bindings/ContainerProxy.py
```diff
@@ -1,4 +1,5 @@
-from PyQt5.QtCore import QObject, pyqtSlot, pyqtSignal
+from PyQt5.QtCore import QObject
+from UM.FlameProfiler import pyqtSlot
```
awhiemstra: It is a shame that there is no less intrusive way to handle this. Because we will be forgetting this at one point or another.

UM/Qt/Bindings/Theme.py
```diff
@@ -1,10 +1,10 @@
 # Copyright (c) 2015 Ultimaker B.V.
 # Uranium is released under the terms of the AGPLv3 or higher.
 
-from PyQt5.QtCore import QObject, pyqtSlot, pyqtProperty, pyqtSignal, QCoreApplication, QUrl, QSizeF
+from PyQt5.QtCore import QObject, pyqtProperty, pyqtSignal, QCoreApplication, QUrl, QSizeF #, pyqtSlot
```
awhiemstra: No need for the comment at the end right?

UM/FlameProfiler.py
awhiemstra: This also applies to the "profile" decorator.

UM/FlameProfiler.py
```diff
@@ -0,0 +1,187 @@
...
+        global stop_record_profile_requested
+        if stop_record_profile_requested:
+            stop_record_profile_requested = False
+            record_profile = False
+            Logger.log('d', 'Stopping record stop_record_profile_requested')
+
+##  Decorator which can be manually applied to methods to record profiling information.
+#
+#   \type{Callable}
+def profile(function):
```
awhiemstra: Can we use something less generic for this? This is going to interfere with the python line_profiler module which I use a lot to profile single functions.

UM/Signal.py
```diff
@@ -121,6 +134,14 @@ def getType(self):
     #   function will be called on the next application event loop tick.
     @call_if_enabled(_traceEmit, _isTraceEnabled())
     def emit(self, *args, **kwargs):
+        FlameProfiler.updateProfileConfig()
```
awhiemstra: Here also, I would greatly prefer it if we can have a switch to disable all this behaviour. Considering the amount of calls to this method, I am really hesitant to add any sort of overhead to it.

UM/FlameProfiler.py
sedwards2009: "URANIUM_FLAME_PROFILER" sound good to you?

UM/Qt/Bindings/ContainerProxy.py
sedwards2009: If we forget it then nothing breaks. It will just be invisible. In some places we may want to remove the pyqtSlot stuff if it isn't producing anything interesting and just spamming up the results.

UM/FlameProfiler.py
awhiemstra: Something like that yes.

UM/FlameProfiler.py
sedwards2009: suggestions?

sedwards2009: Really, it is already specific. It lives in a namespace/module.

UM/Signal.py
sedwards2009: I've changed and moved most of the profiling stuff for emit() such that there is no overheat at runtime. Except for the one big `if` statement in the middle. There are ways to eliminate that too but I fear the result will be too obscure code-wise to justify.

UM/FlameProfiler.py
awhiemstra: It is fine if we then always import it fully qualified. Otherwise, maybe use flameProfile or so, to make it clear where it comes from.

sedwards2009: I've written out those refs the long way now.

PR #207, opened 2017-01-10
Jenkins ci test

awhiemstra:
This introduces a Jenkinsfile which is a script describing the process to build and test Uranium on the new Ultimaker CI system. In addition, it fixes a number of unit tests and marks others to be skipped.


awhiemstra:
Approved by @sedwards2009

PR #208, opened 2017-01-11
Fix build plate color for printers without a build plate texture

fieldOfView:
QImage() returns an image with uninitialised data. This could cause the buildplate of machines that don't have a texture (EG UMO) to become a random color.


awhiemstra:
Tiny change that fixes something ugly, merged.

Ghostkeeper:
Aww, I liked the pink UMO.

PR #209, opened 2017-01-12
Fix exception when file cannot be upgraded

fieldOfView:
This PR checks we got a valid result from upgrade_step() before trying to split the result so we get the intended warning instead of a (double printed?) exception.


Ghostkeeper:
Works a bit better there indeed. Thanks.

PR #210, opened 2017-01-18
Python type hinting

sedwards2009:
This PR adds basic Python 3.5 compatible type hints to Uranium. Type coverage is not 100% (yet).

Some comments about the types of changes you will see in this PR:

* Type hints have been added to a lot of places. There are still a lot of methods which don't have types. The type checker doesn't check the insides of these methods in these case. Note also that a couple of interface (super-classes) have been introduced to make static typing easier and to avoid circular dependencies.

* Ripping out the mega-imports we did inside __init__.py files for packages. Having a root package automatically import its subpackages is a big problem for a number of reasons:
    * Mypy doesn't like it because it is ambiguous and variables defined insides the __init__.py have the same names as subpackages. 
    * Importing everything like this leads to circular dependency hell as there is no nature order in which to import the subpackages.
    * It is better if users of the package import the subpackages they actually need. This makes it easier to see what they depend on and also easier if we need to break the meag-package up later on.

* *Using shorter import names instead of fully qualified.* This is related to the previous point above. Another reason to prefer shorter names is that the longer ones make the method signatures very very long. The argument for fully qualified names as being safer and more specific is less important now because we have a type checker which will flag an name clashes and mix ups.


UM/FileHandler/FileHandler.py
```diff
@@ -87,19 +89,6 @@ def readLocalFile(self, file):
...
-    #   \return A |type{Writer} instance or None if no writer supports the specified mime type. If there are multiple
-    #           writers that support the specified mime type, the first entry is returned.
-    def getWriterByMimeType(self, mime):
-        writer_data = PluginRegistry.getInstance().getAllMetaData(filter={self._writer_type: {}}, active_only=True)
-        for entry in writer_data:
-            for output in entry["mesh_writer"].get("output", []):
-                if mime == output["mime_type"]:
-                    return self._mesh_writers[entry["id"]]
-        return None
-
```
jackha: Nice find

UM/Operations/OperationStack.py
```diff
@@ -12,15 +12,15 @@
...
+    def __init__(self, controller):
         self._operations = [] #List of operations.
         self._current_index = -1 #Index of the most recently executed operation.
         self._lock = threading.Lock() #Lock to make sure only one thread can modify the operation stack at a time.
 
         # The merge behaviour must only occur when an operation is in the middle of a user action.
         # So, whenever an operation is started or ended, we do not want this auto-merge.
         self._merge_operations = False
-        self._controller = UM.Application.getInstance().getController()  # Circular dependency blah
+        self._controller = controller
```
jackha: Better

UM/Qt/QtApplication.py
```diff
@@ -168,8 +169,8 @@ def addCommandLineOptions(self, parser):
                             help="Disable Qt texture loading as a workaround for certain crashes.")
 
     #   Overridden from QApplication::setApplicationName to call our internal setApplicationName
-    def setApplicationName(self, name):
-        Application.setApplicationName(self, name)
+    # def setApplicationName(self, name: str) -> None:
+    #     Application.setApplicationName(name)
```
jackha: Please remove.

UM/Settings/ContainerInterface.py
```diff
@@ -1,98 +1,3 @@
 # Copyright (c) 2016 Ultimaker B.V.
 # Uranium is released under the terms of the AGPLv3 or higher.
```
jackha: Please remove the file.

UM/Settings/ContainerStack.py
```diff
@@ -268,7 +268,7 @@ def deserialize(self, serialized):
 
         if "containers" in parser:
             for index, container_id in parser.items("containers"):
-                containers = UM.Settings.ContainerRegistry.getInstance().findContainers(id = container_id)
+                containers = _containerRegistry.findContainers(id = container_id)
```
jackha: Nice

UM/Settings/SettingInstance.py
```diff
@@ -3,24 +3,29 @@
 
 import enum
 import os
+from typing import Any, List, Set, KeysView
 
+# from UM.Settings.Interfaces import ContainerInterface
```
jackha: Please remove.

UM/Decorators.py
```diff
@@ -61,11 +61,6 @@ def call_direct(decorated_function):
 #   excluding builtin functions like __getattr__. It is also expected to match the signature of
 #   those functions.
 def interface(cls):
-    # First, sanity check the interface declaration to make sure it only contains methods
-    invalid_properties = list(filter(lambda i: not i[0].startswith("__") and not inspect.isfunction(i[1]), inspect.getmembers(cls)))
-    if invalid_properties:
-        raise TypeError("Class {0} is declared as interface but includes non-method properties: {1}".format(cls, invalid_properties))
-
```
sedwards2009: To be honest I can't remember any more.

The decorators in this file perform some type checking related tasks at run time. We now have mypy for type checking. The stuff in here is obsolete now.

UM/Decorators.py
```diff
@@ -89,3 +84,6 @@ def immutable(cls):
     property_names = list(filter(lambda i: isinstance(i, property), inspect.getmembers(cls)))
     cls.__slots__ = property_names
     return cls
+
+def sameSignature(a: inspect.Signature, b: inspect.Signature) -> bool:
+    return len(a.parameters) == len(b.parameters)
```
sedwards2009: It is good enough until we remove it all and rely on mypy. I had to change this part of the code to allow methods which are the same to match in the case where one has types and other doesn't.

UM/FileHandler/FileHandler.py
sedwards2009: mypy found it.

PR #215, opened 2017-01-23
Themeable viewport

fieldOfView:
This PR makes the Theme layer available to python, so it can be used to theme the parts of the UI that is governed by python as opposed to QML, most notably the viewport. This paves the road for a dark theme, among other things.

Note that the theme relies on the qml engine to exist; the theme should not be accessed before Application.engineCreatedSignal is fired.

This PR relies on its Cura counterpart: https://github.com/Ultimaker/Cura/pull/1370

### DIFF
--- a/UM/Scene/ToolHandle.py
+++ b/UM/Scene/ToolHandle.py
@@ -25,14 +25,16 @@ class ToolHandle(SceneNode.SceneNode):
     ZAxis = 4
     AllAxis = 5
 
-    DisabledColor = Color(0.5, 0.5, 0.5, 1.0)
-    XAxisColor = Color(1.0, 0.0, 0.0, 1.0)
-    YAxisColor = Color(0.0, 0.0, 1.0, 1.0)
-    ZAxisColor = Color(0.0, 1.0, 0.0, 1.0)
-    AllAxisColor = Color(1.0, 1.0, 1.0, 1.0)
-
     def __init__(self, parent = None):
         super().__init__(parent)
+
+        self._disabled_axis_color = None
+        self._x_axis_color = None
+        self._y_axis_color = None
+        self._z_axis_color = None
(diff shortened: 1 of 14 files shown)


jackha:
Thanks!

UM/Scene/ToolHandle.py
```diff
@@ -25,15 +25,17 @@ class ToolHandle(SceneNode.SceneNode):
     ZAxis = 4
     AllAxis = 5
 
-    DisabledColor = Color(0.5, 0.5, 0.5, 1.0)
```
awhiemstra: Note that in my opinion these should not be themed. Instead, the display colour and the selection colour should have been split up. Having the selection colour themed makes it much more likely to collide with object colours, since we no longer have precise control over the selection colour.

fieldOfView: I think you have a good point, but looking at the code in SelectionPass it already makes sure that the toolhandle colors don't collide with object (selection-)colors. The only thing that will currently go wrong is if the themer decides it is prettier to let all the axes use the same color.

Having said that, do you have a good idea about how to split selection & display rendering colors? We could give the toolhandle meshes their selectionpass colors, and then in ToolHandle.render somehow detect a "display-render" and override the render color with a themed color. But I don't know how to detect the renderpass from a render() method.

awhiemstra: Toolhandle already has a concept of a "display mesh" versus a "selection mesh" since that was needed for the rotation handles. The simplest would be to simply give all tools a separate display and selection mesh. In fact, I believe the only tool that does not already have a separate mesh is the mirror tool, since it did not need larger hit areas.

fieldOfView: I'll get cracking!

PR #216, opened 2017-01-24
Fix the location of the configuration folder on Windows

fieldOfView:
~/AppData/Local may not be correctly expanded to the AppData folder on systems with itinerant profiles (eg corporate/educational networks where the user profile is stored on the network)

See https://github.com/Ultimaker/Cura/issues/1369 for discussion
Also see the Cura PR: https://github.com/Ultimaker/Cura/pull/1375

### DIFF
--- a/UM/Resources.py
+++ b/UM/Resources.py
@@ -264,7 +264,7 @@ class Resources:
     @classmethod
     def __initializeStoragePaths(cls):
         if platform.system() == "Windows":
-            cls.__config_storage_path = os.path.join(os.path.expanduser("~\\AppData\\Local\\"), cls.ApplicationIdentifier)
+            cls.__config_storage_path = os.path.join(os.getenv("LOCALAPPDATA"), cls.ApplicationIdentifier)
         elif platform.system() == "Darwin":
             cls.__config_storage_path = os.path.join(os.path.expanduser("~/Library/Application Support"), cls.ApplicationIdentifier)
             # For backward compatibility, support loading files from the old storage location


fieldOfView:
Though I feel the PR is more correct than the current situation, it does not fix the problem at hand and the usecase for what it fixes is just too small. The review burden is high enough as it is, so I am closing this PR.

awhiemstra:
Reviewing it is not very hard, the problem with this is testing.

Ghostkeeper:
Also, it is a Windows-only change, so it needs to be tested on Windows...

UM/Resources.py
```diff
@@ -264,7 +264,7 @@ def __find(cls, resource_type, *args):
     @classmethod
     def __initializeStoragePaths(cls):
         if platform.system() == "Windows":
-            cls.__config_storage_path = os.path.join(os.path.expanduser("~\\AppData\\Local\\"), cls.ApplicationIdentifier)
+            cls.__config_storage_path = os.path.join(os.getenv("LOCALAPPDATA"), cls.ApplicationIdentifier)
```
LipuFei: `LocalAppData` is introduced in Vista. I wonder if we should check the windows version, and use `os.path.expanduser("~\\AppData\\Local\\")` as a backwards compatible approach for earlier Windows versions?

LipuFei:
Tested. It works. Just wonder if we should keep some backward compatibility.

UM/Resources.py
Ghostkeeper: I don't think we've ever tried Cura on Windows XP. It's probably going to break anyway.

awhiemstra: In fact, I very consciously used a method in Arcus that is unsupported on XP. So yes, trying to run on XP will break. Which is why we do not support it.

PR #217, opened 2017-01-25
Added translation files for brazilian portuguese

Patola:
As per directions from Cura issue 676, I created this translation file alongside the translation files for Cura to have english strings translated to brazilian portuguese strings using gettext.

### DIFF
--- a/resources/i18n/ptbr/uranium.po
+++ b/resources/i18n/ptbr/uranium.po
@@ -1,0 +1,700 @@
+# Translation file to brazilian portuguese for the Uranium package
+# Copyright (C) 2017
+# This file is distributed under the same license as the Uranium package.
+# Cláudio Sampaio <patola@makerlinux.com.br>, 2017.
+#
+#, fuzzy
+msgid ""
+msgstr ""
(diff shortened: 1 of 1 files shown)


Patola:
Thanks! One question: the code still needs to add brazilian portuguese / portuguese as an interface language, right? Or does it pick this up automatically?

Ghostkeeper:
I hadn't yet checked or merged the Cura side of your translations (a meeting just came up in between). When I've done that, I plan to upload the machine-readable translation files, .mo, for the cura-binary-data repository: https://github.com/Ultimaker/cura-binary-data/tree/master/cura/resources/i18n
Then we need to add the language to the drop-down manually, here: https://github.com/Ultimaker/Cura/blob/master/resources/qml/Preferences/GeneralPage.qml#L93 This is done manually so that we can enable or disable languages if they are incomplete (though this is something that could be done in the CMake script eventually but it's okay). Once that's done, the language should appear in our internal builds as well as the unofficial PPA for Linux.

Other than the unofficial PPA and other people who build Cura themselves, the language will be made public when we make our public beta for 2.5, which will be in a month or two probably (there is no date yet).

PR #218, opened 2017-01-26
Fix toolhandle selection colors

fieldOfView:
When the colors for toolhandles are themed, we cannot rely on these colors being unique between axes, or between potential other "special" colors in the selection pass. This PR restores fixed constant colors per axis, for the selection meshes only. These fixed colors are never visible to the user because they are only used in the selection pass.

This PR is a followup for https://github.com/Ultimaker/Uranium/pull/215

### DIFF
--- a/UM/Scene/ToolHandle.py
+++ b/UM/Scene/ToolHandle.py
@@ -25,6 +25,14 @@ class ToolHandle(SceneNode.SceneNode):
     ZAxis = 4
     AllAxis = 5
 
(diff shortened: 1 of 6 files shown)


Ghostkeeper:
Oh, good point.

Ghostkeeper:
You'd think though that a theme designer would take care of such things, right?

Ghostkeeper:
I made the X axis bright pink but it still works! :heart:

PR #219, opened 2017-02-05
Adding setFileName to MeshData

thopiekar:
Generally this is not needed, but might be useful in some cases.

### DIFF
--- a/UM/Mesh/MeshData.py
+++ b/UM/Mesh/MeshData.py
@@ -131,6 +131,9 @@ class MeshData:
     def hasUVCoordinates(self):
         return self._uvs is not None
 
+    def setFileName(self, filename):
+        self._file_name = filename
+
     def getFileName(self):
         return self._file_name
 


sedwards2009:
This class is immutable and thus doesn't have any mutating methods. It actually already kind of has a `setFilename` method in the form `some_mesh.set(file_name="new filename")`.

thopiekar:
Thanks @sedwards2009. I never saw this kind of setting an attribute before. I'm using `some_mesh._file_name = "foobar"` at the moment and wondered why there is no set[...] function, while there is a get[...] function.

Regards and see you tomorrow 😉

thopiekar:
Replaced here `mesh._file_name = filePath` with `mesh.set(file_name=filePath)` and for me it doesn't work.
It seems like the path doesn't get overwritten.

sedwards2009:
What you are looking for is `new_mesh_with_changed_path = mesh.set(file_name=filePath)`. It doesn't modify `mesh` in place at all. You get a new one.

ChrisTerBeke:
Closing because no agreement from Cura team and it's old.

thopiekar:
Well, basically not needed.

PR #221, opened 2017-02-07
Improve performance of switching variant/materials/quality

awhiemstra:
This reduces the time it takes to switch variant on an UM3 from ~14 seconds on my machine to ~1.6 seconds.

It introduces two changes: 

* A way to postpone signals so they are emitted later, with optional compression (making sure to only emit a signals once).
* A cache for queries in ContainerRegistry::findContainers.
  We make a lot of calls to findContainers, often without any of the underlying list of containers changing. By caching the result, we can improve those cases, since we do not need to execute the expensive search operation again.

This fixes CURA-3312

### DIFF
--- a/UM/Settings/ContainerQuery.py
+++ b/UM/Settings/ContainerQuery.py
@@ -68,7 +68,7 @@ class ContainerQuery:
         return hash(self.__key())
 
     def __eq__(self, other):
-        return self.__key() == other.__key()
+        return isinstance(other, ContainerQuery) and self.__key() == other.__key()
 
     # protected:
 
--- a/UM/Settings/ContainerQuery.py
+++ b/UM/Settings/ContainerQuery.py
@@ -3,6 +3,8 @@
 
 import re
 
+from . import InstanceContainer
+
 ##  Wrapper class to perform a search for a certain set of containers.
 #
 #   This class is primarily intended to be used internally by
@@ -86,10 +88,8 @@ class ContainerQuery:
         elif property_name == "name":
             return value_pattern.match(container.getName())
         elif property_name == "definition":
-            try:
-                return value_pattern.match(container.getDefinition().getId())
-            except AttributeError:  # Only instanceContainers have a get definition. We can ignore all others.
-                pass
+            if isinstance(container, InstanceContainer.InstanceContainer):
+                return container.getDefinition() and value_pattern.match(container.getDefinition().getId())
 
         return value_pattern.match(str(container.getMetaDataEntry(property_name)))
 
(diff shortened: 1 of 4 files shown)


UM/Settings/ContainerQuery.py
```diff
@@ -0,0 +1,133 @@
...
+            if matches_container:
+                containers.append(container)
+
+        self._result = containers
+
+    def __hash__(self):
+        return hash(self.__key())
+
+    def __eq__(self, other):
+        return self.__key() == other.__key()
```
awhiemstra: Good point. Fixed.

UM/Settings/ContainerQuery.py
```diff
@@ -0,0 +1,133 @@
...
+            value_pattern = re.compile(value)
+
+        if property_name == "id":
+            return value_pattern.match(container.getId())
+        elif property_name == "name":
+            return value_pattern.match(container.getName())
+        elif property_name == "definition":
+            try:
+                return value_pattern.match(container.getDefinition().getId())
+            except AttributeError:  # Only instanceContainers have a get definition. We can ignore all others.
```
awhiemstra: It was code copied from the original implementation in ContainerRegistry. But I have now replaced those statements with an isinstance check.

UM/Signal.py
```diff
@@ -285,6 +286,63 @@ def __performEmit(self, *args, **kwargs):
...
+        delattr(signal, "_old_emit")
+        if hasattr(signal, "_postponed"):
+            if compress:
+                signal.emit(*signal._postponed[0], **signal._postponed[1])
+            else:
+                for args, kwargs in signal._postponed:
+                    signal.emit(*args, **kwargs)
+            delattr(signal, "_postponed")
+
+
```
awhiemstra: * I have moved the catching of signals into the emit() method. Note that I am using protected members directly in the context manager as I consider the context manager to be the "public API" for this functionality.
* I added a warning when emit() is called from a different thread than the postponing thread. Note though that it does not necessarily cause issues, it just means the emit() will be called later from a different thread.
* Good point, I have disabled connect/disconnect when a signals is being postponed.

As for the rest, I think having the postpone functionality is useful regardless of how we handle the signals elsewhere. Qt features something similar: http://doc.qt.io/qt-5/qobject.html#blockSignals and I have run into situations where it is really useful.

UM/Signal.py
```diff
@@ -189,6 +189,10 @@ def emit(self, *args, **kwargs):
     #   \param connector The signal or slot (function) to connect.
     @call_if_enabled(_traceConnect, _isTraceEnabled())
     def connect(self, connector):
+        if self._postpone_emit:
+            Logger.log("w", "Tried to connect to signal %s that is currently being postponed, this is not possible", self.__name)
```
sedwards2009: In this situation I think that dumping a stack trace isn't out of the question. At least then we can see what happened via the logs.

PR #224, opened 2017-02-20
Fixed center() error and minor behavior changes

neut:
Fixed AxisAlignedBox.center() method non-callable error and have the .minimum() and .maximum() methods return the actual data rather than the Vector object.


nallath:
This won't work, as we expect the min & max to return a Vector object and not a numpy object. This is also why the unit test failed.

neut:
Fixed on my end. Thnx for the review

PR #228, opened 2017-03-10
Simplified Chinese Translate

alphaarea:
Add Chinese Translate
only translated basic setting word

translated by Poedit 1.8.12, but it still show English on my computer.
I'm not sure if there are any errors


Ghostkeeper:
You are committing Cura files to the Uranium repository. Furthermore the translation is mostly empty, so the text would disappear. We can't accept this pull request.

To actually see your translations in Cura, you'd need to compile these translation files to .mo files using gettext. Place the .mo files in the LC_MESSAGES subdirectory of the language's resource folder. Then change the preference in cura.cfg to your language key (`cn`).

PR #232, opened 2017-03-16
Added invert zoom preference option

nallath:
Adds a way to invert the zoom direction for Cura. A few engineers requested this, as it matches the default solid works behavior better. Also see https://github.com/Ultimaker/Cura/pull/1548

### DIFF
--- a/plugins/Tools/CameraTool/CameraTool.py
+++ b/plugins/Tools/CameraTool/CameraTool.py
@@ -2,6 +2,7 @@
 # Uranium is released under the terms of the AGPLv3 or higher.
 
 from UM.Tool import Tool
+from UM.Preferences import Preferences
 from UM.Event import Event, MouseEvent, KeyEvent
 from UM.Math.Vector import Vector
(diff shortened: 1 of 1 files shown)


plugins/Tools/CameraTool/CameraTool.py
```diff
@@ -40,6 +41,15 @@ def __init__(self):
 
         self._drag_distance = 0.05
 
+        Preferences.getInstance().addPreference("view/invert_zoom", False)
+        self._invert_zoom = Preferences.getInstance().getValue("view/invert_zoom")
+        Preferences.getInstance().preferenceChanged.connect(self._onPreferencesChanged)
+
+    def _onPreferencesChanged(self, name):
+        if name != "view/invert_zoom":
+            return
```
Ghostkeeper: I guess it would've made sense to just go:

    if name == "view/invert_zoom":
        self._invert_zoom = Preferences.getInstance().getValue("view/invert_zoom")

But this is okay too. Not blocking.

PR #234, opened 2017-03-24
CURA-3527 Use different local dir per version

LipuFei:


UM/Resources.py
```diff
@@ -264,33 +265,34 @@ def __find(cls, resource_type, *args):
 
     @classmethod
     def __initializeStoragePaths(cls):
+        storage_dir_name = cls.ApplicationIdentifier + "-" + cls.ApplicationVersion
```
nallath: I thought we were gonna use a subfolder for this?

LipuFei:
@nallath Changed to sub-directories :)

fieldOfView:
How does this work with the upgrade system?

LipuFei:
@fieldOfView That will be finished in another JIRA issue. We separated it into small tasks.

PR #235, opened 2017-03-24
Allow the user to disable/enable plugins

fieldOfView:
This is useful eg when users don't want to use USB printing, and they don't want to the USB printing plugin to reset their arduinos when Cura starts.

NB: The application needs to be restarted for changes to have effect.

### DIFF
--- a/UM/PluginRegistry.py
+++ b/UM/PluginRegistry.py
@@ -5,6 +5,7 @@ import imp
 import os
 from typing import Dict
 
+from UM.Preferences import Preferences
 from UM.PluginError import PluginNotFoundError, InvalidMetaDataError
 from UM.Logger import Logger
 from typing import Callable, Any
@@ -30,6 +31,10 @@ class PluginRegistry(object):
         self._folder_cache = {}
         self._application = None
         self._active_plugins = []
+
+        preferences = Preferences.getInstance()
+        preferences.addPreference("general/disabled_plugins", "")
+        self._disabled_plugins = None
 
     ##  Check if all required plugins are loaded.
     #   \param required_plugins \type{list} List of ids of plugins that ''must'' be activated.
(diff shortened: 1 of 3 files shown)


Ghostkeeper:
Well, it indeed disables a plug-in when you uncheck the tag.

I tested this by disabling the CuraEngine backend. It crashed on start-up. It seems some of the check marks are disabled but clicking the text next to the check mark still enables or disables the plug-in.

Ghostkeeper:
Disabling the "Material Profiles" plug-in breaks just about everything.

Ghostkeeper:
Apart from fixing the clicking of labels to disable/enable required plug-ins, I'd like to suggest a method that is resilient to people editing their config file as well: Add an additional check to ignore any required plug-ins if they happen to occur in that disabled plug-in list in the config file.

fieldOfView:
Plugins that break stuff if they are not there should be marked as required plugins.

> I'd like to suggest a method that is resilient to people editing their config file as well

This is not consistent with decisions made elsewhere, where it was said that the configuration file is to be considered the program state and not user editable. But I could make a check for this.

Ghostkeeper:
Fair enough. So the only issue that still stands from my test is that you can disable required plug-ins by clicking on their label instead of their checkbox.

fieldOfView:
Yeah, I figured that out too. Somewhat asking for trouble to have both a checkbox and a button

UM/PluginRegistry.py
```diff
@@ -31,6 +32,10 @@ def __init__(self):
         self._application = None
         self._active_plugins = []
 
+        preferences = Preferences.getInstance()
+        preferences.addPreference("general/disabled_plugins", "")
+        self._disabled_plugins = None
```
nallath: This should be a list right?

fieldOfView: It gets reinitialised to a list here: https://github.com/fieldOfView/Uranium/blob/3886d9ff2900c2e9d09f78ec9c3416fee0ebdf4f/UM/PluginRegistry.py#L88

At the time the pluginregistry is initialised, the preferences may not yet be available.

fieldOfView: It would have been nice if I had added a comment to that effect. I can do another PR for that, but that may just result in more overhead.

nallath: Fair enough. I will revert the commit & add a comment about why it's none.

PR #236, opened 2017-03-24
Use Roaming instead of Local AppData on Windows

fieldOfView:
Applications should use the Roaming AppData folder for data that is supposed to follow the user around on different computers.

Followup to https://github.com/Ultimaker/Uranium/pull/216
See https://github.com/Ultimaker/Cura/issues/1369 for discussion

This changes the path for configurations, so an upgrade step needs to be implemeted. Now is a good time to make this change, since the configuration folder structure is changing anyway: https://github.com/Ultimaker/Uranium/pull/234

### DIFF
--- a/UM/Resources.py
+++ b/UM/Resources.py
@@ -269,7 +269,7 @@ class Resources:
         storage_dir_name = os.path.join(cls.ApplicationIdentifier, cls.ApplicationVersion)
 
         if platform.system() == "Windows":
-            cls.__config_storage_path = os.path.join(os.getenv("LOCALAPPDATA"), storage_dir_name)
+            cls.__config_storage_path = os.path.join(os.getenv("APPDATA"), storage_dir_name)
         elif platform.system() == "Darwin":
             cls.__config_storage_path = os.path.join(os.path.expanduser("~/Library/Application Support"), storage_dir_name)
             # For backward compatibility, support loading files from the old storage location


nallath:
For our reference; CURA-3593

BagelOrb:
I don't really understand what this is about; is this a change from using AppData/Local to using AppData/Roaming or the other way around?

Wouldn't this break the version upgrader?

Ghostkeeper:
Yes, from AppData/Local to AppData/Roaming and no, this would not break the version upgrader itself but it would require an upgrade step just like we did in 2.1->2.3 where the folder location was changed.

fieldOfView:
This really needs to be fixed at some stage, because it seems to be a somewhat blocking issue for corporate users where the profile is stored on a network server.

Now is probably a good time to change this, since there needs to be an upgrade step to move to the new version-specific settings location anyway.

awhiemstra:
So according to this: https://superuser.com/questions/150012/what-is-the-difference-between-local-and-roaming-folders the "Local" directory should contain machine-specific stuff versus the "Roaming" directory containing machine-independent stuff. I think we should put the definition cache into local, since that is to a certain degree machine-specific, while keeping the configuration data in Roaming.

LipuFei:
I think if we change that to "Roaming", the logs should also be in "Roaming" too?

fieldOfView:
From @awhiemstra's standpoint, the logs should probably remain in Local. From a convenience standpoint, it would be better to have them in the same place as the configuration, otherwise it is going to be harder to explain where to get the logs; currently you can just point to the "Show Configuration Folder" in the Help menu.

PR #240, opened 2017-04-04
CURA-3648 Don't save (auth keys) certain metadata key-values during serialization

LipuFei:
Add some code making it possible to exclude certain keys during serialization, so we can ignore the auth keys.
I believe this one should go to `2.5`.

### DIFF
--- a/UM/Settings/ContainerStack.py
+++ b/UM/Settings/ContainerStack.py
@@ -51,6 +51,7 @@ class ContainerStack(ContainerInterface, PluginObject):
         self._id = str(stack_id)
         self._name = stack_id
         self._metadata = {}
+        self._metadata_serialization_excluded_keys = set()
         self._containers = []
         self._next_stack = None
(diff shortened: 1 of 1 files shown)


UM/Settings/ContainerStack.py
```diff
@@ -232,7 +236,8 @@ def serialize(self):
 
         parser["metadata"] = {}
         for key, value in self._metadata.items():
-            parser["metadata"][key] = str(value)
+            if key not in self._metadata_serialization_excluded_keys:
```
jackha: (non blocking) small taste difference: I would say
if key in self._metadata_serialization_excluded_keys:
  continue
<do all other things>

This prevents indenting a potentially big block. Also, I can wrap my head around it easier. Now you have to read: what does it mean to have a key that's not in the excluded list?

LipuFei:
We just don't need to save the keys to a project file. Closing this one

PR #242, opened 2017-04-10
CURA-3529 Copy the latest cura dir (if present) for upgrade

LipuFei:
In combination with https://github.com/Ultimaker/Cura/pull/1672

### DIFF
--- a/UM/Resources.py
+++ b/UM/Resources.py
@@ -3,6 +3,7 @@
 
 import os
 import os.path
+import re
 import shutil
 from typing import List
 
@@ -413,6 +414,9 @@ class Resources:
 
     @classmethod
     def _findLatestConfigDirInPaths(cls, search_path_list):
+        # version dir name must match: <digit(s)>.<digit(s)><whatever>
+        version_regex = re.compile(r'^[0-9]+\.[0-9]+.*$')
+
         latest_config_path = None
         for search_path in search_path_list:
             if not os.path.exists(search_path):
@@ -427,9 +431,19 @@ class Resources:
                 storage_dir_name_list = sorted(storage_dir_name_list, reverse=True)
                 # for now we use alphabetically ordering to determine the latest version (excluding master)
                 for dir_name in storage_dir_name_list:
-                    if not dir_name.endswith("master"):
-                        latest_config_path = os.path.join(search_path, dir_name)
-                        break
+                    if dir_name.endswith("master"):
+                        continue
+                    if version_regex.match(dir_name) is None:
+                        continue
+
+                    # make sure that the version we found is not newer than the current version
+                    if version_regex.match(cls.ApplicationVersion):
+                        later_version = sorted([cls.ApplicationVersion, dir_name], reverse=True)[0]
+                        if cls.ApplicationVersion != later_version:
+                            continue
+
+                    latest_config_path = os.path.join(search_path, dir_name)
+                    break
             if latest_config_path is not None:
                 break
         return latest_config_path
--- a/UM/Resources.py
+++ b/UM/Resources.py
@@ -9,7 +9,6 @@ from typing import List
 
 from UM.Logger import Logger
 from UM.Platform import Platform
-import UM.VersionUpgradeManager
 
 class ResourceTypeError(Exception):
     pass
@@ -301,9 +300,12 @@ class Resources:
     @classmethod
     def _getPossibleDataStorageRootPathList(cls):
         # Returns all possible root paths for storing app configurations (in old and new versions)
-        data_root_list = [Resources._getDataStorageRootPath()]
+        data_root_list = []
+        if Resources._getDataStorageRootPath() is not None:
+            data_root_list.append(Resources._getDataStorageRootPath())
         if Platform.isWindows():
             # it used to be in LOCALAPPDATA on Windows
+            data_root_list.append(os.getenv("APPDATA"))
             data_root_list.append(os.getenv("LOCALAPPDATA"))
         elif Platform.isOSX():
             data_root_list.append(os.path.expanduser("~"))
@@ -403,8 +405,11 @@ class Resources:
             # If the directory found matches the current version, do nothing
             return
 
+        # Prevent circular import
+        import UM.VersionUpgradeManager
         UM.VersionUpgradeManager.VersionUpgradeManager.getInstance().copyVersionFolder(latest_config_path, this_version_config_path)
-        if latest_data_path is not None and os.path.exists(latest_data_path):
+        # If the data dir is the same as the config dir, don't copy again
+        if latest_data_path is not None and os.path.exists(latest_data_path) and latest_data_path != latest_config_path:
             UM.VersionUpgradeManager.VersionUpgradeManager.getInstance().copyVersionFolder(latest_data_path, this_version_data_path)
 
         # Remove "cache" if we copied it together with config
--- a/UM/Resources.py
+++ b/UM/Resources.py
@@ -3,14 +3,12 @@
 
 import os
 import os.path
-import platform
 import shutil
-import tempfile
 from typing import List
 
 from UM.Logger import Logger
 from UM.Platform import Platform
-
+import UM.VersionUpgradeManager
 
 class ResourceTypeError(Exception):
     pass
@@ -383,13 +381,11 @@ class Resources:
     #   as the base for upgrade
     @classmethod
     def _copyLatestDirsIfPresent(cls):
-        # paths for the version we are running right now
+        # Paths for the version we are running right now
         this_version_config_path = Resources.getConfigStoragePath()
         this_version_data_path = Resources.getDataStoragePath()
 
-        Logger.log("d", "Setting up new paths, either because this is a new install or a upgrade.")
-
-        # find the latest existing directories on this machine
+        # Find the latest existing directories on this machine
         config_root_path_list = Resources._getPossibleConfigStorageRootPathList()
         data_root_path_list = Resources._getPossibleDataStorageRootPathList()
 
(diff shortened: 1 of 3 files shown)


fieldOfView:
Cura workarounds in Uranium? Hath hell frozen over? ;-)
This should probably be in an VersionUpgrade plugin.

LipuFei:
@fieldOfView discussed with @nallath about this. If there is an earlier cura directory, we want to duplicate that one as the start point for upgrading to the newer version. This way the old configs are not touched and running the previous version will still work.

This needs to be done before we start anything, so we put the code here. `VersionUpgrade` plugin will just take care of upgrading the actual files.

nallath:
We could make it a bit more generic by changing 
 "findLatestCuraDirInPaths" to "findLatestDirInPaths", which uses a more generic approach. 
As far as I can see, this should be doable.

fieldOfView:
@LipuFei: I understand what this code is doing, it is just that all the references to Cura (and even a specific hardcoded version of Cura) inside a Uranium file are quite "illegal".

LipuFei:
@fieldOfView @nallath Ah, right. Good point! I forgot we are in Uranium... I will change those too.

LipuFei:
@fieldOfView @nallath code is more generic now, and some unit tests are added.

fieldOfView:
I'm still not a fan of ```_isConfigDirForAppBefore26```. Version 2.6 of what?
Perhaps name it ```_isNonVersionedConfigDir```?

UM/Resources.py
```diff
@@ -264,55 +268,188 @@ def __find(cls, resource_type: int, *args) -> List[str]:
...
-    def __initializeStoragePaths(cls):
-        # use nested structure: cura/<version>/...
-        if cls.ApplicationVersion == "master" or cls.ApplicationVersion == "unknown":
-            storage_dir_name = os.path.join(cls.ApplicationIdentifier, cls.ApplicationVersion)
-        else:
-            from UM.Version import Version
-            version = Version(cls.ApplicationVersion)
-            storage_dir_name = os.path.join(cls.ApplicationIdentifier, "%s.%s" % (version.getMajor(), version.getMinor()))
-
+    def getConfigStorageRootPath(cls):
```
awhiemstra: This exposes an abbreviation as public API. (Config vs. Configuration) We generally try not use abbreviations, and should definitely not expose them as public API.

UM/Resources.py
```diff
@@ -264,55 +268,188 @@ def __find(cls, resource_type: int, *args) -> List[str]:
...
-            cls.__config_storage_path = os.path.join(xdg_config_home, storage_dir_name)
+                config_path = os.path.expanduser("~/.config")
+        else:
+            config_path = "."
 
-            xdg_data_home = ""
+        return config_path
+
+    @classmethod
+    def getPossibleConfigStorageRootPathList(cls):
```
awhiemstra: Why is this here? Can we move this to VersionUpgrade?

UM/Resources.py
```diff
@@ -264,55 +268,188 @@ def __find(cls, resource_type: int, *args) -> List[str]:
...
+        # Returns all possible root paths for storing app configurations (in old and new versions)
+        config_root_list = [Resources.getConfigStorageRootPath()]
+        if platform.system() == "Windows":
```
awhiemstra: Use UM.Platform.isWindows() instead.

UM/Resources.py
```diff
@@ -264,55 +268,188 @@ def __find(cls, resource_type: int, *args) -> List[str]:
...
+            # it used to be in LOCALAPPDATA on Windows
+            config_root_list.append(os.getenv("LOCALAPPDATA"))
+
```
awhiemstra: This misses the old storage location on OSX, which used to be ~/.cura . I cannot recall when we made that change though, so I do not know if we want to keep it.

UM/Resources.py
```diff
@@ -264,55 +268,188 @@ def __find(cls, resource_type: int, *args) -> List[str]:
...
+        config_root_list = [os.path.join(n, cls.ApplicationIdentifier) for n in config_root_list]
+        return config_root_list
+
+    @classmethod
+    def getDataStorageRootPath(cls):
```
awhiemstra: I would prefer it if we change these three to a "getStorageRootPath" that takes an enum to get the path we want. This makes the API more forward compatible and cleaner since adding an extra storage root just means adding an enum entry.

UM/Resources.py
```diff
@@ -264,55 +268,188 @@ def __find(cls, resource_type: int, *args) -> List[str]:
...
 
-            xdg_cache_home = ""
+    @classmethod
+    def getCacheStorageRootPath(cls):
+        # Returns the path where we store different versions of app configurations
+        cache_path = None
+        if platform.system() == "Windows":
+            cache_path = os.getenv("LOCALAPPDATA")
+        elif platform.system() == "Darwin":
+            cache_path = None
```
awhiemstra: Ehm, what? So getting the cache storage path for OSX leads to an invalid path?

UM/Resources.py
```diff
@@ -264,55 +268,188 @@ def __find(cls, resource_type: int, *args) -> List[str]:
...
             except KeyError:
-                xdg_data_home = os.path.expanduser("~/.local/share")
-            cls.__data_storage_path = os.path.join(xdg_data_home, storage_dir_name)
+                data_path = os.path.expanduser("~/.local/share")
+        return data_path
 
-            xdg_cache_home = ""
+    @classmethod
+    def getCacheStorageRootPath(cls):
+        # Returns the path where we store different versions of app configurations
```
awhiemstra: Documentation should be above the method declaration. Also, you are missing documentation for return type. I tend to use the following template:

```
##  Brief description of the method/class/etc.
#
#   Longer description of the item.
#
#   \param paramName \type{type} Description of the parameter. (Repeat for all parameters).
#
#   \return \type{type} Description of the return type.

File: UM/Resources.py [hunk 8]
```diff
@@ -264,55 +268,188 @@ def __find(cls, resource_type: int, *args) -> List[str]:
...
+            storage_dir_name = os.path.join(cls.ApplicationIdentifier, "%s.%s" % (version.getMajor(), version.getMinor()))
+
+        # config is saved in "<CONFIG_ROOT>/<storage_dir_name>"
+        cls.__config_storage_path = os.path.join(Resources.getConfigStorageRootPath(), storage_dir_name)
+
+        # data is saved in
+        #  - on Linux: "<DATA_ROOT>/<storage_dir_name>"
+        #  - on other: "<CONFIG_DIR>" (in the config directory)
+        data_root_path = Resources.getDataStorageRootPath()
+        cls.__data_storage_path = cls.__config_storage_path if data_root_path is None else \
```
awhiemstra: Please just write this as

```
if data_root_path is None:
  cls.__data_storage_path = cls.__config_storage_path
else:
  cls.__data_storage_path = os.path.join(data_root_path, storage_dir_name)
```

This prevents ugly escaping of newlines and is generally much more readable.

UM/Resources.py
```diff
@@ -264,55 +268,188 @@ def __find(cls, resource_type: int, *args) -> List[str]:
...
+
+        if not os.path.exists(cls.__config_storage_path):
+            cls._copyLatestDirsIfPresent()
 
         cls.__paths.insert(0, cls.__data_storage_path)
 
+    ##  Copies the directories of the latest version on this machine if present, so the upgrade will use the copies
+    #   as the base for upgrade
+    @classmethod
+    def _copyLatestDirsIfPresent(cls):
```
awhiemstra: Why is this here? Can we move this to VersionUpgrader?

UM/Resources.py
```diff
@@ -264,55 +268,188 @@ def __find(cls, resource_type: int, *args) -> List[str]:
...
+                for dir_name in storage_dir_name_list:
+                    if not dir_name.endswith("master"):
+                        latest_storage_dir_path = os.path.join(search_path, dir_name)
+                        break
+            if latest_storage_dir_path is not None:
+                break
+        return latest_storage_dir_path
+
+    @classmethod
+    def _isConfigDirForAppBefore26(cls, check_path):
```
awhiemstra: I agree with Aldo that this should have a different name.

UM/Resources.py
```diff
@@ -264,55 +268,188 @@ def __find(cls, resource_type: int, *args) -> List[str]:
...
+            return True
+
+        dirs, files = next(os.walk(check_path))[1:]
+        valid_dir_names = [dn for dn in dirs if dn in Resources.__expected_dir_names_in_config]
+        valid_file_names = [fn for fn in files if not fn.endswith('.log')]  # do not include log files
+
+        return valid_dir_names and valid_file_names
+
+    @classmethod
+    def addExpectedDirNameInConfig(cls, dir_name):
```
awhiemstra: What does this do?

UM/Resources.py
```diff
@@ -264,55 +268,188 @@ def __find(cls, resource_type: int, *args) -> List[str]:
...
+        if not latest_config_path:
+            # no earlier storage dirs found, do nothing
+            return
+
+        if latest_config_path == this_version_config_path:
+            # if the directory found matches the current version, do nothing
+            return
+
+        # we first copy everything to a temporary folder, and then move it to the new folder
+        def _copyDirToDirViaTempDir(src_path, dest_path):
```
awhiemstra: Does this need to be a closure?

LipuFei:
@fieldOfView That's a better name. I will fix it.

UM/Resources.py
LipuFei: The closure can go. It just means that this function is not meant for anything out of this scope.

UM/Resources.py
LipuFei: The VersionUpgrader is an plugin, which means the plugins will already he loaded when VersionUpgrader is launched. A user can have custom plugins in his/her old app config directory, e.g. `cura/plugins`. If the VersionUpgrader decides to upgrade from that old version, Cura needs to restart to load those custom plugins. That's why we decide to run those code before actually doing anything.

UM/Resources.py
LipuFei: Will do

UM/Resources.py
LipuFei: There are some fuzzy logic in the old code too. Basically, there must be a path for the **config dir**. If no place can be found for it, it will be `.` the current directory. But for **data** and **cache**, it depends. Only on Linux do we have separate places for all of them; On Windows **config** and **data** are kept in the Roaming, and **cache** in Local; On Mac all of them are in the same place. This basically means **data** and **cache** are optional to have their own places, thus the `None` in `getCache` and `getData`.

UM/Resources.py
LipuFei: Will fix

UM/Resources.py
LipuFei: This is to detect an unversioned directory.

UM/Resources.py
awhiemstra: Yeah, but the old code ensured we always had a proper path, even if that path might be the same as another. Right now, what this code does is lead to a call like "os.path.join(Resources.getCacheRootPath(), "blah") ending up with either a TypeError or an invalid path.

PR #243, opened 2017-04-12
CURA-3540 update profiles when deserializing them

LipuFei:
Create PR for now to facilitate reviewing and discussion.

What it does:
- Before deserialize a profile, we try to update it first
- Did some hacking to ensure as little change as possible
- I added some **FIXME** marks on the things I'm sure if they are correct.

### DIFF
--- a/UM/Settings/ContainerStack.py
+++ b/UM/Settings/ContainerStack.py
@@ -244,6 +244,10 @@ class ContainerStack(ContainerInterface, PluginObject):
         parser.write(stream)
         return stream.getvalue()
 
+    ##  Deserializes the given data and checks if the required fields are present.
+    #
+    #   The profile upgrading code depends on information such as "configuration_type" and "version", which come from
+    #   the serialized data. Due to legacy problem, those data may not be available if it comes from an ancient Cura.
     def _readAndValidateSerialized(self, serialized: str) -> configparser.ConfigParser:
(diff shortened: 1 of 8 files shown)


UM/Settings/ContainerStack.py
```diff
@@ -244,17 +244,37 @@ def serialize(self):
...
 
+    def _readAndValidateSerialized(self, serialized: str) -> configparser.ConfigParser:
+        parser = configparser.ConfigParser(interpolation=None, empty_lines_in_values=False)
+        parser.read_string(serialized)
+
+        if "general" not in parser or any(pn not in parser["general"] for pn in ("version", "name", "id")):
+            raise InvalidContainerStackError("Missing required section 'general' or 'version' property")
+
+        return parser
+
```
LipuFei: Documented in the last commit, right above the `def` line.

awhiemstra:
Note that this causes almost all the Settings unit tests to fail, since you changed the main interface. This should not be merged before those are fixed.

awhiemstra:
Uhm, what happened to "do not merge this until the unit tests are fixed"?

nallath:
You didn't put it back into to-do? I missed this comment and the code worked.

LipuFei:
@nallath @awhiemstra I've fixed the unit tests. This is the commit: https://github.com/Ultimaker/Uranium/commit/de107ee5b374ded641dbc644f28e7bd8276fae72

PR #245, opened 2017-04-26
Replace "Root" storage path code with QStandardPaths

awhiemstra:
When fixing an issue with the storage path copy code (see https://github.com/Ultimaker/Uranium/commit/1318b11eb6fad99b36da4824fc7932da0593eda3) I realized we have missed a huge opportunity when adding that code, which is to replace our custom "root" path handling code with [QStandardPaths](http://doc.qt.io/qt-5/qstandardpaths.html). This PR fixes that by using QStandardPaths for the base resource directories.

This has a few implications for the paths we use:
- On Linux, nothing changes.
- On Windows, instead of storing preferences in AppData/Roaming, we store them in AppData/Local. QStandardPaths only uses Roaming for the application data.
- On OSX, instead of storing everything in Application Support, we now store preferences in Library/Preferences, cache data in Library/Caches and data in Library/Application Support. This actually matches what Apple recommends Application do.

Since we are already messing around with paths for 2.6, I wanted to have this done before we move to beta, as it is better to only once mess up the locations.

### DIFF
--- a/UM/Qt/QtApplication.py
+++ b/UM/Qt/QtApplication.py
@@ -67,6 +67,9 @@ class QtApplication(QApplication, Application):
 
         super().__init__(sys.argv, **kwargs)
 
+        # Set the QCoreApplication applicationName property
+        self.setApplicationName(self.getApplicationName())
+
         self.setAttribute(Qt.AA_UseDesktopOpenGL)
         major_version, minor_version, profile = OpenGLContext.detectBestOpenGLVersion()
 
--- a/UM/Resources.py
+++ b/UM/Resources.py
@@ -278,6 +278,9 @@ class Resources:
             config_root_list.append(os.path.join(os.path.expanduser("~"), cls.ApplicationIdentifier))
             # Config storage path on OSX used to be ~/Library/Application Support/cura but changed to ~/Library/Preferences/cura for 2.6.
             config_root_list.append(os.path.normpath(QStandardPaths.writableLocation(QStandardPaths.AppDataLocation)))
+        elif Platform.isWindows():
+            # Config used to be in AppData/Local, so make sure to include that.
+            config_root_list.append(os.path.normpath(QStandardPaths.writableLocation(QStandardPaths.AppLocalDataLocation)))
 
         return config_root_list
 
@@ -298,7 +301,11 @@ class Resources:
     # Returns the path where we store different versions of app configurations
     @classmethod
     def _getConfigStorageRootPath(cls):
-        # Equals ~/.config/<appname> on Linux, ~/Library/Preferences/<appname> on OSX, ~/AppData/Local/<appname> on Windows.
+        if Platform.isWindows():
+            # Special case for Windows, since Qt treats Config as local data but we want config in Roaming.
+            return os.path.normpath(QStandardPaths.writableLocation(QStandardPaths.AppDataLocation))
+
+        # Equals ~/.config/<appname> on Linux, ~/Library/Preferences/<appname> on OSX
         return os.path.normpath(QStandardPaths.writableLocation(QStandardPaths.AppConfigLocation))
 
     # Returns the path where we store different versions of app data
--- a/UM/Resources.py
+++ b/UM/Resources.py
@@ -7,6 +7,8 @@ import re
 import shutil
 from typing import List
 
+from PyQt5.QtCore import QStandardPaths
+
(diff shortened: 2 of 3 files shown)


fieldOfView:
> On Windows, instead of storing preferences in AppData/Roaming, we store them in AppData/Local. QStandardPaths only uses Roaming for the application data.

I can read this in two ways. Are you describing the situation before or after the PR?

In order to support itinerant profiles for users on corporate/education networks, data must be stored in Roaming appdata. "Itinerant profiles" are a thing where AppData/Roaming is "mapped" to a networkdrive.

awhiemstra:
Preferences in this context is the "cura.cfg" file. The rest I consider "application data", so things like machines etc. will be stored in Roaming.

awhiemstra:
To clarify:

```
| Item               | Before PR                                                                                   | After PR                                                                              |
|                    | Windows                    | OSX                                      | Linux               | Windows                    | OSX                                | Linux               |
------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------
| Preferences        | ~/AppData/Roaming/cura     | ~/Library/Application Support/cura       | ~/.config/cura      | ~/AppData/Local/cura       | ~/Library/Preferences/cura         | ~/.config/cura      |
| Machines, Profiles | ~/AppData/Roaming/cura     | ~/Library/Application Support/cura       | ~/.local/share/cura | ~/AppData/Roaming/cura     | ~/Library/Application Support/cura | ~/.local/share/cura |
| Cache              | ~/AppData/Local/cura/cache | ~/Library/Application Support/cura/cache | ~/.cache/cura       | ~/AppData/Local/cura/cache | ~/Library/Caches/cura              | ~/.cache/cura       |
```
(Stupid github doesn't allow markdown tables :( )

fieldOfView:
I love the ascii art table though ;-)

Having the cfg in Local would still be a problem for itinerant profiles. This would mean that if a user moves to another machine, that machine would not know there is an active machine, and pop up the Add Printer dialog (which then adds another copy, of the machine instance, even though the previous instance is still there). It also makes it harder to copy an entire configuration to another machine, because you have to copy files from/to two locations.

I would say cura.cfg should go in the same folder as the Machines, Profiles etc folders. You could argue that configuration != preferences

awhiemstra:
I am actually not entirely sure why Qt defines Config as Local. Microsoft's documentation on this is not very clear though, about the only thing they mention is "machine specific stuff should be in local, otherwise put it in roaming". I mostly wanted to avoid having to special-case things for a platform again.

To be fair, there is a bunch of machine specific stuff in the configuration file that may cause problems with roaming profiles. But I can also see that there is a bunch of things that would be nice to have shared. So I'll add special casing again.

awhiemstra:
Though I do believe we should actually fix the machine selection bug as a separate issue.

fieldOfView:
Agreed, if no machine is set or the machine is unparsable, Cura should try to fall back to another machine and show a message.

ChrisTerBeke:
Will this cause regression for version upgrades or other path specific code?

Ghostkeeper:
Developers, see issue CURA-4710.

Ghostkeeper:
It's not working properly for me on Linux. This is being printed at the very beginning:

    [(140180171147072)-MainThread] UM.Resources.__initializeStoragePaths [336]: Initializing storage paths
    [(140180171147072)-MainThread] UM.Resources.__initializeStoragePaths [346]: Config storage path is /home/ruben/.config/cura_app.py/master
    [(140180171147072)-MainThread] UM.Resources.__initializeStoragePaths [349]: Data storage path is /home/ruben/.local/share/cura_app.py/master
    [(140180171147072)-MainThread] UM.Resources.__initializeStoragePaths [352]: Cache storage path is /home/ruben/.cache/cura_app.py/master

So it looks like it thinks that the application name is `cura_app.py` instead of `cura`. As a result it doesn't load the configuration. Maybe this is the result of running from source, but running from source should still allow you to load the configuration.

ianpaschal:
This doesn't seem compatible anymore either. I'm getting `ModuleNotFoundError: No module named 'UM.Settings.PropertyEvaluationContext'`.

ianpaschal:
Just spoke with Mark and since this is not an quick change I'm going to close this.

Sleep well, sweet PR. For what it's worth, you lived a very long, drawn out life.

PR #248, opened 2017-05-05
CURA-3756: Allow deserialization not to connect signals for container stack detection

LipuFei:
IMO Code is not very nice. We may need some discussion to figure out a better solution.

### DIFF
--- a/UM/Settings/ContainerStack.py
+++ b/UM/Settings/ContainerStack.py
@@ -311,7 +311,11 @@ class ContainerStack(QObject, ContainerInterface, PluginObject):
 
     ##  \copydoc ContainerInterface::deserialize
     #
-    #   Reimplemented from ContainerInterface
+    #   Reimplemented from ContainerInterface. Disconnects propertyChanged and find container in
+    #   container registry to connect with (the container must already be in the container registry).
+    #
+    #   \param serialized
+    #   \param connect_signals  connect propertyChanged or not. Only disable if you know what you're doing
     #
     #   TODO: Expand documentation here, include the fact that this should _not_ include all containers
     def deserialize(self, serialized, connect_signals=True):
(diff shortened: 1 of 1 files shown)


UM/Settings/ContainerStack.py
```diff
@@ -314,7 +314,7 @@ def getVersionFromSerialized(self, serialized: str) -> Optional[int]:
     #   Reimplemented from ContainerInterface
     #
     #   TODO: Expand documentation here, include the fact that this should _not_ include all containers
-    def deserialize(self, serialized):
+    def deserialize(self, serialized, connect_signals=True):
```
awhiemstra: This adds a boolean trap. At the very least make it `def deserialize(self, serialized, *, connect_signals = True)` since that would force it to always be a keyword argument. However, it also breaks with the ContainerInterface interface.

UM/Settings/ContainerStack.py
```diff
@@ -335,29 +336,35 @@ def deserialize(self, serialized):
...
-                    raise Exception("When trying to deserialize %s, we received an unknown ID (%s) for container" % (self._id, container_id))
-
-        elif parser.has_option("general", "containers"):
-            # Backward compatibility with 2.3.1: The containers used to be saved in a single comma-separated list.
-            container_string = parser["general"].get("containers", "")
-            Logger.log("d", "While deserializeing, we got the following container string: %s", container_string)
-            container_id_list = container_string.split(",")
-            for container_id in container_id_list:
-                if container_id != "":
+        if connect_signals:
```
awhiemstra: So if connect_signals is false, we do not add the containers and thus create an improperly deserialized container?

LipuFei: Ok, I found another hack, so I'm abandoning this one.

UM/Settings/ContainerStack.py
nallath: Our rule regarding bool traps is to always use them keyed right?

awhiemstra: Yes, but with the * in between you can enforce this.

PR #251, opened 2017-05-20
Wireframe view

BagelOrb:
Changes needed to support edge-overhang for wireframe view.

Note that I haven't been able to get edge-overhang to work, so this PR is not needed for the linked Cura PR.

### DIFF
--- a/UM/View/RenderBatch.py
+++ b/UM/View/RenderBatch.py
@@ -49,6 +49,10 @@ class RenderBatch():
         Triangles = 0x0004
         TriangleStrip = 0x0005
         TriangleFan = 0x0006
+        LinesAdjacency = 0x000a
+        LineStripAdjacency = 0x000b
+        TrianglesAdjacency = 0x000c
+        TriangleStripAdjacency = 0x000d
 
     ##  Blending mode.
     class BlendMode:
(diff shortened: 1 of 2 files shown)


Ghostkeeper:
Devs, see CURA-3849.

UM/View/RenderBatch.py
```diff
@@ -288,10 +292,12 @@ def _renderItem(self, item):
...
         if mesh.hasIndices():
             if self._render_range is None:
+                element_count = mesh.getFaceCount();
                 if self._render_mode == self.RenderMode.Triangles:
-                    self._gl.glDrawElements(self._render_mode, mesh.getFaceCount() * 3 , self._gl.GL_UNSIGNED_INT, None)
-                else:
-                    self._gl.glDrawElements(self._render_mode, mesh.getFaceCount(), self._gl.GL_UNSIGNED_INT, None)
+                    element_count = mesh.getFaceCount() * 3
+                elif self._render_mode == self.RenderMode.TrianglesAdjacency:
+                    element_count = mesh.getFaceCount() * 3 * 4
```
LipuFei: Shall we still have the `else` at the end? Then other cases won't be affected. Like this:
```
else:
    element_count = mesh.getFaceCount()
```

UM/View/Renderer.py
```diff
@@ -12,11 +12,6 @@
 #   objects should be rendered but the actual rendering process happening after a
 #   sorting step.
 class Renderer():
-    RenderTriangles = 1
```
LipuFei: Why are those removed btw?

LipuFei: Ah. We now use `RenderMode`.

UM/View/RenderBatch.py
LipuFei: Ok, that's done before the `if`s. Missed that part

PR #252, opened 2017-05-24
Theme inheritance

fieldOfView:
This PR implements simple theme inheritance. Themes no longer have to be complete; they can simply inherit a default theme.

A very minimal theme that only changes the background of the sidebar to read could be as simple as this theme.json file:
```
{
    "metadata": {
        "name": "Red Sidebar",
        "inherits": "cura"
    },
    "colors": {
        "sidebar": [255, 0, 0, 255]
    }
}
```

In addition to this inheritance, the PR also adds a small function to list all themes in the relevant resource locations (ie: both in the application folder and in the configuration folder). For this to be useful, a small change in Cura is required. See the separate Cura PR: https://github.com/Ultimaker/Cura/pull/1878

### DIFF
--- a/UM/Application.py
+++ b/UM/Application.py
@@ -101,6 +101,12 @@ class Application:
 
         self._operation_stack = OperationStack(self.getController())
 
+        # Ensure configuration folder has a folder for themes
+        try:
+            os.makedirs(os.path.join(Resources.getStoragePath(Resources.Resources), "themes"))
+        except OSError:
+            pass
+
         self._plugin_registry = PluginRegistry.getInstance()
 
         self._plugin_registry.addPluginLocation(os.path.join(Application.getInstallPrefix(), "lib", "uranium"))
--- a/UM/Application.py
+++ b/UM/Application.py
@@ -101,12 +101,6 @@ class Application:
 
         self._operation_stack = OperationStack(self.getController())
 
-        # Ensure configuration folder has a folder for themes
-        try:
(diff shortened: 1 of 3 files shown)


Ghostkeeper:
Devs, see CURA-3873.

UM/Application.py
```diff
@@ -101,6 +101,12 @@ def __init__(self, name: str, version: str, build_type: str = "", **kwargs):
 
         self._operation_stack = OperationStack(self.getController())
 
+        # Ensure configuration folder has a folder for themes
+        try:
+            os.makedirs(os.path.join(Resources.getStoragePath(Resources.Resources), "themes"))
```
awhiemstra: This should not be needed as Resources.getStoragePath() already does makedirs() if the directory does not exist.

See https://github.com/Ultimaker/Uranium/blob/master/UM/Resources.py#L167

fieldOfView: I'll test without it and will see if the folder gets created. I added it preemptively because I remembered people saying they could not find the "plugins" folder in the configuration folder.

fieldOfView: No, the folder does not get created without explicitly creating it here.

fieldOfView: Ah, I see... The folder does not get created because ```Themes``` is not in ```Resource.__types_storage```. Which would be more preferable, creating the folder manually, or insinuating ```themes``` is a storage folder?

Ghostkeeper: Seeing as how themes are listed dynamically now instead of being hard-coded, it feels to me like they are ordinary resources and should be put in the `Resources` class.

PR #257, opened 2017-06-09
Make hull using points returned from ConvexHull

marcio-ao:
This change is a no-op when using the original scipy.spatial.ConvexHull, as that implementation will return the original array in hull_result.points.

However, this change makes the code more robust and will make it operate correctly even if the underlying library returns a changed points array. I ran into this issue when creating my own version of "scipy.spatial.ConvexHull" that used a different underlying algorithm.


Ghostkeeper:
Devs, see CURA-3924.

nallath:
Thanks for the contribution!

PR #258, opened 2017-06-10
Prevent loading containers from "old" folders

fieldOfView:
This PR prevents loading containers from "old/" folders. These files are backups from the version upgrade system. 

This fixes https://github.com/Ultimaker/Cura/issues/1902, https://github.com/Ultimaker/Cura/issues/1855, https://github.com/Ultimaker/Cura/issues/1874 and many others in the forum

See https://github.com/Ultimaker/Cura/issues/1902#issuecomment-305257112

### DIFF
--- a/UM/Settings/ContainerRegistry.py
+++ b/UM/Settings/ContainerRegistry.py
@@ -141,7 +141,7 @@ class ContainerRegistry(ContainerRegistryInterface):
     #   that were already added when the first call to this method happened will not be re-added.
     def load(self) -> None:
         files = []
-        old_file_expression = re.compile("\\" + os.sep + "old\\" + os.sep + "\d+\\" + os.sep)
+        old_file_expression = re.compile(r"\{sep}old\{sep}\d+\{sep}".format(sep = os.sep))
 
         for resource_type in self._resource_types:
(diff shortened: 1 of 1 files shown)


Ghostkeeper:
Devs, see CURA-3923.

UM/Settings/ContainerRegistry.py
```diff
@@ -141,6 +141,8 @@ def getEmptyInstanceContainer(self) -> InstanceContainer:
     #   that were already added when the first call to this method happened will not be re-added.
     def load(self) -> None:
         files = []
+        old_file_expression = re.compile("\\" + os.sep + "old\\" + os.sep + "\d+\\" + os.sep)
```
awhiemstra: I'd prefer to use `re.compile(r"\{sep}old\{sep}\d+\{sep}".format(sep = os.sep)` here since this makes it less of a string puzzle. Additionally it avoids needing to escape the slashes due to using "raw string" mode.

fieldOfView: Agreed, much better.

fieldOfView:
Whoops, rebased against wrong branch.

awhiemstra:
Thanks for the quick fix, merged.

PR #259, opened 2017-06-13
Added shortcuts for MoveTool(Q) ScaleTool(A) RotateTool(Z)

Appesteijn:
As was requested on our forum, I made it possible for Tools to have a shortcut-key

CURA-3933

### DIFF
--- a/UM/Controller.py
+++ b/UM/Controller.py
@@ -12,7 +12,6 @@ from UM.View.View import View
 from UM.InputDevice import InputDevice
 from typing import Optional, Dict
 
-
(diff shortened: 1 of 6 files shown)


fieldOfView:
No love for the Mirror or Per Model Settings tools?

The number keys could be nice shortcuts for extruder assignment.

nallath:
I'd see this more as a proof of concept. We might want to merge this, but in the future i'd love to see some proper shortcut handling. That way plugins could link an action to a prefered shortcut and we can have a shortcut manager to re-bind them (and of course show conflicts)

Appesteijn:
I have added W for the mirrortool.
ALT-1 and ALT-2 are already shortcut keys for extruder selection.

PR #261, opened 2017-06-22
PointingRectangle border

fieldOfView:
This PR implements a true border for UM.PointingRectangle. Since I introduced the PointingRectangle it has always irked me that I needed to add two PointingRectangles to draw one with a border, and hope the second one always nicely aligns with the first one. Now a PointingRectangle can draw its own border, like it should.

### DIFF
--- a/UM/Qt/Bindings/PointingRectangle.py
+++ b/UM/Qt/Bindings/PointingRectangle.py
@@ -105,6 +105,8 @@ class PointingRectangle(QQuickItem):
         target_offset = self._target - QPoint(self.x(), self.y())
 
         arrow_on_side = -1 # no arrow
+        arrow_size = 0
+        arrow_offset = 0
 
         if target_offset.x() >= 0 and target_offset.x() <= self.width():
             arrow_size = min(self._arrow_size, self.width()/2)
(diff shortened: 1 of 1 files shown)


fieldOfView:
Thinking about PointingRectangle, shouldn't it be moved into Cura instead of Uranium?

nallath:
Argus also uses the PointingRectangle, so it's fine to leave it there.

Ghostkeeper:
Devs, see CURA-3994.

fieldOfView:
See https://github.com/Ultimaker/Cura/pull/2059, which depends on this PR.

BagelOrb:
Can somebody explain to me how a rectangle can be pointing?!
It's too symmetrical to be an arrow!

fieldOfView:
1rycmk

The PointingRectangle has been a thing since before Cura 2.1. It is a rectangle with a triangle affixed to it. But RectangleWithPointingTriangleAffixedToIt is such a long name.

nallath:
I think we can't use Rectangle+Triangle as a class name either.

BagelOrb:
I wasn't sure if I was serious or trolling myself. Is both possible?
What about ClassForRepresentingNearlyRectangularBoxesIfItWereNotForTheLittleTriangleThatIsAffixedToIt?

PR #264, opened 2017-06-27
Added zoom to mouse functionality

nallath:
Based on the request https://github.com/Ultimaker/Cura/issues/1572

See also https://github.com/Ultimaker/Cura/pull/2037


Ghostkeeper:
Devs, see CURA-3587.

Ghostkeeper:
Zooming out doesn't work any more. I can only zoom in.

nallath:
I have no idea why this didn't work. I'm very sure I tested it. Oh well, it should be correct again.

PR #266, opened 2017-07-14
Fix system text rendering

fieldOfView:
This PR fixes text rendering in system-themed qml by rounding fractional widths and heights. 

Depending on window/dialog sizes, Label elements could be rendered at fractional pixel offsets. The Text.NativeRendering renderType does not handle that nicely, leading to ugly font rendering on Windows and OSX. In an attempt to fix this, some Labels were switched to Text (which uses Text.QtRendering rendertype). Unfortunately this leads to font-rendering that is distinctively non-native on Windows. Furthermore, Text does not use the system-themed colors used by Label, leading to unreadable text in dark-themed systems (black on black).

By rounding all widths/heights to whole pixels, Text.NativeRendering can be used without rendering problems.

See https://github.com/Ultimaker/Cura/issues/1748 for discussion.  

A followup Cura PR is necessary to fix all cases of fractional offsets and all uses of Text vs Label

### DIFF
--- a/UM/Qt/qml/UM/Preferences/MachinesPage.qml
+++ b/UM/Qt/qml/UM/Preferences/MachinesPage.qml
@@ -36,8 +36,8 @@ ManagementPage {
             elide: Text.ElideRight
         }
 
-        Text { text: catalog.i18nc("@label", "Type"); width: parent.width * 0.2; }
-        Text { text: base.currentItem && base.currentItem.typeName ? base.currentItem.typeName : ""; width: parent.width * 0.7; }
+        Text { text: catalog.i18nc("@label", "Type"); width: (parent.width * 0.2) | 0; }
+        Text { text: base.currentItem && base.currentItem.typeName ? base.currentItem.typeName : ""; width: (parent.width * 0.7) | 0; }
 
         UM.I18nCatalog { id: catalog; name: "uranium"; }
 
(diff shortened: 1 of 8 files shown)


nallath:
CURA-4048

Ghostkeeper:
The `| 0` is meant to be rounding, I suppose.

We're going to forget about that every time.

fieldOfView:
Then let's not even try fixing the issue, and let's keep ignoring the fact that some text is currently rendered unreadable in dark system themes.

awhiemstra:
Have you tried what happens if you change the text rendering type to Text.NativeRendering? See http://doc.qt.io/qt-5/qml-qtquick-text.html#renderType-prop .

I am mostly asking because in my opinion this should be something that should be handled correctly by Qt, instead of needing workarounds on the application side.

fieldOfView:
Label already defaults to Text.NativeRendering, and that has the problem of not rendering nicely when rendered on fractional pixel offsets. Have I tried rendering at Text.QtRendering? Yes, but that looks succinctly different than native text on Windows, to the point that it looks like an entirely different font.

fieldOfView:
Note that I do agree that this should really be fixed in Qt, but until it is we have to work around this issue. 

Instead of examining each occurrence of Text/Label rendering and applying a fix only where needed, I decided to apply a general fix for all occurrences. It won't break if you forget the ```| 0``` once, but it results in better rendering if you include it consistently.

PR #267, opened 2017-07-18
Hide empty categories

fieldOfView:
This PR prevents showing empty category headers in the sidebar.  SettingDefinitionModel:_isAnyDescendantVisible() only checked if there are settings that are marked to be "visible" (in the Settings Visibility page), but not if they are "enabled" for the current machine. 

eg: "Enable Prime Tower" is a setting that is made visible by default, but it is disabled (hidden) with "enabled": "machine_extruder_count > 1". So you get an empty category for all single extrusion printers.

This PR fixes https://github.com/Ultimaker/Cura/issues/2126

The performance impact of this change should be checked.


nallath:
CURA-4055

PR #273, opened 2017-08-10
Pirate Speak

Ghostkeeper:
This be implementin' a "Pirate" speak in the spirit of Talk Like a Pirate Day. Created by cap'n Awhiemstra and me.

The Pirate speak be auto-generated when th' `extract-messages` build target be built. This then be callin' a Python scribble that takes the newly extracted .pot file and generates a .po file with translated strings.

The Python scribble be parsin' the .pot file, pullin' them messages from it and storin' the appropriate context and plural form. This extraction be aimed at landlubber English; thar be no support fer more than a pair o' plural forms. Th' extracted messages be keelhauled through a translation dictionary, where the `\b` regex fer word boundary be added around it, an' each word in the dictionary be looted-and-replaced. The case of the first character of the original text be preserved, soeth the pirateofdoom.py dictionary not be case sensitive fer the first characters of yonder texts. A new .po file then be written with a hard-coded header.

How the Pirate Speak be used be dependin' on the application that uses Uranium, but it be intended as an E@ster egg.

We be requestin' to get this commandeered into 2.7, to be on time for Talk Like a Pirate Day. Appestijn gave verbal permission for this, given the low risk, may the Pirate Code be holdin' the cuss t'his word.

Commandeer requests to keep in yer spyglass too:

* https://github.com/Ultimaker/Cura/pull/2242
* https://github.com/Ultimaker/cura-binary-data/pull/7


scripts/pirate.py
```diff
@@ -0,0 +1,102 @@
+#Creates the Pirate translation files.
+
+import sys #To get command line arguments.
+import pirateofdoom #Contains our translation dictionary.
```
daid: This must be the most obscure reference to the dictionary-of-doom ever :)

Ghostkeeper: In hindsight I should've called it a "code" or "codex" instead.

PR #275, opened 2017-08-23
CURA-4186 Add stack context for function evaluation

LipuFei:
CURA-4186

### DIFF
--- a/UM/Settings/ContainerStack.py
+++ b/UM/Settings/ContainerStack.py
@@ -13,7 +13,8 @@ from UM.PluginObject import PluginObject
 from UM.Logger import Logger
 from UM.MimeTypeDatabase import MimeTypeDatabase, MimeType
 from UM.Settings.DefinitionContainer import DefinitionContainer #For getting all definitions in this stack.
-from UM.Settings.Interfaces import ContainerInterface, ContainerRegistryInterface, PropertyEvaluationContext
+from UM.Settings.Interfaces import ContainerInterface, ContainerRegistryInterface
+from UM.Settings.PropertyEvaluationContext import PropertyEvaluationContext
 from UM.Settings.SettingFunction import SettingFunction
 from UM.Settings.Validator import ValidatorState
 
--- a/UM/Settings/ContainerStack.py
+++ b/UM/Settings/ContainerStack.py
@@ -13,7 +13,7 @@ from UM.PluginObject import PluginObject
 from UM.Logger import Logger
 from UM.MimeTypeDatabase import MimeTypeDatabase, MimeType
 from UM.Settings.DefinitionContainer import DefinitionContainer #For getting all definitions in this stack.
-from UM.Settings.Interfaces import ContainerInterface, ContainerRegistryInterface
+from UM.Settings.Interfaces import ContainerInterface, ContainerRegistryInterface, PropertyEvaluationContext
 from UM.Settings.SettingFunction import SettingFunction
 from UM.Settings.Validator import ValidatorState
 
@@ -194,10 +194,14 @@ class ContainerStack(QObject, ContainerInterface, PluginObject):
     #   Note that if the property value is a function, this method will return the
     #   result of evaluating that property with the current stack. If you need the
     #   actual function, use getRawProperty()
-    def getProperty(self, key: str, property_name: str):
+    def getProperty(self, key: str, property_name: str, context: Optional[PropertyEvaluationContext] = None):
         value = self.getRawProperty(key, property_name)
         if isinstance(value, SettingFunction):
-            return value(self)
+            if context is None:
+                context = PropertyEvaluationContext(self)
+            context.pushContainer(self)
+            value = value(self, context)
+            context.popContainer()
 
         return value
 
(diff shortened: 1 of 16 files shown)


UM/Settings/Interfaces.py
```diff
@@ -1,11 +1,40 @@
...
 
 import UM.Decorators
 from UM.Signal import Signal
 
+##  Context for evaluating a property value
+#   It contains:
+#     1. a stack of containers during the evaluation in the function call stack fashion
+#     2. a context dictionary which contains all the current context
+#
+class PropertyEvaluationContext:
```
Ghostkeeper: Why is this in `Interfaces.py`? There is nothing interface about this.

UM/Settings/Interfaces.py
```diff
@@ -1,11 +1,40 @@
...
+
+    def __init__(self, source_stack = None):
+        self._stack_of_containers = deque()
+        if source_stack is not None:
+            self._stack_of_containers.append(source_stack)
+        self._context = {}
+
+    @property
+    def root_stack(self):
```
Ghostkeeper: Code style: Function names should be lower camel case.

UM/Settings/Interfaces.py
```diff
@@ -1,11 +1,40 @@
...
+        if self._stack_of_containers:
+            return self._stack_of_containers[0]
+
+    @property
+    def stack_of_containers(self):
```
Ghostkeeper: Code style: Function names should be lower camel case.

UM/Settings/ContainerStack.py
```diff
@@ -194,10 +194,14 @@ def setDirty(self, dirty: bool) -> None:
     #   Note that if the property value is a function, this method will return the
     #   result of evaluating that property with the current stack. If you need the
     #   actual function, use getRawProperty()
-    def getProperty(self, key: str, property_name: str):
+    def getProperty(self, key: str, property_name: str, context: Optional[PropertyEvaluationContext] = None):
         value = self.getRawProperty(key, property_name)
         if isinstance(value, SettingFunction):
-            return value(self)
+            if context is None:
+                context = PropertyEvaluationContext(self)
```
Ghostkeeper: You don't need to create a PropertyEvaluationContext here because the `__call__` of `SettingFunction` accepts `None` as context. It would save some time, as this function is very time-critical.

UM/Settings/Interfaces.py
```diff
@@ -1,11 +1,40 @@
...
 
+##  Context for evaluating a property value
+#   It contains:
+#     1. a stack of containers during the evaluation in the function call stack fashion
+#     2. a context dictionary which contains all the current context
+#
+class PropertyEvaluationContext:
+
+    def __init__(self, source_stack = None):
+        self._stack_of_containers = deque()
```
Ghostkeeper: I'd suggest just making this field public rather than exposing it via a load of functions.

UM/Settings/Interfaces.py
```diff
@@ -1,11 +1,40 @@
...
+        if source_stack is not None:
+            self._stack_of_containers.append(source_stack)
+        self._context = {}
+
+    @property
```
Ghostkeeper: Why are these `@property`? That annotation is unused.

UM/Settings/Interfaces.py
```diff
@@ -1,11 +1,40 @@
...
+    def root_stack(self):
+        if self._stack_of_containers:
+            return self._stack_of_containers[0]
+
+    @property
```
Ghostkeeper: Why are these `@property`? That annotation is unused.

UM/Settings/Interfaces.py
```diff
@@ -35,6 +35,10 @@ def pushContainer(self, container):
     def popContainer(self):
         return self._stack_of_containers.pop()
 
+    @property
+    def context(self):
+        return self._context
```
Ghostkeeper: I'd suggest just making the field public rather than this construction.

Once you've done that it appears that the PropertyEvaluationContext is just a namedtuple for the context dictionary and the container stack.

UM/Settings/Interfaces.py
LipuFei: Fixed in 068f7f6e05f5422971036b5fabe7a74001901ba5

UM/Settings/Interfaces.py
LipuFei: Fixed in bf9acbc3a227aaf4be1ef9873ef0b69926e51468

UM/Settings/Interfaces.py
LipuFei: Fixed in bf9acbc3a227aaf4be1ef9873ef0b69926e51468

UM/Settings/Interfaces.py
LipuFei: Fixed in bf9acbc3a227aaf4be1ef9873ef0b69926e51468

UM/Settings/Interfaces.py
LipuFei: Fixed in bf9acbc3a227aaf4be1ef9873ef0b69926e51468

UM/Settings/Interfaces.py
LipuFei: Fixed in bf9acbc3a227aaf4be1ef9873ef0b69926e51468

UM/Settings/ContainerStack.py
LipuFei: Fixed in 8a93ee003

Issue #277, opened 2017-08-30
jp is not a valid locale name


hroncok:
When I'm trying to package this into Fedora, some tools yell at me that `jp` is not a valid locale name. I believe that it should instead be `ja`.

See https://github.com/Ultimaker/Uranium/issues/246 for similar problem

Issue #246, opened 2017-04-28
ptbr is not a valid locale name


hroncok:
When I'm trying to package this into Fedora, some tools yell at me that `ptbr` is not a valid locale name. I believe that it should instead be `pt_BR`.

cc @Patola

Ghostkeeper:
Our project manager just closed this bug because it's more than 3 months old.

hroncok:
Still valid.

PR #278, opened 2017-08-30
Relocate locales to proper locations

hroncok:
Fixes https://github.com/Ultimaker/Uranium/issues/246
Fixes https://github.com/Ultimaker/Uranium/issues/277

See https://github.com/Ultimaker/Cura/pull/2342
See https://github.com/Ultimaker/cura-binary-data/pull/8


nallath:
For our reference; CURA-4263

nallath:
You forgot to do the same for https://github.com/Ultimaker/cura-binary-data

hroncok:
Oh, I didn't know about that one. Will do.

hroncok:
https://github.com/Ultimaker/cura-binary-data/pull/8

Ghostkeeper:
This is a bit more involved, because the language code is also saved in the preferences, and that would mean that the preferences files will have to be upgraded with a version upgrade.

Perhaps it makes sense to change the rest of them also to the xx_XX format instead of leaving the country code out for ones that are not ambiguous.

alekseisasin:
I agree with Ghostkeeper, we need to keep same template for all translations and change rest. 
According to the ISO language_Country codes : https://docs.oracle.com/cd/E13214_01/wli/docs92/xref/xqisocodes.html

German       -> (de_DE)
English        -> (en_EN)
Spanish       -> (es_ES)
Finnish        -> (fi_FI)
French        -> (fr_FR)
Italian          -> (it_IT)
Japanese    -> (ja_JP)
Korean        -> (ko_KR)
Dutch          -> (nl_NL)
Polish          -> (pl_PL)
Portuguese -> (pt_BR)  - Brasil
Russian       -> (ru_RU)
Turkish        -> (tr_TR)
Chinese       -> (zh_CN)

hroncok:
> Perhaps it makes sense to change the rest of them also to the xx_XX format instead of leaving the country code out for ones that are not ambiguous.

It would work, but I don't see the real benefit here.

alekseisasin:
It will help keep structure clear.

hroncok:
OK. My main concern is that all the codes are valid. Other than that, I don't really care.

Ghostkeeper:
I think I'll make the rest of those changes now then, since hroncok has graciously allowed commits to his PR branch :)

PR #285, opened 2017-09-26
Fix screen scale factor

fieldOfView:
This PR attempts to solve the Hi DPI issues on both OSX and Windows. OSX apparently no longer needs the Screen.devicePixelRatio workaround when specifying pixel sizes, but Windows certainly does.

This PR is for testing an approach where instead of using Screen.devicePixelRatio, we use a Uranium-supplied screenScaleFactor. This context variable was introduced a couple of releases ago, but never used.

When testing, also use the related Cura PR (https://github.com/Ultimaker/Cura/pull/2484), as most occurrences where a fix is needed are in the Cura repo.

For additional discussion see https://github.com/Ultimaker/Cura/commit/7bc715a9bd5e61efcfef4b54f7d004b3f6248f5a#commitcomment-24539576

### DIFF
--- a/UM/Qt/QtApplication.py
+++ b/UM/Qt/QtApplication.py
@@ -395,7 +395,7 @@ class QtApplication(QApplication, Application):
         if sys.platform == "darwin":
             return 1.0
         else:
-            return 1.5 #self.devicePixelRatio()
+            return self.devicePixelRatio()
 
     def _getDefaultLanguage(self, file):
         # If we have a language override set in the environment, try and use that.
--- a/UM/Qt/QtApplication.py
+++ b/UM/Qt/QtApplication.py
@@ -390,10 +390,12 @@ class QtApplication(QApplication, Application):
         return QSplashScreen(QPixmap(Resources.getPath(Resources.Images, self.getApplicationName() + ".png")))
 
     def _screenScaleFactor(self):
-        physical_dpi = QGuiApplication.primaryScreen().physicalDotsPerInch()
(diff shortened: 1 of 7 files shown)


UM/Qt/QtApplication.py
```diff
@@ -390,10 +390,12 @@ def _createSplashScreen(self):
...
-        physical_dpi = QGuiApplication.primaryScreen().physicalDotsPerInch()
-        # Typically 'normal' screens have a DPI around 96. Modern high DPI screens are up around 220.
-        # We scale the low DPI screens with a traditional 1, and double the high DPI ones.
-        return 1.0 if physical_dpi < 150 else 2.0
+        # OSX handles sizes of dialogs behind our backs, but other platforms need
+        # to know about the device pixel ratio
+        if sys.platform == "darwin":
+            return 1.0
+        else:
+            return 1.5 #self.devicePixelRatio()
```
Appesteijn: Why did you hardcode 1.5?

fieldOfView: Hmm, strange, I committed test code. My bad.

fieldOfView:
It does not work correctly yet on Windows (but I am interested to hear if it did not break stuff on OSX). Will continue to work on this.

fieldOfView:
I have changed the way the screenScaleFactor is determined, and it now works better than ever on Windows. OSX should be unaffected (because screenScaleFactor is set to 1), but I am interested in the result for Linux distros.

The specific current implementation is open for discussion, since it contains a magic number.

PR #287, opened 2017-10-03
set the font size to points instead of pixels

Appesteijn:
This should fix the garbled character on MacOS. Font sizes will now be calculated in points instead of pixels.

### DIFF
--- a/UM/Qt/Bindings/Theme.py
+++ b/UM/Qt/Bindings/Theme.py
@@ -33,7 +33,7 @@ class Theme(QObject):
             QCoreApplication.instance().setFont(default_font)
 
         self._em_height = int(QFontMetrics(QCoreApplication.instance().font()).ascent())
-        self._em_width = self._em_height;
+        self._em_width = self._em_height
 
         self._initializeDefaults()
 
--- a/UM/Qt/Bindings/Theme.py
+++ b/UM/Qt/Bindings/Theme.py
@@ -184,6 +184,7 @@ class Theme(QObject):
                     QFontDatabase.addApplicationFont(os.path.join(fontsdir, file))
 
         if "fonts" in data:
+            system_font_size = QCoreApplication.instance().font().pointSize()
             for name, font in data["fonts"].items():
                 f = QFont()
                 f.setFamily(font.get("family", QCoreApplication.instance().font().family()))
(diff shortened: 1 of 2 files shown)


fieldOfView:
As expected, the text sizes are changed considerably by this PR. This is on Windows 10, font size at 100%:

Before this PR:
image

With this PR:
image

Appesteijn:
SHould it then also be modified by the screenScaleFactor ?

fieldOfView:
No, I think it should just not use _em_height, but a fixed number instead. I think 11 for Windows, but have not tried that. See eg http://reeddesign.co.uk/test/points-pixels.html

Note: switching back to Labels (ie: switching back to native font rendering) does mean that all texts need to be drawn at whole pixel intervals too. I have done that for system-themed areas but not for "uranium-themed" areas.

fieldOfView:
```f.setPointSize(font.get("size", 1) * 9)``` seems to be correct for Windows, but I am afraid that might be platform-dependent.

**Comment by awhiemstra** [comment]
Note that that would completely disregard system font size, which would be a shame. However, there is a fairly simple fix. The _em_height is already determined from the system font size. What you need to do is reuse the system font size in points as the value for fonts, so that a font of "1" becomes for example 10pt when my system font is 10pt.

**Comment by fieldOfView** [comment]
Good one, @awhiemstra. How about this:
```
        if "fonts" in data:
            system_font_size = QCoreApplication.instance().font().pointSize()
            for name, font in data["fonts"].items():
                f = QFont()
                f.setFamily(font.get("family", QCoreApplication.instance().font().family()))

                f.setBold(font.get("bold", False))
                f.setLetterSpacing(QFont.AbsoluteSpacing, font.get("letterSpacing", 0))
                f.setItalic(font.get("italic", False))
                f.setPointSize(font.get("size", 1) * system_font_size)
                f.setCapitalization(QFont.AllUppercase if font.get("capitalize", False) else QFont.MixedCase)

                self._fonts[name] = f
```

fieldOfView:
Though I think the solution with setPointSize is probably more "correct", I now think an easier fix for the actual problem that @Appesteijn is trying to fix would be to round the pixel size to a whole pixel value. I don't know if setPointSize handles fractional sizes better than setPixelSize. Rounding may be a good idea regardless (round all the things!)

Appesteijn:
Added @fieldOfView his change. Did do a cast to an int, not sure if that is needed.

PR #291, opened 2017-10-10
create_plugin: create a .umplugin file at least

thopiekar:
Cura 2.7 expects either a .curaplugin or a .umplugin file. This script is creating a .plugin file.
I assume we want a .umplugin file at least.

### DIFF
--- a/create_plugin.py
+++ b/create_plugin.py
@@ -34,7 +34,7 @@ if __name__ == "__main__":
 
     if checkValidPlugin(full_plugin_path):
         plugin_name = os.path.basename(os.path.normpath(full_plugin_path))
-        plugin_file_location = plugin_name + ".curaplugin"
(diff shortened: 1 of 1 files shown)


create_plugin.py
```diff
@@ -34,7 +34,7 @@ def zipDirectory(path, zip_handle):
 
     if checkValidPlugin(full_plugin_path):
         plugin_name = os.path.basename(os.path.normpath(full_plugin_path))
-        plugin_file_location = plugin_name + ".plugin"
+        plugin_file_location = plugin_name + ".curaplugin"
```
LipuFei: Shouldn't this be `.umplugin` instead?

thopiekar: Oops 🤓

nallath:
Derp. Don't know why I didn't do that.

PR #292, opened 2017-10-11
Add evaluation context processing for ContainerStack and SettingFunction

ChrisTerBeke:
CURA-4358

- Make getProperty() and getRawProperty() in ContainerStack to be
  evaluation-context-aware.
- Allow getRawProperty() to skip some containers if this info is
  provided by the context.
- SettingFunction can now override operators with those provided in the
  evaluation context.


ChrisTerBeke:
This PR is for other people than current core devs to comment on this change.

PR #293, opened 2017-10-11
Rounding caused characters to be misaligned

Appesteijn:
Together with Cura PR: https://github.com/Ultimaker/Cura/pull/2586

CURA-4437

### DIFF
--- a/UM/Qt/Bindings/Theme.py
+++ b/UM/Qt/Bindings/Theme.py
@@ -2,7 +2,7 @@
 # Uranium is released under the terms of the LGPLv3 or higher.
 
 from PyQt5.QtCore import QObject, pyqtProperty, pyqtSignal, QCoreApplication, QUrl, QSizeF
-from PyQt5.QtGui import QColor, QFont, QFontMetrics, QFontDatabase, QFontInfo
+from PyQt5.QtGui import QColor, QFont, QFontMetrics, QFontDatabase
 from PyQt5.QtQml import QQmlComponent, QQmlContext
 from UM.FlameProfiler import pyqtSlot
 import json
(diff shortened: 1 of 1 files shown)


ChrisTerBeke:
@fieldOfView do you have any ideas on how to fix this? Using int will fix the 'wavy font' issue on MacOS but causes a lot of issues on Windows. We suspect the issue is still somewhere in conversion from points to actual pixels in the font rendering.

Appesteijn:
This is how it looked before this PR (with the Round function):
screen shot 2017-10-11 at 14 23 12

And this is how it looks on Mac after this PR (wit cast to int):
screen shot 2017-10-11 at 14 21 55

fieldOfView:
Try int(round()) instead of int()

ChrisTerBeke:
int(rount()) works like a charm on my machine.

ChrisTerBeke:
Apparently it does not work on retina macbook....

Appesteijn:
Same output on MacOS as in screenshot above :(

fieldOfView:
The only logical conclusion is that the "fix" proposed in the two PRs is purely symptomatic. I think there is still a fractional offset somewhere that is causing the actual issue, but the font rendering on OSX just so happens to round it the proper way when one (or more) of the other sizes are rounded down by a pixel.

fieldOfView:
Note how the printer dropdown renders the text correctly in both screenshots. It could be either the left margin of the other elements, or a vertical size somewhere.

fieldOfView:
Ok, I think I have found the culprit, but I have no way to check on a retina mac.
https://github.com/Ultimaker/Cura/blob/master/resources/qml/SidebarHeader.qml#L20
This should be rounded, or every item but the first item is going to be on a non-integer y offset.

On a personal note: this is not how you do theming; getting a size from the theme and then multiplying it by a magic number.

Appesteijn:
I fixed that but it didn't solve it

fieldOfView:
In SidebarHeader alone, I see a number of other offenders (height * 2 / 3, width * 0.7)

Appesteijn:
These should now be fixed with a Math.floor()

PR #298, opened 2017-10-27
Reload qml

nallath:
It's far from perfect, as it doesn't actually kill instances (windows keep on getting added, even though close is called).

I expect this is caused by some weird garbage collect issue. Could be that I'm missing something super obvious, so i'm still going to make the pull request.


nallath:
@awhiemstra Could you have a look at this? There are some issues with it (lots of error spam, instances not getting killed), but I'm at the end of my Qt fu mastry here.

Ghostkeeper:
This is implementing a function to reload all QML, but the function is not yet called. Is it your intention to call this when changing the language and theme and such?

nallath:
It's called by Cura by means of an action. See the branch of the same name in Cura.

PR #299, opened 2017-10-31
Use the Uranium logging framework to log QML warnings, instead of using stderr.

daid:
This fixes the problem where QML warnings/errors are not ending up in the log files on windows, making debugging harder, as they end up in the event viewer of windows, and thus are not mixed with our normal log files.

### DIFF
--- a/UM/Qt/QtApplication.py
+++ b/UM/Qt/QtApplication.py
@@ -222,6 +222,8 @@ class QtApplication(QApplication, Application):
         Bindings.register()
 
         self._engine = QQmlApplicationEngine()
+        self._engine.setOutputWarningsToStandardError(False)
+        engine.warnings.connect(self.__onQmlWarning)
 
         for path in self._qml_import_paths:
             self._engine.addImportPath(path)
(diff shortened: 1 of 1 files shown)


nallath:
Isn't it better to also pipe the debug print messages as well?

daid:
Sure, got an example of one of those messages? This just fixed my problem of unable to see why my QML didn't work in my plugin.

nallath:
If you just use print("") in the javascript of a qml it should print it as debug.

daid:
Really? I thought that was console.log(), no idea if you can intercept those however.

nallath:
You can also use console.log, but print is easier to type (I found out when I accidentally did it because my mind was still thinking "in python")

ChrisTerBeke:
As this is not part of the default ES standard I guess it's something that Qt added to their interpreter?

awhiemstra:
It is a property of the QML global object, see http://doc.qt.io/qt-5/qtqml-javascript-qmlglobalobject.html . I also do not know if those are easily intercepted.

Ghostkeeper:
Developers, see issue CURA-4542.

PR #300, opened 2017-11-09
Add a function to create a QML component.

daid:
So plugins no longer have to touch our privates (_engine) and creation is done in the same way for all objects. As there are some notes/tricks to this.
I noticed that you have to keep a reference to the created "context" item, else it gets garbage collected and this causes some QML components to fail (for example buttons no longer respond to clicks).
The extension plugin example does this wrong for example:
https://github.com/Ultimaker/UraniumExampleExtensionPlugin/blob/master/ExampleExtension.py#L58
Which costed me quite some time to figure out why my qml window was misbehaving.

To give some examples for this new function, the function in the example can be replaced with:
```
    def _createDialogue(self):
        return Application.getInstance().createQmlComponent(os.path.join(PluginRegistry.getInstance().getPluginPath(self.getPluginId()), "Hello.qml"))
```
Another example, using the context_properties parameter:
https://github.com/Ultimaker/Cura/blob/master/plugins/ImageReader/ImageReaderUI.py#L81
```
    def _createConfigUI(self):
        if self._ui_view is None:
            Logger.log("d", "Creating ImageReader config UI")
            self._ui_view = Application.getInstance().createQmlComponent(QUrl.fromLocalFile(os.path.join(PluginRegistry.getInstance().getPluginPath("ImageReader"), "ConfigUI.qml"), {"manager", self})
            self._ui_view.setFlags(self._ui_view.flags() & ~Qt.WindowCloseButtonHint & ~Qt.WindowMinimizeButtonHint & ~Qt.WindowMaximizeButtonHint);

            self._disable_size_callbacks = False
```

(Note, I didn't test this function inside Uranium, only inside my plugin, so there could be a bug in this pull request, as I lack a full Cura build environment)

### DIFF
--- a/UM/Qt/QtApplication.py
+++ b/UM/Qt/QtApplication.py
@@ -4,10 +4,11 @@
 import sys
 import os
 import signal
+from typing import Dict, Optional
 
 
 from PyQt5.QtCore import Qt, QCoreApplication, QEvent, QUrl, pyqtProperty, pyqtSignal, pyqtSlot, QLocale, QTranslator, QLibraryInfo, QT_VERSION_STR, PYQT_VERSION_STR
-from PyQt5.QtQml import QQmlApplicationEngine
+from PyQt5.QtQml import QQmlApplicationEngine, QQmlComponent, QQmlContext
 from PyQt5.QtWidgets import QApplication, QSplashScreen, QMessageBox, QSystemTrayIcon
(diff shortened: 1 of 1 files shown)


Ghostkeeper:
Developers, see issue CURA-4568.

nallath:
There were a few issues with the code, but i fixed those (I needed this code, so I reviewed & tested it)

ChrisTerBeke:
Just fixed a bunch of errors in the Cura repo as a result of this.

PR #301, opened 2017-11-09
Fix loading values that start with a semi-colon

fieldOfView:
This PR fixes loading values from INI files when they start with a semicolon, eg comments in start/end gcode.

Caveat: this removes the possibility to add comments in the INI files, since they would be considered part of the values now. In the past 2 years there has not been a single cfg.ini file in the Cura distribution that uses ; for non-value comments. 

Fixes https://github.com/Ultimaker/Cura/issues/1237


Ghostkeeper:
I think that's acceptable. Developers, see issue CURA-4570.

PR #304, opened 2017-11-21
pluginsLoaded: Adding signal and emitting it after last checks

thopiekar:
Useful if plugins depend on each other and it needs actions after they are all loaded.


nallath:
Merged this right away. It's such a small change anyway.

PR #306, opened 2017-11-22
UpdateCheckerJob: Use multiple lines to make code better readable

thopiekar:
Just hit this line and thought about correcting it.
Might be that this is not conform to your coding conventions, but it looks better in my eyes.


Ghostkeeper:
There are no code style conventions about this. I agree with you.

fieldOfView:
I agree that using multiple lines here makes it more readable, but this indentation style is not used anywhere else.

thopiekar:
@fieldOfView If you mean the missing whitespace in the third line before the closing bracket.. I see it now, too 🙄

PR #307, opened 2017-11-22
MeshBuilder: Watching files for modifications

thopiekar:
If a modification has been detected, a message will appear and ask for a reload.
I asked for this feature already months before, but decided to invest time on this on my own now.

Optionally: Adding a preference to turn on/off this feature

Note: This has been coded without testing inside Github's webpage.

### DIFF
--- a/UM/Mesh/MeshBuilder.py
+++ b/UM/Mesh/MeshBuilder.py
@@ -1,18 +1,23 @@
 # Copyright (c) 2015 Ultimaker B.V.
 # Uranium is released under the terms of the LGPLv3 or higher.
 
+from UM.Message import Message
 from UM.Mesh.MeshData import MeshData
 from UM.Mesh.MeshData import MeshType
 from UM.Mesh.MeshData import calculateNormalsFromVertices
 from UM.Mesh.MeshData import calculateNormalsFromIndexedVertices
+from UM.Mesh.ReadMeshJob import ReadMeshJob
 from UM.Math.Vector import Vector
 from UM.Math.Matrix import Matrix
 from UM.Logger import Logger
+
+from PyQt5.QtCore import pyqtProperty, Qt, pyqtSignal, pyqtSlot, QUrl, QTimer
 
 import numpy
 import math
 import numbers
 
+import os
 
 ##  Builds new meshes by adding primitives.
 #
@@ -34,6 +39,15 @@ class MeshBuilder:
         self._file_name = None
         # original center position
         self._center_position = None
+        
+        # File-changed notify timer
+        self._file_changed_timer = QTimer()
+        self._file_changed_timer.setInterval(1000)
+        self._file_changed_timer.setSingleShot(True)
+        self._file_changed_timer.timeout.connect(self._check_file_changed)
+        if self._file_name:
+            self._file_changed_stamp = os.path.getmtime(self._file_name)
+            self._file_changed_timer.start()
 
     ##  Build a MeshData object.
     #
@@ -759,3 +773,33 @@ class MeshBuilder:
         self.addQuad(v0, v1, v2, v3, color=color, normal = normal)
 
         return True
+    
+    def _check_file_changed(self):
+        if os.path.isfile(self._file_name):
+            if self._file_changed_stamp is not os.path.getmtime(self._file_name):
+                 message = Message(i18n_catalog.i18nc("@info", "Would you like to reload {filename}?").format(filename = os.path.split(self._file_name)[0]),
+                                   title = i18n_catalog.i18nc("@info:title", "File has been modified"))
+                 message.addAction("reload",
+                                   i18n_catalog.i18nc("@action:button", "Reload"),
+                                   "[no_icon]",
+                                   "[no_description]")
+                 message.actionTriggered.connect(self._callback)
+                 message.show()
+            
+        self._container_change_timer.start()
+    
+    def _onActionTriggered(self, message, action):
+        if action == "reload":
+            if os.path.isfile(self._file_name):
+                job = ReadMeshJob(self._file_name)
+                job._node = self
+                job.finished.connect(self._reloadJobFinished)
+                job.start()
+    
+    def _reloadJobFinished(self, job):
+        # TODO: This needs to be fixed properly. We now make the assumption that we only load a single mesh!
+        mesh_data = job.getResult()[0].getMeshData()
+        if mesh_data:
+            job._node.setMeshData(mesh_data)
+        else:
+            Logger.log("w", "Could not find a mesh in reloaded node.")
--- a/UM/Mesh/MeshBuilder.py
+++ b/UM/Mesh/MeshBuilder.py
@@ -41,7 +41,7 @@ class MeshBuilder:
         # original center position
         self._center_position = None
         
-        Application.getInstance().windowStateChanged.connect(self._check_file_changed)
+        Application.getInstance().windowStateChanged.connect(self._checkFileChanged)
 
     ##  Build a MeshData object.
     #
@@ -768,7 +768,7 @@ class MeshBuilder:
 
         return True
     
-    def _check_file_changed(self, window_state):
+    def _checkFileChanged(self, window_state):
         if os.path.isfile(self._file_name) and window_state == Qt.WindowActive:
             if self._file_changed_stamp is not os.path.getmtime(self._file_name):
                  message = Message(i18n_catalog.i18nc("@info", "Would you like to reload {filename}?").format(filename = os.path.split(self._file_name)[0]),
--- a/UM/Mesh/MeshBuilder.py
+++ b/UM/Mesh/MeshBuilder.py
@@ -1,7 +1,7 @@
 # Copyright (c) 2015 Ultimaker B.V.
+# Copyright (c) 2017 Thomas Karl Pietrowski
 # Uranium is released under the terms of the LGPLv3 or higher.
 
-from UM.Application import Application
 from UM.Message import Message
 from UM.Mesh.MeshData import MeshData
 from UM.Mesh.MeshData import MeshType
@@ -12,7 +12,7 @@ from UM.Math.Vector import Vector
 from UM.Math.Matrix import Matrix
 from UM.Logger import Logger
 
-from PyQt5.QtCore import pyqtProperty, Qt, pyqtSignal, pyqtSlot, QUrl, QTimer
+from PyQt5.QtCore import QFileSystemWatcher
 
 import numpy
 import math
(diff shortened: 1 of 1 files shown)


UM/Mesh/MeshBuilder.py
```diff
@@ -34,6 +39,15 @@ def __init__(self):
         self._file_name = None
         # original center position
         self._center_position = None
+        
+        # File-changed notify timer
+        self._file_changed_timer = QTimer()
+        self._file_changed_timer.setInterval(1000)
+        self._file_changed_timer.setSingleShot(True)
+        self._file_changed_timer.timeout.connect(self._check_file_changed)
+        if self._file_name:
```
fieldOfView: Since self._file_name is set to None 9 lines above this, the timer never gets started. Right?

thopiekar: Nope, it shouldn't. I guess we need for the case, that someone sets up the filename after initializing this class, something like a .setFilename() function, which starts the timer after setting the path to self._file_name.

fieldOfView: My point was that you can *only* set up the filename after the initialization.

fieldOfView:
I don't really like timers for things like this (because they keep firing with little chance of a positive hit). Wouldn't a more meaningful event be more appropriate, such as the application regaining the focus after editing the mesh in another application?

UM/Mesh/MeshBuilder.py
thopiekar: Didn't take a look at the def `__init__` line 😮
Will provide a function then.

Btw. how is the overall impression about the PR? Is the rest good?

UM/Mesh/MeshBuilder.py
```diff
@@ -759,3 +767,31 @@ def addConvexPolygonExtrusion(self, xy_points, y0, y1, color=None):
         self.addQuad(v0, v1, v2, v3, color=color, normal = normal)
 
         return True
+    
+    def _check_file_changed(self, window_state):
```
nallath: Codestyle; needs to be _checkFileChanged

nallath:
Was this also tested on other operating systems? I can imagine that some OSes might be a bit different than others in this regard. 
Also; A single file might cause multiple meshes to be created (eg; 3MF files).

thopiekar:
I haven't tested it so far. Just like i said.. All coded in Github while having fingers crossed.

I'm not sure, but on 3MF files you are removing the file_name to prevent reloading. That's why I'm checking over here whether the variable is None or not. So for 3MF files this feature should be disabled.

UM/Mesh/MeshBuilder.py
thopiekar: Done in https://github.com/Ultimaker/Uranium/pull/307/commits/269fb932a9a5c667c19400aae202fb40f23b48b4

Ghostkeeper:
A cross-platform solution that doesn't require timers or additional libraries would be [QFileSystemWatcher](https://doc.qt.io/qt-5/qfilesystemwatcher.html). We have that already in our QtCore import. You simply add paths that you want to watch and register a callback to be called when those files change. I think that would be a better solution than rolling our own.

Also, it's probably easier and better to simply trigger the reload-all action (as if pressing F5) when the user clicks on the button of the message. It's not as efficient as targeted reloading of specific meshes, but it'll work when there are multiple models on the build plate.

Ghostkeeper:
Developers, see issue CURA-4637.

LipuFei:
If it's a project file, everything on your build plate will be gone.

thopiekar:
@LipuFei Can expect the same for 3MF, right?

Ghostkeeper:
I discussed this with @ChrisTerBeke and we would both prefer a solution using QFileSystemWatcher.
It's no longer checking every second, but QFileSystemWatcher would be a much more elegant solution to this.

thopiekar:
@Ghostkeeper Sure, QFileSystemWatcher knows for each OS what needs to be done.

thopiekar:
Made it now (hopefully) working with QFileSystemWatcher

Ghostkeeper:
The Jira ticket was closed because the misconception arose that this would send e-mails whenever something changed on disk. This is obviously not the case and I don't know where that came from. So I've cloned it and now it's on CURA-4912.

thopiekar:
Finally :yum:

UM/Mesh/MeshBuilder.py
```diff
@@ -1,7 +1,7 @@
 # Copyright (c) 2015 Ultimaker B.V.
+# Copyright (c) 2017 Thomas Karl Pietrowski
```
fieldOfView: Uhm, what?
Changing the year is a good idea, but why change the copyright holder?

ianpaschal: Whoops. I didn't even notice that. I brought up updating license stuff at a meeting once and we agreed to update it when we see it, but otherwise not worry about it.

This is where a contributor license agreement comes in handy...

ChrisTerBeke: Aren't credits are automatically given by Git? You can see exactly who did what and when. I think it's a bit overkill to add all this extra info to the core repository, a LOT of people worked on it in the end.

thopiekar: I heard other things when Ultimaker changed the license. The number of contributors was not that much as expected if I recall correctly.
Sure, but in my opinion, there are not many people who care about Git. But on the other hand, putting the credits on top of the file is also a bit too much. I have no idea how to get a good balance for this.

ChrisTerBeke: Over 100 contributors according to GitHub: https://github.com/Ultimaker/Cura/graphs/contributors. And contributors contribute under the existing license (if they wouldn't agree with it, they wouldn't contribute). Of course, a new file that's created by someone won't contain the Ultimaker copyright header, but given that Ultimaker employees created most of them, it's logical that most do.

thopiekar: After this got merged I couldn't remember that I ever added this line. After seeing it again, I slightly remember that I wanted to make a joke out of it. But I see how delicate this topic seems to be after adding on top my opinion here.

Therefore because I'm not interested to annoy people over here more than needed and wanted:
Please excuse my big mouth and remove this line. :pray:

UM/Mesh/MeshBuilder.py
```diff
@@ -759,3 +773,30 @@ def addConvexPolygonExtrusion(self, xy_points, y0, y1, color=None):
         self.addQuad(v0, v1, v2, v3, color=color, normal = normal)
 
         return True
+    
+    def _onFileChanged(self, file_path):
+        if os.path.isfile(self._file_name) and self._file_name is file_path:
+            message = Message(i18n_catalog.i18nc("@info", "Would you like to reload {filename}?").format(filename = os.path.split(self._file_name)[0]),
```
Ghostkeeper: `i18n_catalog` doesn't exist. This code was never run?

UM/Mesh/MeshBuilder.py
```diff
@@ -759,3 +773,30 @@ def addConvexPolygonExtrusion(self, xy_points, y0, y1, color=None):
...
+                              title = i18n_catalog.i18nc("@info:title", "File has been modified"))
+            message.addAction("reload",
+                              i18n_catalog.i18nc("@action:button", "Reload"),
+                              "[no_icon]",
+                              "[no_description]")
+            message.actionTriggered.connect(self._callback)
```
Ghostkeeper: `self._callback` doesn't exist.

UM/Mesh/MeshBuilder.py
thopiekar: Look at the PR description:
`Note: This has been coded without testing inside Github's webpage.`

There are at the moment some things I need to do. Will care about this PR asap.

Ghostkeeper:
I'm moving this code into UM.SceneNode rather than UM.MeshBuilder... The mesh builder gets deleted after the scene node is complete.

UM/Mesh/MeshBuilder.py
fieldOfView: In your defense, you're not the one who merged it without testing... But it is better practice to only make PRs with tested code.

Ghostkeeper: Well it's been merged now so I need to take care of it...

thopiekar: @Ghostkeeper: Alright, thanks!
@fieldOfView: My intention was to get things moving. So I began to make some code and share it as a PR. But you are right, next time I should practically test it.
Hope it will be much easier now since the PPA is working again.

fieldOfView: aww... :-/ (reverting is also an option)

thopiekar: Yep, and then I have the chance for a practical PR :grimacing:

ianpaschal: Blame where blame is due. It was me merging it without testing. To be honest though I'm always still a little bit fuzzy about this process with respect to Jira. After all... this issue was caught by merging it and then running tests on it, right?

I will discuss with the team tomorrow to clarify my understanding so we don't run into this again.

fieldOfView: You can merge the branch into a local branch and test it locally before merging into master.

thopiekar: @ianpaschal .. and it was me who share this ugly untested code.
Yep, what it definitely needs are not only builds and tests on master branches but also tests on PRs and all other branches. I suggested to use Travis for this in the past, but Ultimaker went for their own Jenkins solution.
Don't blame yourself on this. As I said I also have my part on this. Normally I'm not committing untested code (and make fun of copyrights), therefore let's be fair and say be both f**ked up. 😉

ianpaschal:
Both @LipuFei and I can't actually get this to work. @Ghostkeeper also says it wasn't entirely reliable. Can you look into this, @thopiekar, and make a new PR if it's working reliably?

I'd try to provide more info but it's not a crash or something. Files changed outside of Cura just... don't... do anything.

thopiekar:
No problem. As soon as I have enough time for this I will look try to find a reliable solution for this.
I'm quite sure you made all necessary changes and fixes to the code. So when I'm jumping into this again, then I will take a close look, whether the Qt module has all features enabled. Eg. checking whether QFileSystemWatcher needs to be built with inotify support or not. Just to make sure all boundaries are meet.

What are your plans for the feature now? Are you now going to revert all commits or simply removing all code for now?
Asking this, just before something gets lost. 😉

alekseisasin:
@thopiekar , so far if somebody working on it we will keep it open, for us it is flagged (*waiting you). Of course it would be great if you invest your time and complete it.

PR #311, opened 2017-11-28
CURA-4243 Feature local container server WIP

jackha:
From CURA-4243 issue:

Instead of shipping all printer profiles with each Cura release, I would like to make it possible for users to just download the printer profiles they really need.


UM/Settings/InstanceContainer.py
```diff
@@ -67,9 +69,8 @@ def __hash__(self):
         return id(self)
 
     def __deepcopy__(self, memo):
-        new_container = self.__class__(self._id)
-        new_container._name = self._name
-        new_container._definition = self._definition
+        new_container = self.__class__(self.getId())
+        new_container._name = self.getName()
```
jackha: I think this isn't used anymore?

UM/Settings/Interfaces.py
```diff
@@ -81,25 +85,39 @@ def hasProperty(self, key: str, property_name: str) -> bool:
     #   The serialized representation of the container can be used to write the
     #   container to disk or send it over the network.
     #
-    #   \param ignored_metadata_keys A list of keys that should be ignored when it serializes the metadata.
+    #   \param ignored_metadata_keys A set of keys that should be ignored when
+    #   it serializes the metadata.
     #
     #   \return \type{string} A string representation of this container.
-    def serialize(self, ignored_metadata_keys: Optional[List] = None) -> str:
+    def serialize(self, ignored_metadata_keys: Optional[set] = None) -> str:
```
jackha: The typing is not yet consistent in all places:

```
./UM/Settings/SettingDefinition.py:    def serialize(self) -> str:
./UM/Settings/Interfaces.py:    def serialize(self, ignored_metadata_keys: Optional[set] = None) -> str:
./UM/Settings/ContainerRegistry.py:    def serialize(self, ignored_metadata_keys: Optional[set] = None) -> str:
./UM/Settings/InstanceContainer.py:    def serialize(self, ignored_metadata_keys: Optional[set] = None) -> str:
./UM/Settings/DefinitionContainer.py:    def serialize(self, ignored_metadata_keys: Optional[set] = None):
./UM/Settings/ContainerStack.py:    def serialize(self, ignored_metadata_keys: Optional[set] = None):
./tests/Settings/ContainerTestPlugin/ContainerTestPlugin.py:    def serialize(self, ignored_metadata_keys=set()):
./tests/Settings/TestContainerRegistry.py:    def serialize(self, ignored_metadata_keys = None):
./tests/Settings/TestContainerStack.py:    def serialize(self, ignored_metadata_keys = None)
```

UM/Settings/Interfaces.py
```diff
@@ -110,14 +128,16 @@ def __updateSerialized(self, serialized: str) -> str:
 
     @classmethod
     def getLoadingPriority(cls) -> int:
-        return 9001
+        return 9001 #Goku wins!
```
jackha: https://vignette.wikia.nocookie.net/dbz-dokkanbattle/images/4/45/Weird_Goku_face_000.jpg/revision/latest?cb=20170511204602

plugins/LocalContainerProvider/LocalContainerProvider.py
```diff
@@ -0,0 +1,192 @@
...
+    def getAllIds(self) -> Iterable[str]:
+        if not self._id_to_path:
+            self._updatePathCache()
+        return self._id_to_path.keys()
+
+    def loadContainer(self, container_id: str) -> "ContainerInterface":
+        Logger.log("d", "Loading container {container_id}".format(container_id = container_id))
+        container_class = ContainerRegistry.mime_type_map[self._id_to_mime[container_id].name]
+        if issubclass(container_class, DefinitionContainer): #We may need to load these from the definition cache.
+            container = self._loadCachedDefinition(container_id)
```
jackha: Is it possible to try this first? After retrieving we can do the class test. Is the class test necessary?

UM/Settings/ContainerRegistry.py
```diff
@@ -105,153 +155,159 @@ def findContainerStacks(self, **kwargs) -> List[ContainerStack]:
...
 
         if query in self._query_cache:
-            # If the exact same query is in the cache, we can re-use the query result
-            self._query_cache.move_to_end(query) # Query was used, so make sure to update its position
+            #If the exact same query is in the cache, we can re-use the query result.
+            self._query_cache.move_to_end(query) #Query was used, so make sure to update its position so that it doesn't get pushed off as a rarely-used query.
             return self._query_cache[query].getResult()
 
-        # Execute the query, then add it to the cache
         query.execute()
```
jackha: If "id" is in kwargs, we could provide the candidate list (consisting of max 1 item). Then providing id and some other property will be almost equally fast as id only.

UM/Settings/Models/InstanceContainersModel.py
```diff
@@ -117,6 +117,8 @@ def _fetchInstanceContainers(self) -> Tuple[Dict[str, InstanceContainer], Dict[s
                         containers[metadata["id"]] = registry.findContainers(id = metadata["id"])[0]
                     else:
                         metadatas[metadata["id"]] = metadata
+        if not containers and not metadatas:
+            print("Filter of {class_name} returned nothing! Filters are: {filters}".format(class_name = self.__class__.__name__, filters = self._filter_dicts))
```
fieldOfView: print()?
Also, while I nitpick a work in progress, 'data' is already a plural form.

jackha: print() was an accident ^^

UM/Settings/Models/InstanceContainersModel.py
```diff
@@ -250,4 +252,4 @@ def _updateMetaData(self, container):
     def _onContainerLoadComplete(self, container_id):
         if container_id in self._instance_containers_metadata:
             del self._instance_containers_metadata[container_id]
-            self._instance_containers[container_id] = ContainerRegistry.getInstance().findContainers(id = container_id)
\ No newline at end of file
+            self._instance_containers[container_id] = ContainerRegistry.getInstance().findContainers(id = container_id)[0]
```
jackha: I see this [0] quite often. It may be an idea (for the future) to make a findContainer, which is special casing findContainers when you expect exactly 1 result

nallath:
There are conflicts.

UM/Settings/ContainerRegistry.py
```diff
@@ -289,6 +289,7 @@ def loadAllMetadata(self):
         for provider in self._providers: #Automatically sorted by the priority queue.
             for container_id in list(provider.getAllIds()): #Make copy of all IDs since it might change during iteration.
                 if container_id not in self.metadata:
+                    UM.Qt.QtApplication.QtApplication.processEvents()
```
nallath: Document why this needs to be done here.

jackha:
Reviewed till here

UM/Settings/ContainerRegistry.py
```diff
@@ -105,153 +155,159 @@ def findContainerStacks(self, **kwargs) -> List[ContainerStack]:
...
+    #   \return A list of metadata dictionaries matching the search criteria, or
+    #   an empty list if nothing was found.
+    def findContainersMetadata(self, *, ignore_case = False, **kwargs) -> List[Dict[str, Any]]:
+        #Create the query object.
+        query = ContainerQuery.ContainerQuery(self, ignore_case = ignore_case, **kwargs)
+
+        if "id" in kwargs:
+            if kwargs["id"] not in self.metadata: #If we're looking for an unknown ID, try to lazy-load that one.
+                for provider in self._providers:
+                    if kwargs["id"] in provider.getAllIds():
```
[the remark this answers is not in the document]
Ghostkeeper: I'd rather not add a metadata item for an implementation detail such as this. But we can add a mapping from container ID to their provider.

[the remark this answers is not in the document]
Ghostkeeper: Done: https://github.com/Ultimaker/Uranium/commit/e97d90dc998a3d0cb160a6c30bbe6f8c89d79b16

UM/Settings/DefinitionContainer.py
```diff
@@ -200,65 +199,60 @@ def hasProperty(self, key, property_name):
...
         data["metadata"] = self.getMetaData()
 
         # remove the keys that we want to ignore in the metadata
-        if ignored_metadata_keys:
-            for key in ignored_metadata_keys:
-                if key in data["metadata"]:
-                    del data["metadata"][key]
+        if not ignored_metadata_keys:
+            ignored_metadata_keys = set()
+        ignored_metadata_keys |= {"name", "version", "id", "container_type"}
```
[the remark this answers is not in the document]
Ghostkeeper: The whole idea of using a dictionary as storage for metadata is that it can be extensible from non-centralised places. As in, there is no pre-set set of keys. Otherwise we might as well just use a class or a namedtuple. But having a dictionary allows us to define additional metadata entries from anywhere, including plug-ins and such.

UM/Settings/InstanceContainer.py
Ghostkeeper: You're right. I removed it here: 894c9d581cf7ac6bc250408e53ca448cf91e04fa

UM/Settings/Interfaces.py
Ghostkeeper: I'll check them all:

* `SettingDefinition` is not in this class' family tree.
* `ContainerRegistry._EmptyInstanceContainer` is consistent.
* `InstanceContainer` is consistent.
* `DefinitionContainer` is consistent.
* `ContainerStack` is consistent.
* `ContainerTestPlugin` is consistent.
* `TestContainerRegistry` is consistent.
* `TestContainerStack` is consistent.

So it seems that no change is needed (or not any more).

plugins/LocalContainerProvider/LocalContainerProvider.py
Ghostkeeper: Getting the cache first would work, but it would cause us to check for existence of the cache file, which requires a system call and then hard disk access. That's much slower than checking a property of an instance, so I think it's best to check this first.

plugins/LocalContainerProvider/LocalContainerProvider.py
```diff
@@ -0,0 +1,192 @@
...
+
+        all_resources = set()
+        for resource_type in ContainerRegistry.getInstance().getResourceTypes():
+            all_resources |= set(Resources.getAllResourcesOfType(resource_type)) #Remove duplicates, since the Resources only finds resources by their directories.
+        for filename in all_resources:
+            if re.search(old_file_expression, filename):
+                continue #This is a back-up file from an old version.
+
+            try:
+                mime = MimeTypeDatabase.getMimeTypeForFile(filename)
```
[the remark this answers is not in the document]
Ghostkeeper: I took this structure over from old code.

It doesn't occur right now unless a plug-in registers itself as search location. This happens with the PostProcessing plug-inbut that plug-in doesn't store any profiles as of yet (actually it does but those definitions don't get added to the registry). For the future it's probably safer to let this be for now.

UM/Settings/Models/InstanceContainersModel.py
Ghostkeeper: We have it all over our code since these find functions exist. To me it's just a warning that you might get an empty list so you must be sure that the filter returns something. For instance in this case, it is first checked that `container_id` is in the model and it is guaranteed that anything in the model already exists.

UM/Settings/ContainerRegistry.py
```diff
@@ -203,11 +203,14 @@ def findContainersMetadata(self, *, ignore_case = False, **kwargs) -> List[Dict[
...
-            if query.isIdOnly(): #If we are just searching for a single container by ID, look it up from the ID-based cache.
-                if kwargs["id"] in self.metadata:
-                    return [self.metadata[kwargs["id"]]]
-                else:
-                    return [] #No result, so return an empty list.
+            #Since IDs are the primary key and unique we can now simply request the candidate and check if it matches all requirements.
+            if kwargs["id"] not in self.metadata:
+                return [] #No result, so return an empty list.
+            candidate = self.metadata[kwargs["id"]]
+            if candidate.items() >= kwargs.items():
```
jackha: This doesn't work with wildcards.

jackha: I'll fix it

PR #312, opened 2017-11-28
add traditional chinese translation (zh_TW)

dinowchang:


resources/i18n/zh_TW/uranium.po
```diff
@@ -0,0 +1,868 @@
...
+#: /home/ruben/Projects/Uranium/UM/Qt/QtApplication.py:131
+msgctxt "@info:progress"
+msgid "Loading preferences..."
+msgstr "正在載入偏好設定..."
+
+#: /home/ruben/Projects/Uranium/UM/Mesh/ReadMeshJob.py:67
+#, python-format, python-brace-format
+msgctxt "@info:status"
+msgid "Auto scaled object to {0}% of original size"
+msgstr "自動將物件縮放到原始大小的 {0}% o"
```
Ghostkeeper: Seems like there's a little mistake here?

Ghostkeeper:
I'm merging these into master and then cherry-pick them into 3.1 to get them along with this release.

resources/i18n/zh_TW/uranium.po
dinowchang: The 'o' was put at the end of string intentionally. I use Poedit 2.0.4 to convert po file to mo file. If I removed the 'o', poedit pop up a error message "msgstr is not a valid Python format string."

Ghostkeeper: PoEdit might be mistaken there then. It is indeed not a valid Python format string, but it's not actually formatted in that way so it's a bit of a weird constraint to put on your texts... Oh well, next time maybe.

PR #313, opened 2017-12-02
Adding a factory for Qt

thopiekar:
This PR adds an layer between PyQt5 and Uranium plus Cura. The idea of this layer is to commonize PyQt5 and PySide2. As you can see from UM.Qt.Factory, both bindings are the same except of QtCore.
The reason for this PR are the issues with my PPA (https://github.com/thopiekar/Cura-packaging/issues/30) and since I already own a PPA for PySide2, but haven't updated it for a while, I decided to give it a try.
However, I couldn't get it working so far because of missing functions on old PySide2 builds, but I hope I can get something working soon.

Fixes: https://github.com/thopiekar/Cura-packaging/issues/30


fieldOfView:
A description would be nice... This is to make Uranium compatible with PySide2 as well as PyQt5? Neat. I didn't know PySide was still alive.

thopiekar:
Sorry for the missing description I was busy today and additionally there are some missing functions inside PySide2.
There are still missing functions are around QtGui.QGl*.
Will contact some people to ask whether they can make it happen to provide these functions, so we can use it with Cura and Uranium.
Yes, PySide2 is still alive. Officially supported by the Qt Company.

thopiekar:
But yeah, apart from that it works.

thopiekar:
They changed something to their build system.. Therefore my packages in my PPA are old.
Maybe I can get something working here..

thopiekar:
Ok, Gl things are there, but still missing things in QML:
UM.Qt.Factory.Utils.<module> [33]: Using Qt binding: PySide2
cura.CrashHandler.__init__ [62]: An uncaught exception has occurred!
cura.CrashHandler.__init__ [65]: Traceback (most recent call last):
cura.CrashHandler.__init__ [65]:   File "/foo/bar/Cura/cura_app.py", line 58, in <module>
cura.CrashHandler.__init__ [65]:     import cura.CuraApplication
cura.CrashHandler.__init__ [65]:   File "/foo/bar/Cura/cura/CuraApplication.py", line 7, in <module>
cura.CrashHandler.__init__ [65]:     from UM.Qt.QtApplication import QtApplication
cura.CrashHandler.__init__ [65]:   File "/foo/bar/Uranium/UM/Qt/QtApplication.py", line 19, in <module>
cura.CrashHandler.__init__ [65]:     from UM.Qt.Bindings.Bindings import Bindings
cura.CrashHandler.__init__ [65]:   File "/foo/bar/Uranium/UM/Qt/Bindings/Bindings.py", line 4, in <module>
cura.CrashHandler.__init__ [65]:     from UM.Qt.Factory.QtQml import qmlRegisterType, qmlRegisterSingletonType, qmlRegisterUncreatableType
cura.CrashHandler.__init__ [65]: ImportError: cannot import name 'qmlRegisterSingletonType'

thopiekar:
https://bugreports.qt.io/browse/PYSIDE-574?filter=-2

ChrisTerBeke:
Cool stuff man. We'll be watching this. Once it's stable we can push a ticket in Jira if we feel it's something we want.

thopiekar:
Thanks! Wanted to have this implemented for something big last year.
But won't tell what I have in mind :stuck_out_tongue:

nallath:
PySide2 is actually getting quite a bit of love from Qt.

awhiemstra:
What is the advantage of doing this ourselves over using https://github.com/spyder-ide/qtpy ?

ChrisTerBeke:
That lib seems fairly well maintained at the moment. Not sure if that'll keep happening ofc. We could fork it and use that instead maybe?

thopiekar:
@awhiemstra One dependency less and, hey, my factory is awesome 😎 
But currently it really depends on all the missing functions in PySide2 in my opinion.

@ChrisTerBeke Seems so. I didn't know of this abstraction layer before.

I think it is better to use qtpy then.. Less maintenance for us..

daid:
@thopiekar You should be working on finishing up your paper ;-)

PR #314, opened 2017-12-04
updated brazilian portuguese translations

Patola:


Ghostkeeper:
Looks good. Thanks for the update!

PR #316, opened 2017-12-08
(DONE) Registration of arguments of plugins, add invisible mode (No splashscreen!) and "--debug" mode as argument

thopiekar:
Needed to make Cura run as COM server.
_Features:_
* Hidden mode (no splash) ("Cura.exe --invisible")
* Adding possibility for plugins to add command line arguments
* Keep argument parser alive and throw an error on unknown arguments at the end of initialization

_Needed for:_
 * Individual arguments in the InterfacePlugin (https://github.com/thopiekar/CuraInterfacePlugin)

### DIFF
--- a/UM/Application.py
+++ b/UM/Application.py
@@ -337,7 +337,7 @@ class Application:
     def getCommandlineParser(self, with_help = False):
         if not self._command_line_parser:
             self._command_line_parser = argparse.ArgumentParser(prog = self.getApplicationName(), add_help = with_help) #pylint: disable=bad-whitespace
-            self.addCommandLineOptions(self._command_line_parser)
+            self.addCommandLineOptions(self._command_line_parser, parsed_command_line = self._parsed_command_line)
         if self._command_line_parser.add_help is not with_help:
             self._command_line_parser.add_help = with_help
         return self._command_line_parser
@@ -350,7 +350,7 @@ class Application:
     #
     #   \param parser \type{argparse.ArgumentParser} The parser that will parse the command line.
     @classmethod
-    def addCommandLineOptions(cls, parser):
+    def addCommandLineOptions(cls, parser, parsed_command_line = {}):
         parser.add_argument("--version",
                             action="version",
                             version="%(prog)s {0}".format(cls.getStaticVersion()))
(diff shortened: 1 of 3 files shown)


ChrisTerBeke:
Is this a WIP? I see commit's getting added.

thopiekar:
@ChrisTerBeke Done

nallath:
Cura already had a --headless, so why not append onto that?

ChrisTerBeke:
I'm gonna merge this because it comes in handy for my tickets, but I'll make sure to only use --headless and remove --invisible.

thopiekar:
@ChrisTerBeke : Sorry, that I haven't had time to care about the option names.
Is it still possible to give a veto on the option name?
I mean I wouldn't call it headless because we still initialize OpenGL and other things, I think. In my imagination it is not really "headless". If I hear headless, then I think of an imagine of an application, which is really minimalistic. In our case an application which maximally uses QtCore to create events, etc., but nothing more.

@nallath : To answer your question: I haven't had an idea, what your (Ultimakers) plans were with "--headless". Thus I thought if I need something completely invisible, I just introduce a fully invisible runtime with "--invisible".

And I personally find "--invisible" more fitting to the purpose of the commandline option. I mean this is what it actually makes - it only hides the windows, but does not prevent to load any graphical libraries, eg. QtGui and friends.
If you agree about this, too, then I can care about renaming it in both projects, because it is mainly my interest to call this function appropriately.

ChrisTerBeke:
I'm working right now on making at as headless as possible since that's needed to run in a Docker environment, so for now I'm calling it headless.

thopiekar:
Ok, then I'm fine with that 👍 
Is it also your task to have the possibility to make it visible again, if needed?

Eg. someone starts it invisible (sorry I mean headless), but there is a user action required. Then I would like to make it visible temporary. Eg. asking whether the user really wants to override her/his custom settings.

- Regards!

ChrisTerBeke:
Nope, that's not in the scope. I'm focused on running Cura 3.x in a Docker environment, where it'll never stop running as headless unless it crashes or the cluster is scaled down/upgraded. So 0 user interaction in that workflow.

thopiekar:
Is there a branch(es) you are working on? So I can take a diff before it is done?
Personally I'm interested in this because of my work here.

ChrisTerBeke:
feature_headless_docker. Right now it's the same as master (after merging this).

PR #318, opened 2017-12-08
QtApplication: Prevent using dir

thopiekar:
It is a built-in function

(Managed to push into the wrong branch .. Same as #317)


ChrisTerBeke:
Always a +1 from me when not overriding built-in parameters.

Issue #319, opened 2017-12-10
3.1.0: Failing test suite


hroncok:
This is Fedora rawhide (to be 28), Python 3.6.3. Seems architecture agnostic (reproduced so far on aarch64 and i686).

```console
$ python3 -m pip freeze
numpy==1.13.3
py==1.5.2
pytest==3.2.3
scipy==1.0.0
```

The following tests fail:

 * `tests/Settings/TestContainerRegistry.py`
   * `test_addContainerType`
   * `test_uniqueName`
 * `tests/Settings/TestContainerStack.py``
   * `test_addContainer`
   * `test_findContainer[*]`
   * `test_getContainer`
   * `test_getValue[data1]`
   * `test_getValue[data2]`
   * `test_getValue[data3]`
   * `test_removeContainer`
   * `test_replaceContainer`
   * `test_setNextStack`
   * `test_backwardCompatibility`
   * `test_idSpecialCharacters`

All seems to be (either directly or indirectly, exactly or similar to):

```
E               NotImplementedError: Method <function MockContainer.deserialize at 0xffff88c90400> of class <class 'TestContainerStack.MockContainer'> does not have the same signature as method <function ContainerInterface.deserialize at 0xffff8998ca60> in interface <class 'UM.Settings.Interfaces.ContainerInterface'>: (self, serialized) vs (self, serialized:str, file_name:Union[str, NoneType]=None) -> str
UM/Decorators.py:92: NotImplementedError
```

<details>
<summary>Full tests log</summary>

```
+ /usr/bin/python3 -m pytest -v
============================= test session starts ==============================
platform linux -- Python 3.6.3, pytest-3.2.3, py-1.5.2, pluggy-0.4.0 -- /usr/bin/python3
cachedir: .cache
rootdir: /builddir/build/BUILD/Uranium-3.1.0, inifile: pytest.ini
collecting ... collected 283 items
tests/TestDecorators.py::test_interface PASSED
tests/TestResources.py::TestResources::test_getCacheStorageRootPath_Linux PASSED
tests/TestResources.py::TestResources::test_getCacheStorageRootPath_Mac SKIPPED
tests/TestResources.py::TestResources::test_getCacheStorageRootPath_Windows SKIPPED
tests/TestResources.py::TestResources::test_getConfigStorageRootPath_Linux PASSED
tests/TestResources.py::TestResources::test_getConfigStorageRootPath_Mac SKIPPED
[...]

PR #320, opened 2017-12-10
Fix failing test suite after 435285e

hroncok:
Fixes https://github.com/Ultimaker/Uranium/issues/319

### DIFF
--- a/tests/Settings/ContainerTestPlugin/ContainerTestPlugin.py
+++ b/tests/Settings/ContainerTestPlugin/ContainerTestPlugin.py
@@ -83,7 +83,7 @@ class ContainerTestPlugin(ContainerInterface, PluginObject):
     ##  Deserializes the container from a string representation.
     #
     #   This method is not implemented in the mock container.
-    def deserialize(self, serialized):
+    def deserialize(self, serialized, file_name=None):
         raise NotImplementedError()
 
     def getConfigurationTypeFromSerialized(self, serialized):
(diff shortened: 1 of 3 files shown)


hroncok:
As a Fedora packager, I have noticed that this is not the first time either Cura or Uranium tests suite is broken after the implementation was updated but the tests were not. Having test without constantly running them is worst than not having them at all (it gives you false idea of safety while there is none).


Could you please consider using some kind of mechanism to prevent this in the future? I.e. some kind of Continues Integration or even Continues Integration with Pull Requests? I consider [Travis CI](https://travis-ci.org/) a perfect service for this kind of stuff, it's also free (both as in free beer and freedom). Thanks.

hroncok:
I see you are using [Jenkins somehow](http://hal9000.ultimaker.com:29130/job/Ultimaker/job/Uranium/job/master/), but it's incredibly hard for me to see whether this was happening or not and when it started.

ChrisTerBeke:
@alekseisasin weren't you working on fixing unit tests?

Ghostkeeper:
Alekseisasin and I fixed the tests because a lot of them also got broken due to this architecture rework: https://github.com/Ultimaker/Uranium/pull/311

We also fixed the tests that got broken from the added file_name parameter and the single extrusion extruder stacks. So this is PR now unnecessary and is giving conflicts because we modified the tests for the same purpose.

PR #321, opened 2017-12-11
CURA-4234 Make sidebar collapsible

jackha:
CURA-4234

Ctrl-E or using the View pull down menu

### DIFF
--- a/UM/Qt/Bindings/MainWindow.py
+++ b/UM/Qt/Bindings/MainWindow.py
@@ -52,6 +52,7 @@ class MainWindow(QQuickWindow):
         self._preferences.addPreference("general/window_top", 50)
         self._preferences.addPreference("general/window_state", Qt.WindowNoState)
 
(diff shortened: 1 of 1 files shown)


UM/Qt/Bindings/MainWindow.py
```diff
@@ -52,6 +52,7 @@ def __init__(self, parent = None):
         self._preferences.addPreference("general/window_top", 50)
         self._preferences.addPreference("general/window_state", Qt.WindowNoState)
 
+        self._preferences.addPreference("general/sidebar_collaps", False)
```
jackha: - The preference should be in Cura, not Uranium
- The correct spelling is "sidebar_collapse"

nallath: It also has nothing to do with the main window. It should be in CuraApplication.

fieldOfView:
This PR has no changes now, so it can be closed

PR #322, opened 2018-01-17
Fix a crash when a definition for a setting is not available

fieldOfView:
This PR fixes a crash when switching to a machine for which a setting is not defined which is defined for the current machine. This can happen if a machine defines settings in its def.json file.

I don't remember this being an issue in 3.0, so that would make the crash a regression.


nallath:
CURA-4825

LipuFei:
@fieldOfView Could be the mandatory ExtruderStack we added. The most likely one I think is `material_diameter`. It is a per-extruder setting, but it's only defined in the machine definition, not the extruder definition. Then if an ExtruderStack tries to look for that setting definition in its own definition container, it will break.

PR #323, opened 2018-01-25
Extend seven seas

LoyVanBeek:


scripts/pirateofdoom.py
```diff
@@ -43,9 +44,25 @@
...
     "connect to": "board",
+    "connecting": "boarding",
     "collects": "hoards",
     "prime tower": "buoy",
     "change log": "captain's log",
     "my": "me",
-    "removable drive": "life boat"
+    "removable drive": "life boat",
+    "print core": "scribbler",
+    "printcore": "scribbler",
```
Ghostkeeper: These are not going to be translated. Since they are specific to Ultimaker 3 they are in the UM3 data file and not translated. I'll leave them in for now because it can't hurt.

scripts/pirate.py
```diff
@@ -16,6 +17,10 @@ def translate(english):
         for position in matches:
             #Make sure the case is correct.
             uppercase = english[position].lower() != english[position]
+
+            if isinstance(pir, list):
+                pir = random.choice(pir)
```
Ghostkeeper: This will not choose a different translation every time you load Cura, but every time this script is ran when we create translations. So for every release basically.

scripts/pirateofdoom.py
```diff
@@ -43,9 +44,25 @@
...
+    "start": "Sail ho!",
+    "abort": ["maroon", "abandon"],
+    "aborting": ["marooning", "abandoning"],
+    "aborted": ["marooned", "abandoned"],
+    "connected": ["anchored", "moored"],
+    "developer": "scurvy dog",
+    "wizard": "cap'n",
```
Ghostkeeper: So the "bed levelling wizard" would become "bed levelling cap'n"?

scripts/pirateofdoom.py
```diff
@@ -43,9 +44,25 @@
...
+    "active leveling": "keelhauling",
+    "download": "plunder",
+    "downloaded": "plundered",
+    "caution hot surface": "fire in the hole!",
+    "type": "sort",
```
Ghostkeeper: I like this one. Subtle!

scripts/pirateofdoom.py
```diff
@@ -43,9 +44,25 @@
...
+    "connecting": "boarding",
     "collects": "hoards",
     "prime tower": "buoy",
     "change log": "captain's log",
     "my": "me",
-    "removable drive": "life boat"
+    "removable drive": "life boat",
+    "print core": "scribbler",
+    "printcore": "scribbler",
+    "start": "Sail ho!",
```
Ghostkeeper: This one is a bit of a problem. We often have stuff like "unable to start a print job" in our translations. That would become "unable to sail ho! a print job". We never say "start" separately.

PR #324, opened 2018-01-29
Plugin refactor

ianpaschal:
I've consolidated a lot of functionality within the plugin registry and as far as I can tell, everything is still working, but this is why I wanted to talk with Terri about being better able to know if something is going to fail tests.

### DIFF
--- a/UM/PluginObject.py
+++ b/UM/PluginObject.py
@@ -7,6 +7,7 @@
 #   This class should be inherited by any class that can be provided
 #   by a plugin. Its only function is to serve as a mapping between
 #   the plugin and the object.
+
 class PluginObject:
     def __init__(self):
         self._plugin_id = None
--- a/UM/PluginRegistry.py
+++ b/UM/PluginRegistry.py
@@ -38,13 +38,14 @@ class PluginRegistry(QObject):
 
     def __init__(self, parent = None):
         super().__init__(parent)
-        self._plugins = {}  # type: Dict[str, types.ModuleType]
-        self._plugin_objects = {}  # type: Dict[str, PluginObject]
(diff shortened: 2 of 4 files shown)


fieldOfView:
Something I did not catch before is that you this changed the the Plugins page in Uranium. The page now refers to the Plugin Browser, but the plugin browser only exists in Cura. There are other applications that use Uranium that do not have a plugin browser, and are now not able to manage their plugins (among other applications this includes @nallath and @Ghostkeeper's scanner application and @daid's laser cutter application).

Instead of changing the plugins page in Uranium, you should have removed the Uranium plugins page from the Cura preferences and then add a Cura-specific plugins page in its place, similar to how Cura has a specific General settings page, see https://github.com/Ultimaker/Cura/blob/master/resources/qml/Cura.qml#L508

Remember, the only reason I know this is that I also did this sort of thing wrong, and I was told by the aforementioned people that I broke their stuff.

ianpaschal:
Do you mean the page in the preferences?

daid:
FYI My lasercutter application hasn't been compatible with newer uranium for quite a long time already. Never took the time to update it.

fieldOfView:
Yes, I am talking about PluginsPage.qml

ianpaschal:
Yes, we are aware of that. I made a 5 minute fix and have been working on a more future-proof fix. Will commit when it's of solid quality. I'd also like to refactor the PluginRegistry to actually have the optional ability to fetch a remote plugin list and check for upgrades so that there's neither the creation nor modification of PluginsModels in the PluginBrowser (instead it just displays the given application's single PluginRegistry filtered in different ways).

PR #326, opened 2018-02-04
Update zh_TW translation for 3.2

dinowchang:


Ghostkeeper:
I didn't even see this. I'm sorry!

I'll merge this now but the translation has already been disabled for 3.2, so it'll be for 3.3 again... I'll try to pay better attention to incoming pull requests when the next translation cycle is happening.

PR #330, opened 2018-02-20
CURA-4790 If using "bold", use .setBold(), else use .setWeight()

ianpaschal:
Based on feedback from @fieldOfView.

### DIFF
--- a/UM/Qt/Bindings/Theme.py
+++ b/UM/Qt/Bindings/Theme.py
@@ -198,8 +198,11 @@ class Theme(QObject):
                 f = QFont()
                 f.setFamily(font.get("family", QCoreApplication.instance().font().family()))
 
-                f.setWeight(font.get("weight", 50))
-                # f.setBold(font.get("bold", False))
+                if font.get("bold"):
+                    f.setBold(font.get("bold", False))
+                else:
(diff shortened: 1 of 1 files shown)


fieldOfView:
I would do it the other way around though; have the (more specific) weight override bold, if they are both specified.

ianpaschal:
I went back and forth but decided to make weight the default as every font has "some weight," but not every font has "bold." To me, the default should be some value, not the lack of an extra property. "Bold" is the override to "normal" similar to "italic."

I don't think there's a technical issue to either one, it's just preference.

fieldOfView:
Most - if not all - of the fonts I have installed have a bold, and only the expensive ones have different weights. But it is not critical, as long as either of them works.

ianpaschal:
Actually every font has weight(s). "Bold" is merely a shorthand for "700". The expensive ones just have more weights than 400 and 700. The fact that you can set the weight and bold properties is more a product of not many programmers being typographers. While we're at it, "font family" should really be referred to as "typeface". But whatever, haha.

But yeah, since I think it's personal preference, I'm going to consider this reviewed and merge. If we find a reason in the future to switch it back... well... we'll code that bridge when get to it.

PR #333, opened 2018-03-15
Revert setting visibility preset

fieldOfView:
This PR removes some vestigial code now that https://github.com/Ultimaker/Cura/pull/3274 has moved the setting visibility code into Cura.


Ghostkeeper:
There's a test tool in Cura that still tests if the directory `preset_setting_visibility_groups` exists. I'll remove that tool.

PR #337, opened 2018-04-12
Cura 4644 package reader

ianpaschal:
Not yet ready for merge, but ready to start reviewing!

Please test in conjunction with cura/feature_material_marketplace

### DIFF
--- a/UM/Settings/InstanceContainer.py
+++ b/UM/Settings/InstanceContainer.py
@@ -40,7 +40,7 @@ MimeTypeDatabase.addMimeType(
 #
 @signalemitter
 class InstanceContainer(QObject, ContainerInterface, PluginObject):
-    Version = 3
+    Version = 4
 
     ##  Constructor
     #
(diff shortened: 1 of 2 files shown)


UM/Settings/InstanceContainer.py
```diff
@@ -40,7 +40,7 @@ class DefinitionNotFoundError(Exception):
 #
 @signalemitter
 class InstanceContainer(QObject, ContainerInterface, PluginObject):
-    Version = 3
+    Version = 4
```
Ghostkeeper: Since nothing changed in the storage of these instance containers, I don't see any need to change their version numbers.

LipuFei: Replied in https://github.com/Ultimaker/Cura/pull/3663#discussion_r185174128

To trigger a version upgrade so we can move and rename the existing custom profiles, materials, etc. to have their own namespaces (dirs).

PR #346, opened 2018-05-07
CURA-5164 Make application initialization and start up more clear

jackha:
- Instead of calling getInstance() everywhere, initialize each object
   explicitly in order when application starts and getInstance()s do not
   create instances any more and they merely return the created
   instances.
 - Only set initial values in construtor functions __init__(). Move the
   initialization of context-aware (i.e. things that depend on other
   things) to separate functions.
 - Split application creation and initialziation into several steps and
   them should be called explicitly in the correct order.


ChrisTerBeke:
@LipuFei will this be cherry picked into 3.4? The deprecation warnings now reference Cura 3.3 and I have updated my plugin to use the new preferred methods, but I saw that this is only the master branch, not 3.4.

ChrisTerBeke:
MM, maybe update the deprecation warning (to 3.5) then?

LipuFei:
Added `getPreferences()` in 3.4 942a99f4

LipuFei:
@ChrisTerBeke Don't upload your plugins with that change now because 3.4 beta doesn't have this.

diegopradogesto:
After this we need also to increase the plugin API version since the read method of the MeshReader changed its signature. This changed in ca0b642 and 494185f.

ChrisTerBeke:
@diegopradogesto that'll be for Cura 4.0 right?
@LipuFei I have released a patch of my plugin that supports both options now.

diegopradogesto:
Yes, mandatory (IMO)

PR #349, opened 2018-05-15
Fix a crash when starting Cura.

### LINKED ISSUE #347: Uranium only tries to find its plugins into $prefix/lib/
Title almost tells everything. When starting Cura, it crashes because:
[MainThread] cura.CrashHandler.__init__ [68]: UM.PluginError.PluginNotFoundError: Could not find plugin ConsoleLogger

UM/Application.py only looks in lib/
self._plugin_registry.addPluginLocation(os.path.join(Application.getInstallPrefix(), "lib", "uranium"))

The plugins can be installed anywhere (lib64 in my case)

krop:
If LIB_SUFFIX is used when invoking CMake, we have to let Cura know where the
plugins can be found.

Fixes #347


krop:
Your 2. is a matter of taste (we use foo.cmake for KDE projects to indicate it's a file CMake has to parse), but ok.
About 1, I wonder how Cura may be built then executed from source, even without this patch.
Just tried locally and running cura_app.py complains immediately about the lack of UM.Platform.

I'll add a fallback then prepare the next PR to make cura packager-friendly (ie: use CMake as it should be)

thopiekar:
Num 2 is based on how CuraVersion.py ist generated in Cura. Debian packagers are also adding this suffix often.
Also adding .py as prefix of the suffix makes it easier to understand that Python syntax is expected inside this file.

Num 1 is also implemented around CuraVersion.py. Since these projects are written in Python, most people like to skip the building part. So when someone is developing, then he can set the different PYTHONPATHs (incl build arcus and the engine) and she/he can start testing her/his changes directly.

Ghostkeeper:
Did you get around to making a new PR?

PR #351, opened 2018-05-18
CURA-5357 drag and drop packages

ianpaschal:
This hack is needed in order to not display an error when dragging and dropping Cura packages.

This hack will no longer be needed when we build a proper package reader with libCharon and other shiny features.

### DIFF
--- a/UM/FileHandler/ReadFileJob.py
+++ b/UM/FileHandler/ReadFileJob.py
@@ -7,6 +7,7 @@ from UM.Logger import Logger
 from UM.Mesh.MeshReader import MeshReader
 
 import time
+import os
 
 from UM.i18n import i18nCatalog
 i18n_catalog = i18nCatalog("uranium")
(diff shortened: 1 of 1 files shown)


UM/FileHandler/ReadFileJob.py
```diff
@@ -24,10 +25,17 @@ def getFileName(self):
         return self._filename
 
     def run(self):
+        # HACK: Packages shouldn't be checked for a reader. Maybe someday we build a reader. Maybe not. But for now,
+        # this hack is the only solution.
+        filename, file_extension = os.path.splitext(self._filename)
+        if file_extension == ".curapackage":
```
LipuFei: This used to be a hack in `Cura.qml`, check this commit for example:
https://github.com/Ultimaker/Cura/commit/0f966115e65762149d61009b4968598a5bb33c20#diff-4b601941ef414d7ed8e6b673a2d46752

LipuFei: OK, that way doesn't work in this case..

LipuFei:
Use a simpler approach instead: https://github.com/Ultimaker/Cura/pull/3845

PR #353, opened 2018-06-01
Add bundled packages to package manager

Ghostkeeper:
This adds the bundled packages of Uranium to the package manager, so that they can be upgraded by adding a new package.

It depends on https://github.com/Ultimaker/Uranium/pull/352 and https://github.com/Ultimaker/Cura/pull/3850, so please merge those two first before merging this PR.


UM/Qt/QtApplication.py
```diff
@@ -551,6 +549,10 @@ def getMeshFileHandler(self) -> MeshFileHandler:
     def getWorkspaceFileHandler(self) -> WorkspaceFileHandler:
         return self._workspace_file_handler
 
+    @pyqtSlot(result = QObject)
+    def getPackageManager(self, *args):
+        return self._package_manager
+
```
ianpaschal: In relation to the comment above, I don't think the package manager (or in this case `self._package_manager` not being `None`) should be a required part of `QtApplication.py`.

PR #354, opened 2018-06-04
CURA-5440 fix 3.4 upgrade

LipuFei:


LipuFei:
Together with https://github.com/Ultimaker/Cura/pull/3883

PR #356, opened 2018-06-28
Ignore cura.log when checking for files to be updated

fieldOfView:
This PR stops VersionUpgradeManager from opening and reading Cura.log (multiple times!) to see whether it needs updating. The application log can not be upgraded, so this is a waste of memory and cpu cycles.

Ignoring a 220 MB accumulated Cura.log file bumped the boot time down from more than 22 seconds to less than 10 on my system.


fieldOfView:
The PR now also ignores logs and old backups when copying an old configuration folder (if a configuration folder for the current version is not present).

Appesteijn:
CURA-5530

PR #359, opened 2018-07-06
Add typing

ianpaschal:


diegopradogesto:
Fix merge conflict.

ianpaschal:
Oh hmmm. You're right. In this case leave it because using `Optional[int]` doesn't seem to trigger the code style tests anymore.

PR #361, opened 2018-07-09
CURA-5389 Check and confirmation before Remove package 

jackha:
Added functions to make it able to explore package contents for above functionality in Cura.

### DIFF
--- a/UM/PackageManager.py
+++ b/UM/PackageManager.py
@@ -7,12 +7,14 @@ import os
 import shutil
 import zipfile
 import tempfile
+import urllib.parse  # For interpreting escape characters using unquote_plus.
 
 from PyQt5.QtCore import pyqtSlot, QObject, pyqtSignal, QUrl
 
 from UM import i18nCatalog
 from UM.Logger import Logger
 from UM.Message import Message
+from UM.MimeTypeDatabase import MimeTypeDatabase, MimeType  # To get the type of container we're loading.
 from UM.Resources import Resources
 from UM.Version import Version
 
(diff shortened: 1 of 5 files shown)


UM/PackageManager.py
```diff
@@ -402,3 +404,51 @@ def getPackageLicense(self, filename: str) -> Optional[str]:
                                             file_info.filename)
                         license_string = None
         return license_string
+
+    ##  Find the package files by package_id by looking at the installed folder
+    @staticmethod
+    def getPackageFiles(package_id) -> List[Tuple[str, List[str]]]:
+        data_storage_dir = os.path.abspath(Resources.getDataStoragePath())
```
ianpaschal: Suspect issues if `Resources.getDataStoragePath()` returns `None`

UM/PackageManager.py
```diff
@@ -402,3 +404,51 @@ def getPackageLicense(self, filename: str) -> Optional[str]:
...
+
+        os_walk = []
+        dirs_to_check = []
+        result = []  # 2-tuples of (dir, file_names)
+        for root_path, dir_names, file_names in os.walk(data_storage_dir):
```
ianpaschal: Can be errors if `data_storage_dir` is `None`

UM/PackageManager.py
LipuFei: `Resources.getDataStoragePath()` always returns a `str`, so that's fine.

UM/PackageManager.py
jackha: data_storage_dir is the result of os.path.abspath and that's always a string.
os.path.abspath's input is Resources.getDataStoragePath() and that also always returns a string.

PR #363, opened 2018-07-11
Fix unit tests & typing

LipuFei:


LipuFei:
@ianpaschal

ianpaschal:
Pfff I'm afraid to merge this. Don't want to run into the same problem as in Cura where the answer to "What should be tested for this?" was "Everything." and not everything was tested and some things broke.

plugins/LocalContainerProvider/LocalContainerProvider.py
```diff
@@ -105,26 +119,26 @@ def saveContainer(self, container: "ContainerInterface") -> None:
     #   \param container_id The ID of the container to load the metadata of.
     #   \return The metadata of the specified container, or ``None`` if the
     #   metadata failed to load.
-    def loadMetadata(self, container_id: str) -> Optional[Dict[str, Any]]:
+    def loadMetadata(self, container_id: str) -> Dict[str, Any]:
```
ianpaschal: Aww yisss. Fewer optionals.

plugins/LocalContainerProvider/LocalContainerProvider.py
```diff
@@ -105,26 +119,26 @@ def saveContainer(self, container: "ContainerInterface") -> None:
...
         registry = ContainerRegistry.getInstance()
         if container_id in registry.metadata:
             return registry.metadata[container_id]
 
         filename = self._id_to_path[container_id] #Raises KeyError if container ID does not exist in the (cache of the) files!
         clazz = ContainerRegistry.mime_type_map[self._id_to_mime[container_id].name]
 
-        requested_metadata = None
+        requested_metadata = {}  # type: Dict[str, Any]
```
ianpaschal: Don't think we need a typing here, right? Since the typing is already defined for the return value. Also because in two cases below we return an empty dict without typing.

plugins/LocalContainerProvider/LocalContainerProvider.py
```diff
@@ -288,7 +302,9 @@ def _pathToMime(self, path) -> Optional[MimeType]:
...
 
     ##  Converts a file path to the ID of the container it represents.
-    def _pathToId(self, path) -> Optional[str]:
+    def _pathToId(self, path: str) -> Optional[str]:
+        result = None
         mime = self._pathToMime(path)
         if mime:
-            return urllib.parse.unquote_plus(self._pathToMime(path).stripExtension(os.path.basename(path)))
+            result = urllib.parse.unquote_plus(mime.stripExtension(os.path.basename(path)))
+        return result
```
ianpaschal: Can be simplified to:

```python
if mime:
    return urllib.parse.unquote_plus(mime.stripExtension(os.path.basename(path)))
return None
```

Nice to make fewer variables right? Especially if they're gonna get mutated.

LipuFei: This only has one `return` in this function which is easier to read IMO.

plugins/LocalContainerProvider/LocalContainerProvider.py
LipuFei: Unfortunately, without it, I got this:
```
plugins/LocalContainerProvider/LocalContainerProvider.py:130: error: Need type annotation for 'requested_metadata'
```

plugins/LocalContainerProvider/LocalContainerProvider.py
ianpaschal: ok, fair enough

plugins/LocalContainerProvider/LocalContainerProvider.py
ianpaschal: Weird. Wonder why it's needed there but not when returning the empty dict.

PR #364, opened 2018-07-12
CURA-5476 Remove non-native file dialog

LipuFei:
@ianpaschal


Appesteijn:
The bug is back, but the filemenu is fancy...

PR #369, opened 2018-07-26
CURA-5594 Make sure there is MeshData before using it

LipuFei:


diegopradogesto:
Closing this because we decided to go for a different solution.

PR #374, opened 2018-08-03
Add some extra information when the FileWriters can't write the files.

diegopradogesto:
That will give some more information to the user when a message is shown in the frontend.

### DIFF
--- a/UM/FileHandler/FileWriter.py
+++ b/UM/FileHandler/FileWriter.py
@@ -11,6 +11,13 @@ class FileWriter(PluginObject):
 
     def __init__(self):
         super().__init__()
+        self._information = "" # type: str
 
     def write(self, stream, data):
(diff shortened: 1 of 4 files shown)


thopiekar:
Every time I see setInformation being used, something is send to the Logger, too.
How about moving it to setInformation, too?

BagelOrb:
Looking at the PR it seems that now users will get the message "There is no mesh to save." when they are trying to save the *gcode* before slicing.

Shouldn't there be a separate setInformation in the gcodewriter?

diegopradogesto:
Yes, but it's part of Cura. Here it is: https://github.com/Ultimaker/Cura/pull/4187
@thopiekar I agree, I will do it!

thopiekar:
Great, thanks 😉👍

PR #375, opened 2018-08-06
CURA-5538 Fix intersection function

LipuFei:
Home made solution has bugs. For example, if it's processing two almost identical rectangles (e.g. with 0.0001 differences), then it can return an empty Polygon, indicating that there is no intersection.

Companion PR: https://github.com/Ultimaker/cura-build/pull/184

### DIFF
--- a/UM/Math/Polygon.py
+++ b/UM/Math/Polygon.py
@@ -1,16 +1,12 @@
-# Copyright (c) 2016 Ultimaker B.V.
+# Copyright (c) 2018 Ultimaker B.V.
 # Uranium is released under the terms of the LGPLv3 or higher.
 
 import numpy
 import scipy.spatial
-import shapely.geometry
 
-from UM.Math.Float import Float #For fuzzy comparison of edge cases.
-from UM.Math.LineSegment import LineSegment #For line-line intersections for computing polygon intersections.
-from UM.Math.Vector2 import Vector2 #For constructing line segments for polygon intersections.
 from UM.Logger import Logger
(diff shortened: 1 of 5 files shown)


LipuFei:
The jenkins error should be gone once the companion PRs get merged.

LipuFei:
Closed the wrong PR...

ianpaschal:
screen shot 2018-08-08 at 14 01 19

😍

thopiekar:
Btw. that means for me that there is a new dependency I need to add to PPA packages right?

LipuFei:
@thopiekar Actually 2, `trimesh` and `shapely`

LipuFei:
Well, trimesh is not used at the moment, so you can skip it.

ianpaschal:
Well, we did decide today in Cookies & Cura that we're going to attempt to use Trimesh, so I'd include it @thopiekar

thopiekar:
Ok, shapely is included in Ubuntu's repository. Any minimum version we need?
And for trimesh I will need to make packaging files. Couldn't find it in Debian's repo either 🤔

LipuFei:
@thopiekar `shapely 1.6.4` and `trimesh 2.31.45`, see https://github.com/Ultimaker/cura-build-environment/pull/43

thopiekar:
Alright. Thank you for the summary.

In case the older version doesn't work I'll add an updated package then. 😉

LipuFei:
I did a few more refactor.

PR #378, opened 2018-08-07
CURA-5601 save open recent

LipuFei:


LipuFei:
@rburema I think the problem is not with the `LocalFileOutputDevice`, but the callback function `_onJobFinished()` in `QtApplication.py` which also decides whether a file should be saved into the recent files list. There it only save files into the list for certain read file jobs but not write file jobs.

See https://github.com/Ultimaker/Uranium/blob/3c3c5b3bd57569b581a254e120c4b3ee535b209c/UM/Qt/QtApplication.py#L269

ianpaschal:
This shouldn't have passed code review; CI is not green. 😉

UM/Qt/QtApplication.py
```diff
@@ -267,14 +268,23 @@ def recentFiles(self) -> List[QUrl]:
         return self._recent_files
 
     def _onJobFinished(self, job: Job) -> None:
-        if (not isinstance(job, ReadMeshJob) and not isinstance(job, ReadFileJob)) or not job.getResult():
+        if isinstance(job, WriteFileJob):
+            if not job.getResult() or not job.getAddToRecentFiles():
+                # For a write file job, if it failed or it doesn't need to be added to the recent files list, we do not
+                # add it.
+                return
+        elif (not isinstance(job, ReadMeshJob) and not isinstance(job, ReadFileJob)) or not job.getResult():
```
ChrisTerBeke: isinstance accepts a tuple of objects, so this could be reduced to isinstance(job, (ReadMeshJob, ReadFileJob)).

PR #380, opened 2018-08-08
".umplugin" added to the list of excluded extensions

bockha:
On ubuntu linux I had trouble using `create_plugin.py` since it tried to zip the newly created umplugin-file into itself until my laptop ran out of disk space.

This pr adds the extension `.umplugin` to the list of excluded extensions to solve this issue.


diegopradogesto:
@bockha in the current master branch, we don't longer support umplugins, instead we change to have packages, so this file is no longer needed or useful. In fact I was discussing with the rest of the team that the best option is to delete the `create_plugin.py` since is now useless.

Sorry for the misunderstanding. I will close this PR and delete the file. You can still use and create umplugins in previous versions.

PR #386, opened 2018-08-15
Increase APIVersion and remove OctoPrint hack

fieldOfView:
This PR increases the API Version and removes a hack to disallow certain versions of the OctoPrint plugin.

Commit 46b6c7e33e08154260db17eb1b8707604fa66d99 changed the way metadata for container stacks is handled in a way that makes certain plugins (such as the OctoPrintPlugin) incompatible with master. I can change my plugins to be compatible again, but I cannot change the 10k+ installations already in the field, so these must be invalidated by changing the plugin API version*. This means that all plugins must be updated before the next release (even if only updating their API versions, if there are no other incompatibilities).

*: note that I have already implemented a precaution in OctoPrintPlugin so it does not load with versions of Cura it has not been tested against at the time of releasing the plugin. So there will not be crashing Cura installs due to this issue, and all users must update their OctoPrintPlugin for the next release of Cura.

### DIFF
--- a/UM/PluginRegistry.py
+++ b/UM/PluginRegistry.py
@@ -41,7 +41,7 @@ if TYPE_CHECKING:
 #   [plugins]: docs/plugins.md
 
 class PluginRegistry(QObject):
-    APIVersion = 4
+    APIVersion = 5
 
     def __init__(self, application: "Application", parent: QObject = None) -> None:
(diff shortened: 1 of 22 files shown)


ChrisTerBeke:
Why was this merged? It breaks the unit tests.... :(

thopiekar:
:see_no_evil:

PR #387, opened 2018-08-15
Add shutdown hook and cleanup lock file upon exit

LipuFei:
- Add shutdown hook to Application
 - The application lock file should be removed upon exit

### DIFF
--- a/UM/Application.py
+++ b/UM/Application.py
@@ -3,6 +3,7 @@
 
 import argparse
 import os
+import signal
 import sys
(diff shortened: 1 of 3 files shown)


LipuFei:
@diegopradogesto We can close the timeout solution if this works.

diegopradogesto:
Hey @LipuFei , this PR can be closed, right? Or is still valid?

LipuFei:
Let's close it.

PR #389, opened 2018-08-16
CURA-5649 Allow to hide file types in save file dialog

LipuFei:
- New field "hide_in_file_dialog" for a file type.
 - For a write file request, provide a complete list of preferred file
   types and let it pick the best matching one.


UM/Qt/Bindings/OutputDeviceManagerProxy.py
```diff
@@ -68,11 +68,11 @@ def activeDeviceDescription(self) -> str:
     def requestWriteToDevice(self, device_id: str, file_name: str, kwargs: Mapping[str, str]) -> None:
         limit_mimetypes = kwargs.get("limit_mimetypes", None)
         file_type = kwargs.get("file_type", "mesh")
-        preferred_mimetype = kwargs.get("preferred_mimetype", None)
+        preferred_mimetypes = kwargs.get("preferred_mimetypes", None)
```
alekseisasin: I am worried about using different preferred_mimetype or preferred_mimetypes.
We still have many places where we use "preferred_mimetype"

PR #390, opened 2018-08-17
fix for CURA-5645 -- forgot to exclude other types of files from 'inc…

LipuFei:
…lude saved projects in recent files'


UM/Qt/QtApplication.py
```diff
@@ -263,6 +263,9 @@ def _onJobFinished(self, job: Job) -> None:
         elif (not isinstance(job, ReadMeshJob) and not isinstance(job, ReadFileJob)) or not job.getResult():
             return
 
+        if isinstance(job, WriteFileJob) and not job.getFileName().endswith(".curaproject.3mf"):
```
LipuFei: Uranium doesn't know about Cura, so we should not put this here. And we just want to above gcode here.

PR #391, opened 2018-08-19
QtApplication: Set Qt theme only when application is frozen

thopiekar:
Together with the PPA on Ubuntu/Linux the problem does not occur, because the system theme is properly fetched without this line being applied.
This simple check should avoid that the workaround is applied, when installing Cura via PPA on Ubuntu/Linux.

### DIFF
--- a/UM/Qt/QtApplication.py
+++ b/UM/Qt/QtApplication.py
@@ -122,8 +122,10 @@ class QtApplication(QApplication, Application):
         self._mesh_file_handler = MeshFileHandler(self) #type: MeshFileHandler
         self._workspace_file_handler = WorkspaceFileHandler(self) #type: WorkspaceFileHandler
 
-        # Remove this and you will get Windows 95 style for all widgets if you are using Qt 5.10+
-        self.setStyle("fusion")
+        # The following workaround is only meant to be used with Ultimaker's software distributions
+        if hasattr(sys, 'frozen'):
+            # Remove this and you will get Windows 95 style for all widgets if you are using Qt 5.10+
+            self.setStyle("fusion")
 
         self.setAttribute(Qt.AA_UseDesktopOpenGL)
         major_version, minor_version, profile = OpenGLContext.detectBestOpenGLVersion()


thopiekar:
Btw: I'm pretty sure that something else is causing the problem, because why should setting "fusion" bring the systems default theme if fusion is a fallback theme?
When setting it, I would expect the fusion to appear and not a system's default. When this is applied here on the PPA based install, then I really get the fusion theme.

diegopradogesto:
CI is failing!

thopiekar:
Oh, didn't noticed that. Will take a look onto it.

LipuFei:
@thopiekar I suspect that this may break the AppImage build too because its default theme is Windows 95 style.

thopiekar:
Alright, because this was the idea - catching all the distributions (AppImage, NSIS, etc) by Ultimaker by checking whether it is frozen or not.
Will try to fix the CI problems soon.

thopiekar:
Not sure, but think there was a PR which has been merged too early when I created this one.
Now after merging master into this PR, CI says that everything is good 😉

Ghostkeeper:
The code should be frozen inside the AppImage, too, yes. "Frozen" means that the code is embedded inside a place where it can't be edited (without admin rights). Examples are the AppImage, /usr on Linux, an App on MacOS, or Program Files on Windows.

thopiekar:
@Ghostkeeper Yeah, you are right. It is!
I manually removed that line and thought "frozen" is an applications which gets progressed through either cx_Freeze, py2win, py2app or anything else.
Maybe I can write something on a lower level. Adding an variable like "QT_LOCKED_THEME" and make it None by default, so it takes the system default.
Additionally respecting your work, I could add a CMake variable, which allows setting this variable inside Uranium.
What do you think?

Ghostkeeper:
In general I'm a proponent of making the source version use the same logic and the same theme as the official distribution (from Ultimaker). If the official distribution uses Fusion style dialogues then the source version should too. As far as I know, the distribution doesn't use Fusion style right now, but I could be wrong because this changed a couple of times.

thopiekar:
Ok. But if you are arguing like that, then you should also care about locking the color pallette to some defaults. For example, I have at many places still dark elements on dark background.
Your point is not bad at all, because I also tend to focus first on one thing before doing something complex (doesn't mean that it is easy for me). On the other hand, if things already mixed like this, then we could also take some time and make Cura directly fully responsive to the underlying OS.

Ghostkeeper:
If it's responding to OS-theme in the build, it should also respond to OS-theme in source. That way we can test with various OS-themes from source just as well.

We are agreeing that Cura should function without responding to OS-theme or completely responding to OS-theme, either one or the other. We're currently leaning to having the Cura theme be in complete control rather than the OS. Not there yet though, but that is another issue that is not solved by this PR. So we'll close this PR then. Sorry!

PR #392, opened 2018-08-22
Fix fx reset button

alekseisasin:
CURA-5646


LipuFei:
No change here. I will close this one.

PR #396, opened 2018-09-07
[master] Remove debugging message that fills the logs

diegopradogesto:
We don't need to know when a file is not being updated. Also add the `packages.json` file to the list of files that don't need updating.

### DIFF
--- a/UM/VersionUpgradeManager.py
+++ b/UM/VersionUpgradeManager.py
@@ -84,7 +84,7 @@ class VersionUpgradeManager:
         PluginRegistry.addType("version_upgrade", self._addVersionUpgrade)
 
         # Files that should not be checked, such as log files
-        self._ignored_files = ["uranium.lock", "plugins.json"]  # type: List[str]
+        self._ignored_files = ["uranium.lock", "plugins.json", "packages.json"]  # type: List[str]
 
     ##  Registers a file to be ignored by version upgrade checks (eg log files).
     #   \param file_name The base file name of the file to be ignored.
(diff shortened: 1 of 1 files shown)


fieldOfView:
Actually, I think it is important to log this; it means that these files have an unexpected version, which could mean they have inexpected content too. 

If this occurs often “in the field” it must mean something is potentially corrupting configuration files, and that is something that *should* be logged.

fieldOfView:
I do agree that package.json should be ignored by versionupgraders.

diegopradogesto:
Yes, but it's also outputting messages when the file is up-to-date and it has nothing to upgrade. Then it shows one message for each version upgrade that does nothing with the file.

fieldOfView:
Either way, it is commented-out code, which is against policy.

thopiekar:
I agree with removing dead code, but would remove the debug messages.
Usually, if people find the messages too noisy, then there must be a way to silence them. In case someone needs to debug an error, then there are log parsers to filter the logs.
(We had a ticket in the past about setting the verbosity level, but I think it has never been finished.)

Therefore I agree with @fieldOfView. These lines will be still useful in the future to understand why an upgrade fails or misbehaves.

But yeah, that's only my opinion.

diegopradogesto:
Oh, that commented-out code was my mistake. I was doing tests first and I forgot to actually remove the lines. Done in 95f0525

Regarding if the files are useful or not, I would say that the next log line that says "Create upgrade task for configuration file..." is the one that we want. In case this line is not logged when doing a version upgrade, then something is wrong. I think it is enough.

fieldOfView:
I misread the purpose of this PR, and now agree that it would be a great improvement.

PR #399, opened 2018-09-23
Complete Polish translation for version 3.5

drzejkopf:
Translation was incomplete, phrases related to functions implemented after version 3.3 were missing. I have translated all missing phrases in Polish translation of files cura.po, uranium.po and fdmprinter.def.json.po. For now, translation is complete. I tried running it on Cura 3.5 beta and I can't encounter any issues.

### DIFF
--- a/resources/i18n/pl_PL/uranium.po
+++ b/resources/i18n/pl_PL/uranium.po
@@ -8,15 +8,15 @@ msgstr ""
 "Project-Id-Version: Uranium 3.5\n"
 "Report-Msgid-Bugs-To: r.dulek@ultimaker.com\n"
 "POT-Creation-Date: 2018-09-19 17:07+0200\n"
-"PO-Revision-Date: 2018-03-30 20:39+0200\n"
-"Last-Translator: 'Jaguś' Paweł Jagusiak and Andrzej 'anraf1001' Rafalski\n"
+"PO-Revision-Date: 2018-09-21 22:02+0200\n"
+"Last-Translator: 'Jaguś' Paweł Jagusiak, Andrzej 'anraf1001' Rafalski and Jakub 'drzejkopf' Świeciński\n"
 "Language-Team: reprapy.pl\n"
 "Language: pl_PL\n"
 "MIME-Version: 1.0\n"
 "Content-Type: text/plain; charset=UTF-8\n"
 "Content-Transfer-Encoding: 8bit\n"
 "Plural-Forms: nplurals=2; plural=(n != 1);\n"
-"X-Generator: Poedit 2.0.6\n"
+"X-Generator: Poedit 2.1.1\n"
 
 #: /home/ruben/Projects/Uranium/plugins/Views/SimpleView/__init__.py:12
 msgctxt "@item:inmenu"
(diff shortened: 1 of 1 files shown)


resources/i18n/pl_PL/uranium.po
```diff
@@ -37,13 +37,13 @@ msgstr "Plik Wavefront OBJ"
 #: /home/ruben/Projects/Uranium/plugins/FileHandlers/OBJWriter/OBJWriter.py:23
 msgctxt "@error:not supported"
 msgid "OBJWriter does not support non-text mode."
-msgstr ""
+msgstr "Zapisywacz OBJ Nie obsługuje trybu nietekstowego."
```
thopiekar: Should "Nie" (which means "not") not be written lowercase?

resources/i18n/pl_PL/uranium.po
```diff
@@ -324,11 +326,14 @@ msgid ""
 "{error}.\n"
 "Please try to upgrade again later."
 msgstr ""
+"Wystąpił błąd podczas odinstalowywania pakietu {package} przed instalacją nowej wersji:\n"
+"{error}.\n"
+"Spróbuj zaktualizować ponownie później."
```
thopiekar: Last sentence: "Please" is missing.
Try to make the sentence more polite.

thopiekar:
I've been too slow :(

resources/i18n/pl_PL/uranium.po
diegopradogesto: @drzejkopf is that "Proszę, spróbuj zaktualizować ponownie później." correct for this?

I can do it directly in 3.5.

thopiekar: @diegopradogesto I have Polish roots, I can't write that stuff but read and talk :sweat_smile: 
Yes, your text is in my point of view correct. Just don't know whether they place a comma after the "Please". But I guess, it does not hurt to place more commas as in English.

But, yeah, wait for @drzejkopf better.

drzejkopf: It should be "Proszę spróbować zaktualizować ponownie później." to keep impersonal construction of sentence (it's good practice to use impersonal forms when translating promps to Polish), but it is common that promps and messages like this one are being translated without polite forms. Both versions are correct, none of them is impolite, but if you want more straightforward translation, you can use form I mentioned above :)

resources/i18n/pl_PL/uranium.po
drzejkopf: Yes, that's correct, it should be "Zapisywacz OBJ nie obsługuje trybu nietekstowego.", just a typo

diegopradogesto: I decide to keep the name "OBJWriter", because this is a plugin name and it shouldn't change.

resources/i18n/pl_PL/uranium.po
diegopradogesto: Ok. Thanks to both of you. Done in 9c04fcf89516dc3972efbe0d9df5bc71d561be4b

resources/i18n/pl_PL/uranium.po
drzejkopf: Of course, in my opinion it looks better, I translated it that way to keep the manner of translating "xxxWriter" and "xxxReader" plugins names as they were translated previously in Cura repository. I can revise and change all these names in Cura repository if you wish

thopiekar: @diegopradogesto Well, then "Zapisywacz OBJ" -> "OBJWriter".
@drzejkopf Right?

But I see the same also in other languages. Like Russian, if I'm not wrong. Just remember that I saw something like this.

drzejkopf:
@diegopradogesto  sorry for asking it here, but why Cura 3.5 stable don't have official support for Polish language? I've seen in Cura repository that it failed shell script check, so I assume it's problem with Uranium

thopiekar:
Official support means that the translations have not been done by Ultimaker. They have a partner company, who is translating into languages in which Ultimaker is personally (business-wise) interested.
That's all. Doesn't mean your translations are worse than theirs. :wink:

diegopradogesto:
@drzejkopf my fault. I forgot to enable the Polish on the list of available languages. I did enabled them yesterday here: 6abd43f6903c7e0e1f3509a5ffacb27d22115882

We plan to release a 3.5.1 in the next days to include the Polish translations and some bug fixes.

I really apologize for that.

PR #401, opened 2018-10-03
CURA-5791 Make a "singleton" OutputDevicesModel

LipuFei:
`OutputDevicesModel` has a connection from `OutputDeviceManager.outputDevicesChanged` which is a custom signal. When the custom signal gets trigger, it will invoke the callback function, regardless if the target is a `QtObject` and if it has been destroyed as a `QtObject`, and this can cause errors and crashes.

To fix this, make it a "singleton" model that can be retrieved from `QtApplication`. This way the custom signal connection will never invoke method in a destroyed `QtObject`.


nallath:
I was able to fix this in a much simpler way by setting the objectOwnership.

PR #403, opened 2018-10-05
Improve lock file management in Windows

diegopradogesto:
The Windows API is now used to create and delete the lock file. This file is now created with a flag DELETE_ON_CLOSE that deletes the file when closing or when the application gets interrupted.

### DIFF
--- a/UM/Application.py
+++ b/UM/Application.py
@@ -63,6 +63,8 @@ class Application:
         self._is_headless = False #type: bool
         self._use_external_backend = False #type: bool
 
+        self.CONFIG_LOCK_FILENAME = "{name}.lock".format(name = self._app_name)
+
         self._cli_args = None #type: argparse.Namespace
         self._cli_parser = argparse.ArgumentParser(prog = self._app_name, add_help = False) #type: argparse.ArgumentParser
 
--- a/UM/Application.py
+++ b/UM/Application.py
@@ -63,7 +63,7 @@ class Application:
         self._is_headless = False #type: bool
         self._use_external_backend = False #type: bool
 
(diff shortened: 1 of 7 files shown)


UM/Application.py
```diff
@@ -63,7 +63,7 @@ def __init__(self, name: str, version: str, build_type: str = "", is_debug_mode:
         self._is_headless = False #type: bool
         self._use_external_backend = False #type: bool
 
-        self.CONFIG_LOCK_FILENAME = "{name}.lock".format(name = self._app_name)
+        self.CONFIG_LOCK_FILENAME = "{name}.lock".format(name = self._app_name) # type: str
```
fieldOfView: Out of interest, why not ```self._config_lock_filename```? IE: why is this not a private variable like any other?

diegopradogesto: Oh, yes. I just copied as it was in the ContainerRegistry. It makes sense to change it.

thopiekar:
Are you sure that this functionality is not delivered by the tempfile module?
I remember that there was a function which was supposed to create a temporary file, which gets removed after being closed.
This sounds pretty close to what you want and have. Don't know whether it acts the same.

LipuFei:
@thopiekar temp file does automatically remove it when it's closed, but we need that file to be exclusively owned by the process that created it, thus the name "lock file".

thopiekar:
Hmm, I still have something in mind that tells me that there are no real
lockfiles on Windows.
Therefore I thought using a tempfile would be good for Windows only.

But yeah, if that's now or if it was always supported on windows, then
excuse the noise on this 😅

PR #406, opened 2018-10-09
Preferences: Making note about deprecation more clear

thopiekar:
Now the developer knows which trivial change needs to be done.

### DIFF
--- a/UM/Preferences.py
+++ b/UM/Preferences.py
@@ -224,7 +224,7 @@ class Preferences:
 
     ##  This method is still used by some external plugins and it needs to be kept as deprecated
     @classmethod
-    @deprecated("Please use the getPreferences function in Application", "3.3")
+    @deprecated("Please replace this function with Application.getPreferences() [Uranium]", "3.3")
     def getInstance(cls) -> "Preferences":
         from UM.Application import Application
         return Application.getInstance().getPreferences()


nallath:
Except that getPreferences() isnt a class function, so it should be Application.getInstance().getPreferences()

fieldOfView:
I don't agree with the revised text; it instructs the developer to change Preferences.py ("please change this function").

fieldOfView:
```Please use Application.getInstance().getPreferences() instead of Preferences.getInstance()```

thopiekar:
Oh, yes, that's correct @nallath

@fieldOfView: Can't follow on this. I'm recommending to replace one function call with another. I don't see where I tell someone to modify Preferences.py 🤷

fieldOfView:
The word "this" refers to something very close to the word this (as opposed to "that", which refers to something further away). The closest thing to what you refer to is the actual getInstance() method text. So you are telling me to replace the code of the Preferences.getInstance() method with the code of the Application.getInstance().getPreferences() method. Which to me is more confusing than the original text.

By using "use" instead of "replace with", you prevent that whole confusion.

thopiekar:
To be honest, I'm confused by this confusion now. I mean why should someone modify files delivered by Cura? 😅
But you are right with this. By choosing the correct words it is definite without exceptions.

nallath:
How about: @deprecated("Please use Application.getInstance().getPreferences() instead", "3.3")

nallath:
Fixed it in a53d0c37cacf0935105212b654b9604b212f405c

PR #407, opened 2018-10-11
[3.5][regression] Make sure visible rows are updated before trying to get index of a setting

fieldOfView:
This PR fixes the behavior of the "manage visible settings" button, when pressing the "cog" icon of a category in Cura.

This PR fixes ... hmm... can't find the issue. Has been reported in the forum, but don't know if it has been reported properly.

The issue is that when you press the "cog", Cura 3.5 no longer scrolls the SettingVisibility page to the category corresponding to the clicked cog.

Would be good if this could be merged into 3.5.1


nallath:
It was decided to merge this into master (and not 3.5).

nallath:
CURA-5823

fieldOfView:
Ok, I've changed the base branch... :-/

PR #408, opened 2018-10-17
Only show update once

Ghostkeeper:
When automatically checking for updates, only show the same update once. So it'll say once that there is an update for 3.5.1, but not again unless you manually check for updates. Once 3.6 rolls around, it'll show that there is an update again.

This is in line with how most update systems behave in other applications.

### DIFF
--- a/plugins/UpdateChecker/UpdateChecker.py
+++ b/plugins/UpdateChecker/UpdateChecker.py
@@ -31,7 +31,7 @@ class UpdateChecker(Extension):
 
         self._download_url = None
 
-        #Which version was the latest shown in the version upgrade dialog. Don't show these updates twice.
+        # Which version was the latest shown in the version upgrade dialog. Don't show these updates twice.
         Application.getInstance().getPreferences().addPreference("info/latest_update_version_shown", "0.0.0")
 
     ##  Connect with software.ultimaker.com, load latest.json and check version info.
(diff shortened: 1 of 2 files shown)


plugins/UpdateChecker/UpdateChecker.py
```diff
@@ -27,19 +27,26 @@ def __init__(self):
 
         Application.getInstance().getPreferences().addPreference("info/automatic_update_check", True)
         if Application.getInstance().getPreferences().getValue("info/automatic_update_check"):
-            self.checkNewVersion(True)
+            self.checkNewVersion(silent = True, display_same_version = False)
 
         self._download_url = None
 
+        #Which version was the latest shown in the version upgrade dialog. Don't show these updates twice.
```
nallath: PEP feels that there should be a space between # and the actual text.

Ghostkeeper:
Developers, see issue CURA-5832.

plugins/UpdateChecker/UpdateCheckerJob.py
```diff
@@ -96,4 +83,27 @@ def run(self):
...
-            Message(i18n_catalog.i18nc("@info", "No new version was found."), title = i18n_catalog.i18nc("@info:title", "Version Upgrade")).show()
\ No newline at end of file
+            Message(i18n_catalog.i18nc("@info", "No new version was found."), title = i18n_catalog.i18nc("@info:title", "Version Upgrade")).show()
+
+    def showUpdate(self, newest_version: Version, download_url: str) -> None:
+        preferences = Application.getInstance().getPreferences()
+        latest_version_shown = preferences.getValue("info/latest_update_version_shown")
+        if latest_version_shown == newest_version and not self.display_same_version:
+            return #Don't show this update again. The user already clicked it away and doesn't want it again.
+        preferences.setValue("info/latest_update_version_shown", str(newest_version))
```
diegopradogesto: I would suggest to move these 5 lines out to a different method, in order to separate concerns and being sure that the `showUpdate` only shows the update message.

PR #409, opened 2018-10-18
Change text requested by MJ

Appesteijn:


fieldOfView:
The .pot file is automatically generated. There's little sense in changing the text there. Change the text in UpdateCheckerJob.py instead (as you did for some strings, but not all).

Appesteijn:
Don't want to change the application name to Ultimaker Cura because I don't know what other things will break then..

nallath:
Changing the application name to Ultimaker Cura is a really bad idea. That would have broken a lot of resource loading stuff.

fieldOfView:
Note that I did not suggest in any way that the application_name should be changed (please, no!), but you changed one occurrence of Cura into Ultimaker Cura in the .pot that you did not change in the .py (and changing them in the .pot has no effect anyway). ```Cura {0} is available!```

diegopradogesto:
Closing this PR in favor of https://github.com/Ultimaker/Uranium/pull/413

PR #414, opened 2018-11-01
Fix passing definitioncontainers metadata to QML

fieldOfView:
This PR fixes passing definitioncontainers metadata to QML. For fully loaded definitions, the metadata is an OrderedDict which does not pass to QML correctly. This became apparent in Cura here:
https://github.com/Ultimaker/Cura/pull/4498#issuecomment-434966876


fieldOfView:
Tag: CURA-5847

PR #416, opened 2018-11-01
Cura 5840 make plugin sdk use semantic versioning

nallath:
CURA-5840

See also; https://github.com/Ultimaker/Cura/pull/4705


UM/Version.py
```diff
@@ -14,7 +15,7 @@ class Version:
     #   the major, minor and revision version numbers. All text is ignored.
     #
     #   \param version A string or bytes representing a version number.
-    def __init__(self, version: Union[str, bytes, List[Union[int, str]]]) -> None:
+    def __init__(self, version: Union[str, bytes, List[Union[int, str]], int]) -> None:
```
ChrisTerBeke: It might be a bit more readable if int was together with str and bytes instead of behind the List thing.

UM/PluginRegistry.py
```diff
@@ -397,14 +395,23 @@ def loadPlugin(self, plugin_id: str) -> None:
...
-        if self._metadata[plugin_id].get("plugin", {}).get("api", 0) != self.APIVersion:
+        # If API version is incompatible, don't load it.
+        plugin_api_version = self._metadata[plugin_id].get("plugin", {}).get("api", Version("0"))
+
+        if plugin_api_version.getMajor() != self.APIVersion.getMajor():
             Logger.log("w", "Plugin %s uses an incompatible API version, disabling", plugin_id)
             self.disablePlugin(plugin_id)
             return
 
+        if plugin_api_version > self.APIVersion:
```
alekseisasin: @ChrisTerBeke 
According to the SDK versioning document in confluence:

5.2 → 5.2.0 | 5.1 → 5.1.0 | --> The major number is the same, and the minor number requirement is lower can currently available, so we load it.

I changed the plugin "Console Logger" API to 5.2 and   "self.APIVersion" to 5.1. The plugin is loaded but disabled. I think we need to update the document.

ChrisTerBeke: No, we should fix the behavior in Cura then. A plugin 'created' for 5.1 should load fine in 5.2 as that version change should only include new functionality and no breaking changes (as per SemVer specification).

tests/TestVersion.py
```diff
@@ -3,30 +3,53 @@
...
-    def test_other_comparisons(self):
-        self.assertFalse(Version("1.1.0-beta.2") == Version("1.1.0-alpha.2"))
-        self.assertFalse(Version("1.1.0-beta.2") > Version("1.1.0-alpha.2"))
-        self.assertFalse(Version("1.1.0-beta.2") < Version("1.1.0-alpha.2"))
+
+def check_version_equals(first_version: Version, second_version: Version):
+    assert first_version == second_version
+    assert first_version.getMajor() == second_version.getMajor()
+    assert first_version.getMinor() == second_version.getMinor()
+    assert first_version.getRevision() == second_version.getRevision()
```
alekseisasin: Again according to the SDK versioning document:

Patch: When this number changes there are no changes in the SDK, but only bug fixes behind the scenes. Everything is fully compatible with the previous versions in the same major and minor range.

tests/TestVersion.py
```diff
@@ -3,30 +3,53 @@
...
+@pytest.mark.parametrize("first_version", major_versions)
+@pytest.mark.parametrize("second_version", major_versions)
+def test_major_equals(first_version, second_version):
+    check_version_equals(first_version, second_version)
+
+
+@pytest.mark.parametrize("first_version", major_minor_versions)
+@pytest.mark.parametrize("second_version", major_minor_versions)
+def test_major_and_minor_equals(first_version, second_version):
+    check_version_equals(first_version, second_version)
```
alekseisasin: Something goes wrong with these tests: I added print("Compare: "+ str(first_version) + " _ " + str(second_version))
All the time it compares same versions, I get each time 
Compare: 1.2.0 _ 1.2.0
Compare: 1.2.0 _ 1.2.0
...

nallath: This is intentional; it's creating a bunch of Version objects in all the ways that it can and subsequently compare them. What this test checks is that Version("1.1") is indeed the same as Version([1,1])

UM/PluginRegistry.py
nallath: If Cura has plugin version 5.1 and the plugin needs 5.2 to work, it makes total sense that it doesn't work... 

Cura 5.1, plugin 5.0 (Should work)
Cura 5.1, plugin 5.1 (Should work)
Cura 5.1, plugin 5.2 (Should not work) 
Cura 6.0, plugin 5.2 (Should not work (different major version nr))

As far as I can see, the code does exactly that.

ChrisTerBeke: Ah wait, @nallath is correct. The other way around should work, but it should not load when Cura is at a lower version than the plugin requires.

nallath: Ah, I found the issue; It used to disable plugins if they had a wrong version number at one point. After that, they never got re-evaluated. I've fixed that now.

LipuFei: I moved the version compatibility check into a separate function and added unit tests for it in 
14c7523

UM/Application.py
```diff
@@ -44,13 +45,15 @@ class Application:
     #   \param version \type{string} Version, formatted as major.minor.rev
     #   \param build_type Additional version info on the type of build this is, such as "master".
     #   \param is_debug_mode Whether to run in debug mode.
-    def __init__(self, name: str, version: str, app_display_name: str = "", build_type: str = "", is_debug_mode: bool = False, **kwargs) -> None:
+    def __init__(self, name: str, version: str, api_version: str = "5.0.0", app_display_name: str = "", build_type: str = "", is_debug_mode: bool = False, **kwargs) -> None:
```
ChrisTerBeke: I think passing a default value here is dangerous as it can unexpectedly set the SDK version without knowing about it.

LipuFei: I agree. Fortunately, nothing else seems to depend on this "hard-coded" value. Changed this in 78b8372c

PR #421, opened 2018-11-08
Search upgrade files in more levels deeper

diegopradogesto:
Allow the version upgrade to go more levels in depth to search for files to upgrade.

Contributes to CURA-5915.

### DIFF
--- a/UM/VersionUpgradeManager.py
+++ b/UM/VersionUpgradeManager.py
@@ -244,7 +244,6 @@ class VersionUpgradeManager:
     #   \return The filename of each file relative to the specified directory.
     def _getFilesInDirectory(self, directory: str) -> Iterator[str]:
         for (path, directory_names, file_names) in os.walk(directory, topdown = True):
-            directory_names[:] = []  # Only go to one level.
             for filename in file_names:
                 relative_path = os.path.relpath(path, directory)
                 yield os.path.join(relative_path, filename)


UM/VersionUpgradeManager.py
```diff
@@ -244,7 +244,6 @@ def _findShortestUpgradeRoutes(self) -> Dict[Tuple[str, int], Tuple[str, int, Ca
     #   \return The filename of each file relative to the specified directory.
     def _getFilesInDirectory(self, directory: str) -> Iterator[str]:
         for (path, directory_names, file_names) in os.walk(directory, topdown = True):
-            directory_names[:] = []  # Only go to one level.
```
LipuFei: @diegopradogesto I think the reason this only goes one level deep is because of the `old` folders. I would suggest to avoid creating the `old` folders so we avoid further issues.

LipuFei: Ah, `old` is already ignored. Then this is fine.

diegopradogesto: Wait, now I just realize that the version upgrade is also analyzing all the files in the plugins folder, looking for version number of those files. That will take a lot of time if people have many plugins with lots of files.

fieldOfView:
Awesome, now versionupgrader loads, parses and skips all installed plugins! My logs are getting more and more beautiful.

fieldOfView:
I made #424 to keep my logs somewhat more sane.

PR #423, opened 2018-11-14
Remove tray icon while the application is still alive

fieldOfView:
This PR actively removes the tray icon before the application is shut down, so the tray icon does not get orphaned. Without this PR, an orphan tray icon is left behind each time Cura is shut down.

prevents this

### DIFF
--- a/UM/Qt/QtApplication.py
+++ b/UM/Qt/QtApplication.py
@@ -417,6 +417,9 @@ class QtApplication(QApplication, Application):
         Logger.log("d", "Shutting down %s", self.getApplicationName())
         self._is_shutting_down = True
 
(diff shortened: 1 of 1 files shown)


Ghostkeeper:
Developers, see issue CURA-5935.

fieldOfView:
It would be a shame if this persisted in Cura 4.0

fieldOfView:
Just to emphasize that this simple, low-risk fix needs to be in Cura 4.0, I have changed the base branch to 4.0. This issue is (unjustifiably) seen as an indication of a memory leak: https://community.ultimaker.com/topic/24900-cura-35-memory-leak-in-monitor/?page=2&tab=comments#comment-222467

ChrisTerBeke:
Merged together with @nallath

PR #424, opened 2018-11-14
Ignore plugins folder when checking files for version upgrade

fieldOfView:
This PR ignores the plugins folder when checking files for version upgrade. https://github.com/Ultimaker/Uranium/pull/421 changed the version upgrade manager to recurse into all configuration folders, which also causes plugin folders to be examined. VersionUpgradeManager cannot handle plugin files, so the logs get spammed with lots of files being skipped.

The PR also changes the regex for plugins.json and package.json, to be able to handle .\plugins.json and .\package.json

### DIFF
--- a/UM/VersionUpgradeManager.py
+++ b/UM/VersionUpgradeManager.py
@@ -84,7 +84,7 @@ class VersionUpgradeManager:
         PluginRegistry.addType("version_upgrade", self._addVersionUpgrade)
 
         #Regular expressions of the files that should not be checked, such as log files.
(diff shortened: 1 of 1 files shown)


fieldOfView:
The logs become particularly nasty in 3.6 when the Cura Settings Guide plugin is installed.

nallath:
CURA-5956

nallath:
@Ghostkeeper Has fixed this during his valiant quest against the elusive NoneType crash during the upgrade process.

PR #427, opened 2018-12-05
CURA-5867 scale tool

LipuFei:


LipuFei:
@alekseisasin you can revert the last commit e8fdd7f

PR #430, opened 2018-12-11
[4.0] CURA-6005 Add code to handling bundled and installed packages

LipuFei:
This PR fixes issues in scenarios such as: a user has installed plugin `MyPlugin` version `1.0` in a previous Cura release and in the new Cura release, `MyPlugin` version `1.1` is bundled. In this case, the old installed `MyPlugin` should be removed to favor the newer bundled version.


UM/PackageManager.py
```diff
@@ -102,6 +102,56 @@ def _loadManagementData(self) -> None:
                 Logger.log("i", "User package management file %s doesn't exist, do nothing", self._user_package_management_file_path)
                 return
 
+        # For packages that become bundled in the new releases, but a lower version was installed previously, we need
+        # to remove the old lower version that's installed in the user's folder.
```
Ghostkeeper: Do we? Should the package manager not just load the most modern version of the two? I thought we spend quite some thought on this about a year ago to ensure that.

UM/PackageManager.py
```diff
@@ -102,6 +102,56 @@ def _loadManagementData(self) -> None:
...
+    # Compares the SDK versions and the package versions of the given package info dicts.
+    # Returns -1, 0, 1 indicating if the versions in dict1 is lower than, equal to, or higher than dict2.
+    #  - The package with the higher SDK version is considered having the higher version number. If they are the same,
+    #  - if the bundled package version is greater than or equal to the given package, -1 is returned. Otherwise, 1.
+    def _comparePacakgeWithBundledVersions(self, info_dict, bundled_info_dict) -> int:
+        # If the bundled version has a higher SDK version, use the bundled version by removing the installed one.
+        info_dict1 = Version(info_dict["sdk_version"])
+        info_dict2 = Version(bundled_info_dict["sdk_version"])
+        if info_dict1 < info_dict2:
+            return -1
```
Ghostkeeper: This could be summarised by just removing all packages that are not using the current major SDK version.

UM/PackageManager.py
```diff
@@ -102,6 +102,56 @@ def _loadManagementData(self) -> None:
...
+                to_remove_package_ids.add(package_id)
+                continue
+        for package_id in to_remove_package_ids:
+            del self._to_install_package_dict[package_id]
+
+    # Compares the SDK versions and the package versions of the given package info dicts.
+    # Returns -1, 0, 1 indicating if the versions in dict1 is lower than, equal to, or higher than dict2.
+    #  - The package with the higher SDK version is considered having the higher version number. If they are the same,
+    #  - if the bundled package version is greater than or equal to the given package, -1 is returned. Otherwise, 1.
+    def _comparePacakgeWithBundledVersions(self, info_dict, bundled_info_dict) -> int:
```
Ghostkeeper: Missing typing and documentation of these parameters.

UM/PackageManager.py
LipuFei: Unfortunately, `PluginRegistry` doesn't do that.

UM/PackageManager.py
LipuFei: Typing added in 46996ee

PR #433, opened 2018-12-12
Accept multiple values in regular expression

alekseisasin:
CURA-6011


UM/Settings/ContainerQuery.py
```diff
@@ -73,6 +73,8 @@ def execute(self, candidates: Optional[List[Any]] = None) -> None:
             if isinstance(value, str):
                 if ("*" or "|") in value:
                     key_filter = lambda candidate, key = key, value = value: self._matchRegExp(candidate, key, value)
+                elif value[0] == "[" and value[-1] == "]":
```
nallath: Didn't putting | in the filter work? That's what the two lines above this seem to imply.

LipuFei:
I'll do a rebase for this branch. It's not based on `4.0`

UM/Settings/ContainerQuery.py
LipuFei: Fixed this by moving the if-block to be the first one. Made a few changes to the matching pattern and added a unit test for this.

tests/Settings/TestContainerQuery.py
```diff
@@ -0,0 +1,49 @@
...
+                   "value": "[abc|_test_|test2|tst]",
+                   },
+         "match": False,
+         },
+    ]
+
+    for test_case in test_cases:
+        result = cq._matchRegMultipleTokens(**test_case["input"])
+        if test_case["match"]:
+            print("!!!! result = ", result)
```
Ghostkeeper: Please remove this debug print.

Ghostkeeper: Nevermind, I already removed it.

PR #459, opened 2019-01-08
CURA-6085 Add support for lang-specific fonts in Theme

LipuFei:


fieldOfView:
Now we can have a pirate font on talk like a pirate day!
https://www.fontspace.com/gemfonts/treasure-map-deadhand

PR #415, opened 2018-11-01
STLReader: Adding fallback to legacy code

thopiekar:
For some reason, I have trouble with NumPy-STL provided by Ubuntu when using my PPA.
Therefore took some time digging into this issue and added this fallback, because the legacy code works (still) just perfectly fine!

PR #460, opened 2019-01-14
Add supported_sdk_versions as a way to indicate support for multiple …

nallath:
…versions

CURA-6087

### DIFF
--- a/UM/PluginRegistry.py
+++ b/UM/PluginRegistry.py
@@ -404,10 +404,16 @@ class PluginRegistry(QObject):
             return
 
         # If API version is incompatible, don't load it.
-        plugin_api_version = self._metadata[plugin_id].get("plugin", {}).get("api", Version("0"))
-        if not self._isPluginApiVersionCompatible(plugin_api_version):
-            Logger.log("w", "Plugin [%s] with API version [%s] is incompatible with the current API version [%s].",
-                       plugin_id, plugin_api_version, self._api_version)
+        supported_sdk_versions = self._metadata[plugin_id].get("plugin", {}).get("supported_sdk_versions", [Version("0")])
+        is_plugin_supported = False
+        for supported_sdk_version in supported_sdk_versions:
+            is_plugin_supported |= self._isPluginApiVersionCompatible(supported_sdk_version)
+            if is_plugin_supported:
+                break
+
+        if not is_plugin_supported:
+            Logger.log("w", "Plugin [%s] with supported sdk versions [%s] is incompatible with the current sdk version [%s].",
+                       plugin_id, [str(version) for version in supported_sdk_versions], self._api_version)
             self._outdated_plugins.append(plugin_id)
             return
 
@@ -627,12 +633,15 @@ class PluginRegistry(QObject):
                         raise InvalidMetaDataError(plugin_id)
 
                     # Check if the plugin states what API version it needs.
-                    if "api" not in meta_data["plugin"]:
-                        Logger.log("e", "api must be set!")
+                    if "api" not in meta_data["plugin"] and "supported_sdk_versions" not in meta_data["plugin"]:
+                        Logger.log("e", "The API or the supported_sdk_versions must be set!")
                         raise InvalidMetaDataError(plugin_id)
                     else:
                         # Store the api_version as a Version object.
-                        meta_data["plugin"]["api"] = Version(meta_data["plugin"]["api"])
+                        if "supported_sdk_versions" in meta_data["plugin"]:
+                            meta_data["plugin"]["supported_sdk_versions"] = [Version(supported_version) for supported_version in meta_data["plugin"]["supported_sdk_versions"]]
+                        else:
+                            meta_data["plugin"]["supported_sdk_versions"] = [Version(meta_data["plugin"]["api"])]
 
                     if "i18n-catalog" in meta_data["plugin"]:
                         # A catalog was set, try to translate a few strings
--- a/UM/PluginRegistry.py
+++ b/UM/PluginRegistry.py
@@ -638,7 +638,7 @@ class PluginRegistry(QObject):
                         raise InvalidMetaDataError(plugin_id)
                     else:
                         # Store the api_version as a Version object.
-                        all_supported_sdk_versions = []
+                        all_supported_sdk_versions = []  # type: List[Version]
                         if "supported_sdk_versions" in meta_data["plugin"]:
                             all_supported_sdk_versions += [Version(supported_version) for supported_version in meta_data["plugin"]["supported_sdk_versions"]]
                         if "api" in meta_data["plugin"]:
--- a/UM/PluginRegistry.py
+++ b/UM/PluginRegistry.py
@@ -373,7 +373,7 @@ class PluginRegistry(QObject):
                     pass
 
     # Checks if the given plugin API version is compatible with the current version.
-    def _isPluginApiVersionCompatible(self, plugin_api_version: "Version") -> bool:
+    def isPluginApiVersionCompatible(self, plugin_api_version: "Version") -> bool:
         return plugin_api_version.getMajor() == self._api_version.getMajor() \
                and plugin_api_version.getMinor() <= self._api_version.getMinor()
 
@@ -407,7 +407,7 @@ class PluginRegistry(QObject):
         supported_sdk_versions = self._metadata[plugin_id].get("plugin", {}).get("supported_sdk_versions", [Version("0")])
         is_plugin_supported = False
         for supported_sdk_version in supported_sdk_versions:
-            is_plugin_supported |= self._isPluginApiVersionCompatible(supported_sdk_version)
+            is_plugin_supported |= self.isPluginApiVersionCompatible(supported_sdk_version)
             if is_plugin_supported:
                 break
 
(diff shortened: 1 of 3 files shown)


thopiekar:
Isn't it better to call the variable "supported_api_versions"?
From what I understand it should do the same as "api", but allowing a list of api versions.
Furthermore, wouldn't it be more logical to announce a "minimum_api" variable for the lowest value?
Normally, you are only interested to exclude older versions after a time.
Using a list you are kind of forced to update it in your package after each Cura release.
Sounds to me like too much maintenance work.

diegopradogesto:
> Furthermore, wouldn't it be more logical to announce a "minimum_api" variable for the lowest value?

But what if you want to skip one version? We thought that then it's better to indicate a list.

nallath:
Forcing the user to specify what numbers are supported is by design. We can't guarantee that a new major number won't break things (that's why it's a major change).  

So a "6.0.0" plugin is going to be compatible with "6.1.0" sdk of Cura, but a "5.0.0" plugin is only compatible with "6.0.0" if "6.0.0" is also in the supported list.

thopiekar:
Ok, so if a plugin is intended to work with multiple api versions I should keep the "supported_api_versions" list up-to-date and if a plugin is only intended to work with one version, when I should use "api", right?

ChrisTerBeke:
@thopiekar using the supported_api_versions list but with just a single entry should also work I think?

diegopradogesto:
> @thopiekar using the supported_api_versions list but with just a single entry should also work I think?

Yes unless you want your plugin being loaded in 3.5 or 3.6 (aka api=5), in this case you need to keep it because Cura 3.5 or 3.6 don't understand "supported_sdk_versions"

ChrisTerBeke:
> > @thopiekar using the supported_api_versions list but with just a single entry should also work I think?
> 
> Yes unless you want your plugin being loaded in 3.5 or 3.6 (aka api=5), in this case you need to keep it because Cura 3.5 or 3.6 don't understand "supported_sdk_versions"

Ah true, this is for newer versions of Cura only ofc

thopiekar:
Good point @ChrisTerBeke! So since it is your intention to force people maintaining this list, wouldn't it be a good idea to show a warning in the logs to use "supported_api_versions" instead?
Something like a deprecation warning?

@diegopradogesto But for cases, where nobody cares about older versions, he should better use "supported_api_versions", right? Don't see a reason, why plugin maintainers want to support a year old (or even older version) of Cura and test against all of these. Takes too much time and got no profit in development in caring about this.
Didn't take a deep look into @nallath's changes, but I would ignore the "api" variable, whenever "supported_sdk_versions" is present.

diegopradogesto:
Yes @thopiekar that is the case. If the key `supported_sdk_versions` is present, then the `api` is ignored. Of course if you don't want to support older versions, it's better if you get rid of the `api` key and use `supported_sdk_versions` instead.

thopiekar:
So how about adding the deprecation warning in case "api" is present only?
Would be great to have it included, so this change would be noted to the developers automatically, whenever they look into the logs. That's at least a place where most plugins developers should look at, I think 😉

ChrisTerBeke:
A deprecation warning for the 'old' field would be nice indeed.

fieldOfView:
> (this never was in the legacy system).

A workaround would be if the backend would be capable of retaining older versions of the plugin for older versions of the SDK/API.

ChrisTerBeke:
> > (this never was in the legacy system).
> 
> A workaround would be if the backend would be capable of retaining older versions of the plugin for older versions of the SDK/API.

We don't plan on adding support to patch legacy situations for now as this would require significant changes in this case.

thopiekar:
Can you explain why supporting 4 and 5 is not possible?
While working on the plugins through different Cura versions, I tried to keep them alive over all APIs by adding if-clauses, which check for the Cura version (since I had no clue how to get the API enum) and added these kinds of lines:

image

Therefore I'm pretty sure my plugins work on 2.7 for example, too. :sweat_smile:

nallath:
Well, I still think we should consider it. It makes quite a bit of sense to have it. Now it puts unneeded strain on plugin developers (we should decrease the burden on outside contributors because it should be as easy as possible to create content for Cura)

ChrisTerBeke:
@thopiekar although the plugin might work, the `api` field in `package.json` only allows a single value, and Cura versions older than 4.0 only look at that field when loading plugins. This is just how Cura used to work.

thopiekar:
Well, by setting api=5 and supported_sdk_versions=["5.0.0", "6.0.0"] I don't see a reason why all Cura versions couldn't be supported. Older Cura versions will be fine with api=5 and newer Cura versions will also be fine with what I set for supported_sdk_versions.
Or am I missing something? :confused:

ChrisTerBeke:
No, Cura's before 3.5 will look at the `api` field and see that it's 5, but they expect 4 or even lower. So they won't load the plugin.

thopiekar:
@ChrisTerBeke Ah, ok. But then it is your problem with the contributers portal, no? :laughing: 
Then you need to generate a package for each API/SDK version..

.. (On your latest post..): Ahh. That's crap :disappointed:

ChrisTerBeke:
@thopiekar It was deliberately not supported to not have to deal with legacy situations all the time. We know from usage statistics that there are not many people using Cura 3.5 and earlier.

thopiekar:
@ChrisTerBeke No problem. Just believed all the time that this is possible. In that case, your contributors portal should also mention this and prevent a passed check on Cura 2.7+.
Anyway, good to know and yes, never heard any complaints on needed or missing backwards compatibility, so doesn't seem that legacy features are needed that much. :wink:

fieldOfView:
> It was deliberately not supported to not have to deal with legacy situations all the time. 

So, now you've come to the realization that that decision was not in the best interest of your contributors OR your users, you make the deliberate choice to fix that, right? Because this issue is not just one time; it will continue to be a problem with all future Cura/API updates for some plugins.

ChrisTerBeke:
We just pushed a feature where if there's incorrect configuration between your plugin.json and the selected versions in the UI.

@fieldOfView not sure, depends on all parties involved. Anyway I'm not here to discuss that again.

fieldOfView:
Will Ultimaker accept a plugin that provides an alternative Marketplace?

ChrisTerBeke:
> Will Ultimaker accept a plugin that provides an alternative Marketplace?

Not up to me, but I doubt it...

Ghostkeeper:
> We know from usage statistics that there are not many people using Cura 3.5 and earlier.

But we do know that three quarters of our users are currently using 3.6. So for 5 out of 9 weeks, the plug-in developers have submitted SDK6 versions of their plug-ins that removed their plug-in for three quarters of the users during the 4.0 beta period.

I think it makes sense to keep supporting Cura versions from 2 months ago.

fieldOfView:
> we do know that three quarters of our users are currently using 3.6.

What's the rest using? Not all 4.0 beta, right? Sounds like a sizeable number of people are using older versions of Cura after all...

thopiekar:
@fieldOfView
A marketplace which finally cares about which OS is running? Waiting for this for a while, but whenever people contact me on this, then I mention that Ultimaker didn't put attention on this so far.
Even if I had the time and nerves: The server-side is closed source and even trivial PR's[1] don't go through.
Therefore: Why should I care? I better focus then on other things. Time is limited... :grimacing:

Well difficult to say. I have never seen any statistics. I even have the feeling that many people are (and want) to use the most recent version. At least from what I know from my work on the PPA.
In my opinion, there must be a conclusion found for the future. Something that is (a) pain-free and easy to use[2] and easy to distribute as @fieldOfView (probably) wants as I do.

As mentioned before, my code (should) works across many Cura versions and I don't care about how to reach all the Cura users. Since you introduced the contributors portal I trust that you are doing everything possible to reach as many people as possible 😉 👍 

[1] - https://github.com/Ultimaker/Uranium/pull/415
[2] - https://github.com/Ultimaker/Uranium/pull/460#issuecomment-454344554 @nallath

ChrisTerBeke:
@thopiekar filtering on OS (or other stuff) could theoretically be supported by using the tags feature we have implemented. The Marketplace UI in Cura is simply not using that yet, but might if there's more content in the future. But hey, let's start small and see where it goes.

thopiekar:
@ChrisTerBeke Yep, as mentioned before, I have no other choice :rofl: 
By the way, I resubmitted the updated plugin with the changes we discussed before[1].
Don't know who tests the plugins, but mentioned you in the notes and that we talked about the problem here.
Just for you to be warned :stuck_out_tongue_closed_eyes:

thopiekar:
Nope, I guess most of the 3.2.1 users are Debian (incl. all devariants)
users. Cura 3.2.1 is in the software repositories for a while. But since
@nallath has access to the stats, I guess he can answer the best how many
Linux users are included in 3.2.1 😉

>

nallath:
3.4.1 is also about 5%
3.3.0 is ~ 2%
3.1.0 is ~1%

From the stats is also pretty clear to see that the 75% that upgrades tend to do so within a week or two, but the people that stay with an old version tend to stay with it a long time (indicating that it's a conscious decision, since they get popups every time a newer version is there).

PR #462, opened 2019-01-18
CURA-6094 Prevent intersection attempt (and crash) on nonvalid polygons.

rburema:


rburema:
Closing this PR, because since it is a bugfix, not a feature, it should be for 4.0 instead of master.

PR #463, opened 2019-01-21
Prevent intersection attempt (and crash) on nonvalid polygons.

rburema:
Made a new PR for 4.0, instead of master, since it was a bugfix, not a feature.


rburema:
New 'vesion' of PR #462

PR #467, opened 2019-02-12
CURA-6120 arrange child nodes fix

rburema:
Need this polygon operation for a fix in Cura (child-nodes like support-blocker where arranged separatelty when 'Arrange Models' was selected).

### DIFF
--- a/UM/Math/Polygon.py
+++ b/UM/Math/Polygon.py
@@ -161,7 +161,7 @@ class Polygon:
         if polygon_union.area == 0:
             return Polygon()
 
-        return Polygon(points = [list(p) for p in polygon_union.exterior.coords[:4]])
+        return Polygon(points = [list(p) for p in polygon_union.exterior.coords[:-1]])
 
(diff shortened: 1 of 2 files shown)


UM/Math/Polygon.py
```diff
@@ -141,6 +141,28 @@ def intersectionConvexHulls(self, other: "Polygon") -> "Polygon":
...
 
+    #  Computes the convex hull of the union of the convex hulls of this and another polygon.
+    #
+    #   \param other The other polygon to combine convex hulls with.
+    #   \return The convex hull of the union of the two polygons' convex hulls.
+    def unionConvexHulls(self, other: "Polygon") -> "Polygon":
+        my_hull = self.getConvexHull()
+        other_hull = other.getConvexHull()
+
+        if len(my_hull._points) <= 2:
```
nallath: We should probably change the isValid function to do this check, since any polygon with <=2 points is not valid.

PR #469, opened 2019-02-25
Fix install prefix when installing in symlinked directory

Ghostkeeper:
If there is a symlink to the installation directory and it appears before the actual installation directory on the user's PATH, it would get a wrong path for `sys.argv[0]`. This then messes up the install location which consequently messes up the data and config storage directories.
This solution is provided by CapnKernel and should resolve the symlink so that sys.argv[0]='/usr/bin/python3' again and the storage data and config storage directories become something like /usr/share/... again.

This should fix https://github.com/Ultimaker/Cura/issues/5290 .

### DIFF
--- a/UM/Application.py
+++ b/UM/Application.py
@@ -398,9 +398,10 @@ class Application:
     @staticmethod
     def getInstallPrefix() -> str:
         if "python" in os.path.basename(sys.executable):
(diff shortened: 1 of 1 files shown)


Ghostkeeper:
Developers, see issue CURA-6251.

PR #475, opened 2019-03-20
New onboarding flow

Ghostkeeper:
Still a work in progress, but we're starting a review.

This changes the first launch sequence, by redesigning the add printer wizard and consolidating the many messages and pop-ups into a wizard.

This is a prerequisite of https://github.com/Ultimaker/Cura/pull/5449.

Implements CURA-6057 [ETA by rburema: and it's constituent CURA-6294]


rburema:
@nallath Thanks, I fixed those things in [commit ce6d274|https://github.com/Ultimaker/Uranium/pull/475/commits/ce6d27456b440b3abf2b0bd3318e4eaa6a1c67df] except for the what if it isn't bytes, that's only _properly_ in the next one.

UM/OutputDevice/OutputDeviceManager.py
```diff
@@ -218,8 +265,58 @@ def removeOutputDevicePlugin(self, plugin_id: str) -> None:
...
+        for priority in priority_order:
+            if priority not in plugins_by_priority:  # None of the plugins that specified 'NO' will be requested.
+                continue
+            for plugin_id in plugins_by_priority[priority]:
+                plugin = self._plugins[plugin_id]
+                if plugin:
+                    plugin.addManualDeviceSignal.connect(self._onManualDeviceAdded)
+                    plugin.removeManualDevice("", address)
+                    plugin.addManualDevice(address)
+                    return
```
Ghostkeeper: You could simplify this a bit with Python's built-ins for sorting and filtering. It's probably faster, too. I'm thinking something like:

```
#Only allow plugins that are in the priority order.
accepted_plugins = list(filter(lambda plugin: plugin.canAddManualDevice(address) in priority_order, self._plugins.values()))
if not accepted_plugins:
    Logger.log("d", "Could not find a plugin blablabla")

return max(accepted_plugins, key = lambda plugin: priority_order.index(plugin.canAddManualDevice(address)))
```
(Not tested yet though.)

UM/Qt/Bindings/OutputDeviceManagerProxy.py
```diff
@@ -27,7 +27,10 @@ def __init__(self, parent = None) -> None:
         self._device_manager.activeDeviceChanged.connect(self._onActiveDeviceChanged)
         self._onActiveDeviceChanged()
 
+        self._manualDeviceInfo = {}
```
Ghostkeeper: I'd expect MyPy to complain about this missing type documentation.

UM/OutputDevice/OutputDeviceManager.py
rburema: Used your suggestion to halve the linecount of that part :-)

Not entirely happy about the two calls to `canAddManualDevice` instead of one, but overall it's better (and that function is explicitly there to _quickly_ check if that plugin can add the machine, so its fine).

UM/Qt/Bindings/OutputDeviceManagerProxy.py
rburema: Fixed!

PR #478, opened 2019-04-04
Opt to use OpenGL 2.0 instead of software rendered OpenGL 4.1

fieldOfView:
This PR checks if the best detected OpenGL version isn't using software rendering provided by the OS instead of a GPU renderer. With newer versions of Qt on OSX (and Linuxes ?), requesting a 4.1 context on a GPU that does not support hardware accelerated 4.1 rendering results in a software rendered context which is much slower than a hardware accelerated 2.0 context. This PR detects software rendering OpenGL 4.1 renderers, and falls back to OpenGL 2.0 instead. 

Contributes to https://github.com/Ultimaker/Cura/issues/4505 (CURA-6092)

This will need testing on many systems.

### DIFF
--- a/UM/Qt/QtApplication.py
+++ b/UM/Qt/QtApplication.py
@@ -135,10 +135,9 @@ class QtApplication(QApplication, Application):
             QMessageBox.critical(None, "Failed to probe OpenGL",
                                  "Could not probe OpenGL. This program requires OpenGL 2.0 or higher. Please check your video card drivers.")
             sys.exit(1)
-
-        opengl_version_str = OpenGLContext.versionAsText(major_version, minor_version, profile)
-        Logger.log("d", "Detected most suitable OpenGL context version: %s", opengl_version_str)
-
+        else:
+            opengl_version_str = OpenGLContext.versionAsText(major_version, minor_version, profile)
+            Logger.log("d", "Detected most suitable OpenGL context version: %s", opengl_version_str)
         if not self.getIsHeadLess():
             OpenGLContext.setDefaultFormat(major_version, minor_version, profile = profile)
 
--- a/UM/Qt/QtApplication.py
+++ b/UM/Qt/QtApplication.py
@@ -9,8 +9,8 @@ from typing import Any, cast, Dict, Optional
 
 from PyQt5.QtCore import Qt, QCoreApplication, QEvent, QUrl, pyqtProperty, pyqtSignal, pyqtSlot, QT_VERSION_STR, PYQT_VERSION_STR
 from PyQt5.QtQml import QQmlApplicationEngine, QQmlComponent, QQmlContext, QQmlError
-from PyQt5.QtWidgets import QApplication, QSplashScreen, QMessageBox, QSystemTrayIcon, QOpenGLWidget
-from PyQt5.QtGui import QIcon, QPixmap, QFontMetrics, QSurfaceFormat, QOpenGLContext
+from PyQt5.QtWidgets import QApplication, QSplashScreen, QMessageBox, QSystemTrayIcon
+from PyQt5.QtGui import QIcon, QPixmap, QFontMetrics
 from PyQt5.QtCore import QTimer
 
 from UM.Backend.Backend import Backend #For typing.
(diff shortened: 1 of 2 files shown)


fieldOfView:
I had a choice to "contaminate" QtApplication with some OpenGL, or "contaminate" OpenGLContext with some Qt. I chose the former.

nallath:
You're hive mind is strong here. I was just typing a message as to why you chose to do it like this.

nallath:
Shit. It seems that this isn't a solution we can take :( 

The line `gl = QOpenGLContext.currentContext().versionFunctions()` is causing a crash for me (`ModuleNotFoundError: No module named 'PyQt5._QOpenGLFunctions_4_5_Core'`). I really hope this is just an issue with my setup...

nallath:
I have found a way to keep it in the 'detectBestOpenGLVersion' though, but I can't actually run it due to the crash :(

fieldOfView:
I am now using QOpenGLContext.currentContext().versionFunctions() the same way OpenGL.py does it. Could you try again?

nallath:
Nope, still nothing :( ModuleNotFoundError: No module named 'PyQt5._QOpenGLFunctions_4_5_Core'

I've also tried to use ctx.versionFunctions() and ctx.versionFunctions(gl_profile) , but neither work.

nallath:
What version of PyQt are you using? I'm getting the feeling it has to do with that.

fieldOfView:
I'm using 5.10.

nallath:
Hmm. That must be it then, I'm on 5.11.3

fieldOfView:
Cura uses 5.10 though, right?

nallath:
Yeah, it does. But this was the only version I could get to run with qml debugging. I could downgrade back again, but that's just going to postpone the problem until we do...

fieldOfView:
But at least it would tell us if it is indeed the PyQt version.

Note that I don't understand how it fails. OpenGL.py does exactly the same here: https://github.com/Ultimaker/Uranium/blob/master/UM/View/GL/OpenGL.py#L54

nallath:
Yeah, I'm also at a loss here. I guess it has something to do with the moment when it happens. The openGL object is created on the first render call from the window (as the first time render is called, it creates a qtRenderer, which in turn creates that object).

fieldOfView:
In that case, could you try changing gl_widget.showMinimized() to gl_widget.show(), because then the window actually gets painted.

nallath:
Tried that one already, it didn't matter. I still get the same result

nallath:
Using PyOpenGL does seem to do the trick though. But that would mean that we'd have to add an extra dependency for just this....

fieldOfView:
I think I found the issue:
https://riverbankcomputing.com/pipermail/pyqt/2017-January/038640.html

> PyQt currently only implements 2.0, 2.1 and 4.1Core

Looks like the detection finds the possibility to use 4.5Core on your system, but then PyQt cannot handle that context/version when we try to create it.

I've changed the code to always create the test-context with 4.1 even if something >4.1 is detected.

nallath:
That did the trick, no more crashes.

thopiekar:
And what happens if I run this in a virtual machine (KVM)? There I want software rendering if I have no 3D accelerator. Or is Qt providing its own renderer?

fieldOfView:
The PR only rejects software 4.1 contexts. The application then continues to check if a 2.1 context is available. There is no check if the 2.1 context uses software rendering. So in your VM scenario, you end up with a software rendered gl context. In my testing, a software rendered 2.1 context performs better than a software 4.1 context (especially in Layer View, where the geometry shader does not have to draw in software mode).

thopiekar:
Great! Thank you for your feedback! 😃

fieldOfView:
I will see if I can create a 4.0 build with a backport of this fix this weekend, for affected OSX users to test (and use).

fieldOfView:
Looks like this needs a bit more work when running "frozen" on OSX.

LipuFei:
@fieldOfView I've created a branch from yours and I will create Cura builds with it. Branch: https://github.com/Ultimaker/Uranium/tree/CURA-6092_aldo_opengl_fix_mac

LipuFei:
@fieldOfView I got a crash on my Mac (Mojave 10.14.3) :(
```
Traceback (most recent call last):
  File "<frozen importlib._bootstrap>", line 969, in _find_and_load
  File "<frozen importlib._bootstrap>", line 958, in _find_and_load_unlocked
  File "<frozen importlib._bootstrap>", line 664, in _load_unlocked
  File "<frozen importlib._bootstrap>", line 634, in _load_backward_compatible
  File "/Users/ultimaker/build/env/master/inst/lib/python3.5/site-packages/cx_Freeze/initscripts/__startup__.py", line 12, in <module>
  File "<frozen importlib._bootstrap>", line 969, in _find_and_load
  File "<frozen importlib._bootstrap>", line 958, in _find_and_load_unlocked
  File "<frozen importlib._bootstrap>", line 664, in _load_unlocked
  File "<frozen importlib._bootstrap>", line 634, in _load_backward_compatible
  File "/Users/ultimaker/build/env/master/inst/lib/python3.5/site-packages/cx_Freeze/initscripts/Console.py", line 21, in <module>
  File "<frozen importlib._bootstrap>", line 969, in _find_and_load
  File "<frozen importlib._bootstrap>", line 958, in _find_and_load_unlocked
  File "<frozen importlib._bootstrap>", line 664, in _load_unlocked
  File "<frozen importlib._bootstrap>", line 634, in _load_backward_compatible
  File "/Users/ultimaker/build/master/build/inst/bin/cura_app.py", line 136, in <module>
  File "/Users/ultimaker/build/master/build/inst/lib/python3.5/site-packages/cura/CuraApplication.py", line 716, in run
  File "/Users/ultimaker/build/master/build/inst/lib/python3.5/site-packages/UM/Qt/QtApplication.py", line 291, in run
  File "/Users/ultimaker/build/master/build/inst/lib/python3.5/site-packages/UM/Application.py", line 209, in run
  File "/Users/ultimaker/build/master/build/inst/lib/python3.5/site-packages/cura/CuraApplication.py", line 316, in initialize
  File "/Users/ultimaker/build/master/build/inst/lib/python3.5/site-packages/UM/Qt/QtApplication.py", line 131, in initialize
  File "/Users/ultimaker/build/master/build/inst/lib/python3.5/site-packages/UM/View/GL/OpenGLContext.py", line 135, in detectBestOpenGLVersion
AttributeError: 'NoneType' object has no attribute 'initializeOpenGLFunctions'

```

nallath:
I do have an idea for a band-aid solution to the issue; We can also add a message that if the 4.1 context is selected with a software renderer, it warns the user that this is the case and provides a button / action to force the compatibility mode.

It isn't as elegant as the automatic solution, but it's pretty much guaranteed to work (since it's just adding a message and doesn't need any "fundamental" changes to our detection)

fieldOfView:
@LipuFei, that's what I meant with that it needs some work. Apparently opening the hidden window does not let Qt initialise OpenGL on OSX (it does on Windows).

What GPU does your Mac have? Does it normally have hardware accelerated OpenGL 4.1?

nallath:
We only have macs that have a hardware accelerated 4.1 OpenGL (which is why we didnt find the issue in the first place).

fieldOfView:
I have found a better way to initialize OpenGL to get the renderer, which also works on OSX. 

This PR works properly on my OS X VM, which now correctly creates a (software) 2.0 context instead of a software 4.1 context. Would be nice to know if it still creates a 4.1 context on real hardware that supports it.

LipuFei:
@fieldOfView 👍 This works on my Mac (Mojave 10.14.4).

fieldOfView:
🍾

Reports here are also favorable for macs that are affected by the issue (with 4.1-challenged gpus)
https://github.com/Ultimaker/Cura/issues/4505#issuecomment-482462148

I have not tested this latest iteration on Linux yet.

diegopradogesto:
I checked on macOS as well and it seems to work fine. @nallath checked on Linux and said it was OK.

Thanks for contributing on fixing this issue.

PR #479, opened 2019-04-10
CURA-6447 Add an application flag indicating if we just updated from an older version

LipuFei:


UM/AppVersionFile.py
```diff
@@ -0,0 +1,54 @@
...
+
+    #
+    # Compares the current application version with the version in the given file.
+    # If the file reading fails, or it's an invalid INI file, or it's missing fields, an exception will be raised.
+    # Otherwise, the comparison result will be returned as an integer:
+    #   -1  : If the current version is lower than the version in the file.
+    #    0  : If the current version is the same as the version in the file.
+    #    1  : If the current version is
+    #
+    def compareVersionWithFile(self, filename: str) -> int:
```
nallath: It's a bit easier to seperate the reading and the comparing in different functions. That way we can actually write a test for the comparing without having it to read all different things we want to test from file.

So compareVersionWithFile can stay the same, but move the actual comparing to a protected function (and make use of the change to also add tests).

UM/AppVersionFile.py
```diff
@@ -0,0 +1,54 @@
...
+
+
+class AppVersionFile:
+
+    #
+    # Constructor. The app_name and version provided here should be the information of the current application.
+    #
+    def __init__(self, app_name: str, version: str) -> None:
+        self.app_name = app_name
+        self.version = version
```
nallath: I guess it's better to use the version object here right away (otherwise we can get confusion about in which cases we use str or Version)

PR #480, opened 2019-04-12
Handle 'log once'.

rburema:
Part of CURA-6457.

### DIFF
--- a/UM/Decorators.py
+++ b/UM/Decorators.py
@@ -16,7 +16,7 @@ def deprecated(message, since = "Unknown"): #pylint: disable=bad-whitespace
     def deprecated_decorator(function):
         def deprecated_function(*args, **kwargs):
             warning = "{0} is deprecated (since {1}): {2}".format(function, since, message)
-            Logger.log("w", warning)
+            Logger.log("w_once", warning)
             warnings.warn(warning, DeprecationWarning, stacklevel=2)
             return function(*args, **kwargs)
         return deprecated_function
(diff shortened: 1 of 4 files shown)


Ghostkeeper:
I added a testing deprecation warning around `ContainerRegistry.addContainer` and found that before the console logger is loaded the duplicates are not removed yet. This is a segment at the beginning of my cstdout:

```
[MainThread] UM.Settings.ContainerRegistry.addContainer [390]: Container [empty_material] added.
[MainThread] UM.Decorators.deprecated_function [19]: <function ContainerRegistry.addContainer at 0x7fd1e9da9a60> is deprecated (since 4.1): !!!!!!!!!!!!!!!!!!!!!!Test deprecation!
[MainThread] UM.Settings.ContainerRegistry.addContainer [390]: Container [empty_quality] added.
[MainThread] UM.Decorators.deprecated_function [19]: <function ContainerRegistry.addContainer at 0x7fd1e9da9a60> is deprecated (since 4.1): !!!!!!!!!!!!!!!!!!!!!!Test deprecation!
(... continues a couple of times)
```

This is because the logging plug-ins are not loaded yet at this point, and then it uses the fallback logger which doesn't handle any log level stuff such as `_once`. I think we'll accept this because the logic needs to happen per plug-in because you want the messages to appear once for every logger. It's going to be hard to fix and not really a problem if it's not fixed.

fieldOfView:
I have a fundamental issue with this functionality.  The fact that a log entry is shown multiple times should not be hidden away. It should be taken as a sign that a piece of code is getting executed over and over again, which points to a likely performance issue.

thopiekar:
I fully agree 👍

rburema:
@fieldOfView We won't use 'log once' lightly! The normal way of logging will still (and for the future) be to just push it to the log. `?_once` (where the ? mark means any of the i,w,d,e,c log types) is very usefull for deprecation messages, which will then show only once per location.

thopiekar:
Well, personally I don't care whether such messages appear multiple times. Once the get fixed they will all disappear at once.

Something that would be more important is to keep logs of X days only. I personally don't understand why we keep all the lines forever. For example the engine is probably much noisier then the deprecation messages you mentioned.
However, asked for this management of log files years ago.  🤷‍♂️

nallath:
The deprecation messages are orders of magnitude more than the engine log.

PR #481, opened 2019-04-15
Switch out the implementation of decompose for a faster variant

nallath:
On my machine, this made the decompose step 67% faster. This, in turn, made all scene operations 24.4% faster. Mileage may vary from system to system but I'm pretty sure that this should be better on most systems.

### DIFF
--- a/UM/Math/Matrix.py
+++ b/UM/Math/Matrix.py
@@ -62,6 +62,9 @@ class Matrix:
         # Note that actually using Matrix(self._data) (without the deepcopy) is another factor 3 faster.
         return Matrix(self._data)
 
+    def copy(self) -> "Matrix":
+        return Matrix(self._data)
+
     def __eq__(self, other: object) -> bool:
         if self is other:
             return True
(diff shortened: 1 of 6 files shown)


fieldOfView:
Has this been compared to the decompose function in Trimesh?
https://trimsh.org/trimesh.html#module-trimesh.transformations

fieldOfView:
Never mind, the old version was quite literally what Trimesh does:
https://github.com/mikedh/trimesh/blob/master/trimesh/transformations.py#L797

nallath:
Yup. When looking into the trimesh I figured that out. This triggered me to have a look at the original source of the algorithm (which stated that it wasn't the best optimised code out there). That, in turn, lead me on a merry chase to a version that is a bit more well behaved.

All in all, the speedup isn't that much, but all of these minor things combined should make it feel a bit more smooth (preventing the recreation of the bounding box mesh also knocks off a few % of transformation operations).

Ghostkeeper:
See issue CURA-6466.

PR #486, opened 2019-05-03
Reduce the amount of checks done on property values

nallath:
As we don't actually allow instance containers to hold anything but values, we can quite often just return None if anything asks for a property that isn't value.

The performance tests that I did on it (boot Cura with and without this change. Once it's booted, check how much time the various getProperty calls took) seem to indicate that we can knock off about 500 ms (1.8s for 100187 calls to getRawProperty without the changes and 1.3s for 100778 calls to getRawProperty with the changes).

This should get some very, very extensive testing before we merge it as this touches a very fundamental part of the setting system.


Ghostkeeper:
Developers, see issue CURA-6507.

UM/Settings/ContainerStack.py
```diff
@@ -241,15 +241,20 @@ def getRawProperty(self, key: str, property_name: str, *, context: Optional[Prop
...
             containers = self._containers[start_index:]
-
-        for container in containers:
-            if skip_until_container and container.getId() != skip_until_container:
-                continue #Skip.
-            skip_until_container = None #When we find the container, stop skipping.
-
-            value = container.getProperty(key, property_name, context)
+        if property_name != "value":
+            # Definition containers are the only one that set something for value, so just skip to the def container.
```
LipuFei:
```suggestion
            # Definition containers are the only one that set something for non-value, so just skip to the def container.
```

PR #487, opened 2019-05-08
CURA-6490 Fix setting visibility

LipuFei:
Use stacks instead of a single `DefinitionContainer` to handle setting visibility.

### DIFF
--- a/UM/Settings/Models/SettingDefinitionsModel.py
+++ b/UM/Settings/Models/SettingDefinitionsModel.py
@@ -516,7 +516,7 @@ class SettingDefinitionsModel(QAbstractListModel):
         # the sub-definition.
         # Get all definitions grouped by key recursively from bottom up, so if a same definition key is found in a
         # sub-definition and a base definition, the one from the sub-definition will be used.
-        all_definitions_dict = {}
+        all_definitions_dict = {}  # type: Dict[str, SettingDefinition]
         def createStackList(definition_dict, s):
             if s is None:
                 return
--- a/UM/Settings/Models/SettingDefinitionsModel.py
+++ b/UM/Settings/Models/SettingDefinitionsModel.py
@@ -38,9 +38,6 @@ class SettingDefinitionsModel(QAbstractListModel):
         self._container = None
         self._i18n_catalog = None
 
-        self._root_key = ""
-        self._root = None
-
         self._definition_list = []
         self._row_index_list = []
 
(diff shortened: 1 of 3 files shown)


tests/Settings/TestSettingDefinitionsModel.py
```diff
@@ -54,6 +61,8 @@ def test_getAndSet(data):
     # Ensure that the value got set
     assert getattr(model, data["attribute"]) == data["value"]
 
+    # Check properties via function calls
+
```
Ghostkeeper: You're missing some tests here.

UM/Settings/Models/SettingDefinitionsModel.py
```diff
@@ -505,16 +501,33 @@ def forceUpdate(self) -> None:
...
+        # sub-definition and a base definition, the one from the sub-definition will be used.
+        all_definitions_dict = {}
+        def createStackList(definition_dict, s):
+            if s is None:
+                return
+            createStackList(definition_dict, s.getNextStack())
+            definitions = s.getBottom().findDefinitions()
+            for d in definitions:
+                definition_dict[d.key] = d
+        createStackList(all_definitions_dict, self._stack)
```
Ghostkeeper: Why are you defining this function here and immediately using it once and throwing it away? It would be more clear with just a block of code, in my opinion.

LipuFei: This is a recursion, so a function is needed...

tests/Settings/TestSettingDefinitionsModel.py
LipuFei: That's some dev code. Removed in 448bc041

Ghostkeeper:
This is causing a bug: Settings are no longer shown translated.

UM/Settings/Models/SettingDefinitionsModel.py
```diff
@@ -532,19 +501,33 @@ def forceUpdate(self) -> None:
...
     def _update(self) -> None:
-        if not self._container:
+        if not self._stack:
+            self._definition_dict.clear()
+            self._definition_list.clear()
             return
 
         # Try and find a translation catalog for the definition
-        for file_name in self._container.getInheritedFiles():
+        for file_name in self._stack.getBottom().getInheritedFiles():
```
LipuFei: @Ghostkeeper The translation problem is because of this.

LipuFei: 3ed873d adds fallbacks to translations to fix this problem

UM/i18n.py
```diff
@@ -30,12 +30,15 @@ class i18nCatalog: # [CodeStyle: Ultimaker code style requires classes to start
     #   \param language The language to load. Valid values are language codes or
     #   "default". When "default" is specified, the language to load will be
     #   determined based on the system"s language settings.
+    #   \param fallback_names The fallback names of the catalog to load. If a translation is not found in the primary
+    #   name, it will try to find it in the next fallback.
```
Ghostkeeper: Can you explain to me again what the problem was that necessitates this?

We are here adding a structure to the i18n catalogue that when reading the i18n catalogue code makes no sense as to why it is necessary to have fallback names. We only have one translation active at a time, so why would a translation not be found and where would you look otherwise?

Ghostkeeper: Can you explain to me again what the problem was that necessitates this?

We are here adding a structure to the i18n catalogue that when reading the i18n catalogue code makes no sense as to why it is necessary to have fallback names. We only have one translation active at a time, so why would a translation not be found and where would you look otherwise?

LipuFei: Explained in the JIRA ticket

PR #490, opened 2019-05-21
Support for orthographic view

nallath:
This adds the much awaited orthographic view! 

It requires https://github.com/Ultimaker/Cura/pull/5768 to work 

CURA-5395


LipuFei:
Sorry @nallath , ScaleTool doesn't work well, and RotateTool doesn't work at all in orthographic view :(

PR #491, opened 2019-05-25
WIP: Allowing to regsiter cli options via plugins

thopiekar:
That was working in the past. Have the suspect that the function is just called too early, so newly registered options don't show up.

### DIFF
--- a/UM/Application.py
+++ b/UM/Application.py
@@ -211,11 +211,14 @@ class Application:
     #   \exception NotImplementedError
     def run(self):
         self.addCommandLineOptions()
-        self.parseCliOptions()
         self.initialize()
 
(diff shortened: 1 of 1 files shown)


thopiekar:
I fully agree with all of your points except that '--help' needs to wait
for all plugins to be loaded, so it can show their options, too.
For all the other options, I agree that it shouldn't work like this and it
needs rework.

Thank you for reviewing!

nallath:
@thopiekar This has been open for a while. Do you still want to pick this up, or should this be closed?

thopiekar:
Since there are no many changes made here, you can close it.
But I definitely want to pick this up again, but that's very low on my list.

Issue #496, opened 2019-07-08
Test failure on various architectures


hroncok:
Uranium 4.1.0 fails a test on i686 and possibly other arches:

```
__________________________ TestVector.test_normalize ___________________________
self = <TestVector.TestVector testMethod=test_normalize>
    def test_normalize(self):
        vector = Vector(10, 10, 10)
>       assert vector.normalized().length() == 1
E       AssertionError: assert 0.9999999999999999 == 1
E         -0.9999999999999999
E         +1
tests/Math/TestVector.py:33: AssertionError
```

This seems like it would benefit a [`pytest.approx`](https://docs.pytest.org/en/latest/reference.html#pytest-approx) call. I'll send a PR.

PR #497, opened 2019-07-08
Allow approximate equality in TestVector.test_normalize for smaller arches

hroncok:
Fixes https://github.com/Ultimaker/Uranium/issues/496


Ghostkeeper:
That test still works on my computer, so if this fixes it for lower resolution floats, that's fine with me.

A different test has been failing since the previous commit before you, but that's unrelated.

hroncok:
Thanks.

PR #499, opened 2019-07-22
Set border_node's geometry and material before appending it to paint_node

smartavionics:
Fixes one of the problems discussed in https://github.com/Ultimaker/Cura/issues/6008.

### DIFF
--- a/UM/Qt/Bindings/PointingRectangle.py
+++ b/UM/Qt/Bindings/PointingRectangle.py
@@ -162,8 +162,9 @@ class PointingRectangle(QQuickItem):
 
         if self._border_width > 0:
             if paint_node.childCount() == 0:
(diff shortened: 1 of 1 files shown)


smartavionics:
Just noticed that using Qt 5.13 on Linux, the info box that shows the times and material usage is mangled. The white box is very small and the text doesn't look right either. It's not this PR that has caused the problem because I have just checked a windows build that uses qt 5.10 and this PR is OK. So I guess it's a difference between Qt 5.10 and 5.13 on Linux that's causing the problem?

fieldOfView:
That rendering difference was reported by @thopiekar some time ago (and was the start of the "please use the AppImage" trope)

smartavionics:
Aha, I guess it will need to be looked at sometime if UM ever want to move on from Qt 5.10.

PR #500, opened 2019-07-22
ShaderEffect crashes on Linux with Qt 5.13 so replace with ColorOverlay.

smartavionics:
Replace ShaderEffect with ColorOverlay. As far as I can tell, the visual effect is equivalent. Tested on Linux using Qt 5.10 and Qt 5.13.

Possibly fixes https://github.com/Ultimaker/Cura/issues/6008.

### DIFF
--- a/UM/Qt/qml/UM/RecolorImage.qml
+++ b/UM/Qt/qml/UM/RecolorImage.qml
@@ -2,6 +2,7 @@
 // Uranium is released under the terms of the LGPLv3 or higher.
 
 import QtQuick 2.1
+import QtGraphicalEffects 1.0
 import UM 1.3 as UM
 
 Item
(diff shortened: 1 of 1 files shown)


Ghostkeeper:
Developers, see issue CURA-6678.

smartavionics:
Also tested on Windows using Qt 5.10, OK.

nallath:
Also works fine for my setup (fedora) with Qt 5.11.3

smartavionics:
Thanks for merging that quickly. Is it also worth looking at the other PR (https://github.com/Ultimaker/Uranium/pull/499)? That fixes an assertion failure that occurs when using the debug version of Qt but I don't know if that assertion failure indicates a real problem or can be ignored.

nallath:
I'm running the debug version of qt 5.11.3 and at least there I don't get the assertion. But at the very least I will have a look if those changes break anything for me (and merge it if everything seems okay)

smartavionics:
I got the assertion when testing with 5.13.

nallath:
Since both situations work for me and the second one works on both cases, I'm going to merge your pull request.

PR #503, opened 2019-07-26
Add some camera functions to use in a plugin

diegopradogesto:
Because I'm creating a plugin that uses the camera (https://github.com/diegopradogesto/cura-camera-position), I need some specific functions.

Let me know if it's a good idea to do it this way.


diegopradogesto:
Devs. CURA-6560

UM/Scene/Camera.py
```diff
@@ -105,7 +103,7 @@ def _updatePerspectiveMatrix(self):
         projection_matrix = Matrix()
         if self.isPerspective():
             if view_width != 0 and view_height != 0:
-                projection_matrix.setPerspective(30, view_width / view_height, 1, 500)
+                projection_matrix.setPerspective(30 + self._zoom_factor, view_width / view_height, 1, 500)
```
nallath: The zoom of the perspective camera is done by setting the position and not by changing the projection matrix.

arjendirks:
Hi @diegopradogesto, can you work on this issue, finish it up, and be forever known as a contributor to the Ultimaker Plugin platform?

arjendirks:
Hi, @diegopradogesto I understand the challenge. I do think it is nice for the marketplace but will be mostly benefitted by our marketing department.

@nallath can you take a look at the pull request Diego made and maybe see whether we can fix it ourselves? We would still make Marketing happy with this.

Ghostkeeper:
I've reverted the changes that Nallath was talking about. They were actually changing the perspective of the camera, according to our tests (but maybe our tests didn't create the exact same circumstances as the real scene).

This revert means that a plug-in would now need to make the decision manually: If in perspective mode you need to dolly forwards and backwards. If in orthographic mode you need to actually zoom in. These changes mean that something would also need to be changed in the plug-in that Diegopradogesto made, to correctly zoom when in perspective mode.

PR #504, opened 2019-07-26
Fix snarf and barf regression!

smartavionics:
Fix regression introduced in https://github.com/Ultimaker/Uranium/commit/661bbfabb036c38ae53847203ef256380858da28.


smartavionics:
Fixes https://github.com/Ultimaker/Cura/issues/6065.

PR #506, opened 2019-07-28
Remove recently added noise.

smartavionics:


nallath:
Whoops. Thanks for fixing it.

PR #514, opened 2019-08-19
Refactor PluginError

digitalfrost:


UM/PluginError.py
```diff
@@ -7,29 +7,18 @@
...
 ##  Raised when a plugin could not be found.
-class PluginNotFoundError(Exception):
-    def __init__(self, name):
-        super().__init__()
-        self._name = name
-
+class PluginNotFoundError(PluginError):
     def __str__(self):
-        return "Could not find plugin " + self._name
+        name = self.args[0]     # pylint: disable=unsubscriptable-object
```
nallath: I think you can just use cast (from typing import cast)  to resolve the typing error (eg; cast(str, self.args[0]))

digitalfrost: Using cast(str, self.args[0])

We still get the warning

```
Value 'self.args' is unsubscriptablepylint(unsubscriptable-object)
```
from pylint 

Please refer to update to the pull request for proposed alternative solution

nallath: Ah, I stand corrected then.

PR #518, opened 2019-08-30
Fix tools when dragging over mouseareas

fieldOfView:
This PR fixes the behavior of tools (translate, rotate, scale) when dragging the mouse over QML elements that contain a MouseArea (such as SettingItems in the sidebar, or the object list) during the operation.

Steps to reproduce the issue:
* load a model
* select the model
* drag the red axis, all the way over the settings float/sidebar

Result:
While over parts of the settings floater, the model jumps to the left of the window, until the pointer is dragged off the QML element that contains a MouseArea.

### DIFF
--- a/UM/Qt/QtMouseDevice.py
+++ b/UM/Qt/QtMouseDevice.py
@@ -17,19 +17,19 @@ class QtMouseDevice(InputDevice):
 
     def handleEvent(self, event):
         if event.type() == QEvent.MouseButtonPress:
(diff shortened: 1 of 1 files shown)


fieldOfView:
Might contribute to https://github.com/Ultimaker/Cura/issues/6268

Ghostkeeper:
The fix seems logical to me and fixes the positioning of objects when dragging over the setting text fields. I haven't seen the issue with the selection zones yet. We'll see if it is fixed by this.

PR #519, opened 2019-09-01
Align Face With Build-Plate

rburema:
Note: Instead of just passing the object-id, also show each gl_PrimitiveID in the shader output.
In order to make it still work for old, old OpenGL versions, I had to hide the information in gl_FragColor somehow: The upper bits of the color repressent the object, to lower bits repressent the faces.

_ edit: CURA-6745 _


UM/Math/Color.py
```diff
@@ -70,6 +78,32 @@ def fromARGB(value):
             (value & 0xff000000) >> 24
         )
 
+    @staticmethod
+    def fromARGBLowBits(value):
+        return Color(
+            (value & 0x000f0000) >> 16,
```
LipuFei: I think you can just do `(value >> 16) & 0xf`, so you don't need to calculate where that `f` is in the `hex` value.

plugins/Tools/RotateTool/RotateTool.qml
```diff
@@ -45,6 +45,23 @@ Item
...
 
+    Button
+    {
+        id: alignFaceButton
+
+        anchors.left: layFlatButton.right;
+        anchors.leftMargin: UM.Theme.getSize("default_margin").width;
+
+        text: catalog.i18nc("@action:button","Align selected face with bottom")
+        iconSource: UM.Theme.getIcon("rotate_layflat");  // TODO!
```
LipuFei: There's a `TODO` here.

rburema: Yes, because I don't have the actual logo yet. We should remember to make one before 4.3-final.

diegopradogesto: Maybe @Yian-Lai can help with that.

rburema: I'm not creating one now, because then there is a large chance that it'll stay there forever.

UM/Math/Color.py
rburema: This is congruent with fromARGB, which already existed, which I then also would have to change (it's right above the added functions).

This is just as many operations, and I think you could argue that it reads a bit better (despite the search for the 'right' bite) because it's closer to the intuition of what's going on.

If you still feel I should change it please let me know.

plugins/Tools/RotateTool/RotateTool.qml
rburema: @diegopradogesto Yes, but not _today_ ;-)

PR #520, opened 2019-09-05
Disable select face feature when gl_PrimitiveID not supported by GLSL.

smartavionics:
The new feature to select a face and then rotate the model so that the face is at the bottom relies on a GLSL feature (gl_PrimitiveID) that isn't in versions of GLSL < 1.5. Therefore, for those early versions of GLSL, the select face feature is hidden.

### DIFF
--- a/UM/View/GL/OpenGL.py
+++ b/UM/View/GL/OpenGL.py
@@ -98,6 +98,11 @@ class OpenGL:
 
         self._opengl_version = self._gl.glGetString(self._gl.GL_VERSION) #type: str
 
+        try:
+            self._opengl_shading_language_version = float(self._gl.glGetString(self._gl.GL_SHADING_LANGUAGE_VERSION))
+        except:
+            self._opengl_shading_language_version = 1.0
+
         if not self.hasFrameBufferObjects():
             Logger.log("w", "No frame buffer support, falling back to texture copies.")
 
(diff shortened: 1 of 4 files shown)


smartavionics:
You may think that no platforms would use GLSL older than 1.5 but, unfortunately, the Raspberry Pi currently only supports version 1.2! It may be that other systems that people are using with the compatibility mode enabled would also require this feature to be disabled?

Ghostkeeper:
We're still messing around with this a bit. We might add the necessary information to the vertices instead of this solution, so that the drop to build plate still works. Not sure yet.

rburema:
My bad, I thought `gl_PrimitiveID` was a built-in from the start (or at least there for longer), I really should've looked it up beforehand.

In any case, since we're able to load models with index-buffers, putting the necessary information within a vertex attribute isn't possible, since vertices can share more than one face in that scenario. I've tried to come up with ways around this, but it just becomes more and more hacky (at least if we want to keep the performance).

I think I might use this pull as a basis for further work.

smartavionics:
It's probably worth keeping the code for querying the GLSL version even if you don't use the select face specific changes.

rburema:
> It's probably worth keeping the code for querying the GLSL version even if you don't use the select face specific changes.

I agree, and it's one of the reasons I'll probably base the fixes/completion planned for this feature on this branch.

rburema:
See also:
 - https://github.com/Ultimaker/Uranium/pull/521
 - https://github.com/Ultimaker/Cura/pull/6328

PR #521, opened 2019-09-10
CURA-6754 / CURA-6761 Fix Select Face

rburema:
Wraps up development for the 'lay flat by face' feature, at least concerning the 4.3-BETA (we'll probably revisit it in the near future).

(See also the PR in Cura, https://github.com/Ultimaker/Cura/pull/6328)

### DIFF
--- a/UM/View/GL/OpenGL.py
+++ b/UM/View/GL/OpenGL.py
@@ -98,6 +98,11 @@ class OpenGL:
 
         self._opengl_version = self._gl.glGetString(self._gl.GL_VERSION) #type: str
 
+        try:
+            self._opengl_shading_language_version = float(self._gl.glGetString(self._gl.GL_SHADING_LANGUAGE_VERSION))
+        except:
+            self._opengl_shading_language_version = 1.0
+
         if not self.hasFrameBufferObjects():
             Logger.log("w", "No frame buffer support, falling back to texture copies.")
 
@@ -105,6 +110,7 @@ class OpenGL:
         Logger.log("d", "OpenGL Version:  %s", self._opengl_version)
         Logger.log("d", "OpenGL Vendor:   %s", self._gl.glGetString(self._gl.GL_VENDOR))
         Logger.log("d", "OpenGL Renderer: %s", self._gpu_type)
+        Logger.log("d", "GLSL Version:    %f", self._opengl_shading_language_version)
 
     ##  Check if the current OpenGL implementation supports FrameBuffer Objects.
     #
@@ -117,6 +123,12 @@ class OpenGL:
     #   \return Version of OpenGL
     def getOpenGLVersion(self) -> str:
         return self._opengl_version
+
+    ##  Get the current OpenGL shading language version.
+    #
+    #   \return Shading language version of OpenGL
+    def getOpenGLShadingLanguageVersion(self) -> float:
+        return self._opengl_shading_language_version
 
     ##  Get the current GPU vendor name.
     #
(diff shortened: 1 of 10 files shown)


UM/View/GL/OpenGL.py
```diff
@@ -98,13 +98,19 @@ def __init__(self) -> None:
 
         self._opengl_version = self._gl.glGetString(self._gl.GL_VERSION) #type: str
 
+        try:
+            self._opengl_shading_language_version = float(self._gl.glGetString(self._gl.GL_SHADING_LANGUAGE_VERSION)[0:3])
+        except:
+            self._opengl_shading_language_version = 1.0
```
LipuFei: I would use this as `UM.Version` object which you can use for version comparison. If later the version includes one more part, it won't work with `float`.

UM/View/GL/OpenGL.py
```diff
@@ -118,6 +125,12 @@ def hasFrameBufferObjects(self) -> bool:
     def getOpenGLVersion(self) -> str:
         return self._opengl_version
 
+    ##  Get the current OpenGL shading language version.
+    #
+    #   \return Shading language version of OpenGL
+    def getOpenGLShadingLanguageVersion(self) -> float:
```
LipuFei: Same here, better to use `UM.Version`.

UM/View/SelectionPass.py
```diff
@@ -36,22 +42,34 @@ def __init__(self, width, height):
...
 
+    def _onSelectedFaceChanged(self):
+        self._mode = SelectionPass.SelectionMode.FACES if Selection.getFaceSelectMode() else SelectionPass.SelectionMode.OBJECTS
+
     ##  Perform the actual rendering.
     def render(self):
+        if self._mode == SelectionPass.SelectionMode.OBJECTS:
+            self._renderObjectsMode()
+        elif self._mode == SelectionPass.SelectionMode.FACES:
+            self._renderFacesMode()
```
LipuFei: I'd add an entry for `else` to log as a warning in case the mode doesn't match any of the above.

UM/View/SelectionPass.py
```diff
@@ -80,6 +98,24 @@ def render(self):
 
         self.release()
 
+    def _renderFacesMode(self):
+        batch = RenderBatch(self._face_shader)
+
+        selectable_objects = False
+        for node in Selection.getAllSelectedObjects():
+            if isinstance(node, ToolHandle):
```
LipuFei: I think `node.isSelectable() and node.getMeshData()` this check should be enough. I don't know, but I don't think `ToolHandle` is selectable.

plugins/Tools/RotateTool/RotateTool.py
```diff
@@ -179,12 +183,52 @@ def event(self, event):
...
+    def _onSelectedFaceChanged(self):
+        self._handle.setEnabled(not Selection.getFaceSelectMode())
+
+        selected_face = Selection.getSelectedFace()
+        if not Selection.getSelectedFace() or not (Selection.hasSelection() and Selection.getFaceSelectMode()):
+            return
+
+        original_node, face_id = selected_face
+        meshdata = original_node.getMeshDataTransformed()
+        if not meshdata or face_id < 0:
```
LipuFei: Does it also need to check if `face_id` is larger than the max face ID?

plugins/Tools/SelectionTool/SelectionTool.py
```diff
@@ -60,6 +60,8 @@ def event(self, event):
             self._selection_pass = self._renderer.getRenderPass("selection")
 
         self.checkModifierKeys(event)
+        #if event.type == MouseEvent.MouseMoveEvent and Selection.getFaceSelectMode():
```
LipuFei: Remove those 2 lines?

PR #523, opened 2019-09-20
Add "restore window geometry" preference

fieldOfView:
This PR adds a preference around restoring the previous window position/size to the last used position/size on start up. This would be a workaround for those setups where starting Cura on a secondary screen will prevent it from working. 

Contributes to https://github.com/Ultimaker/Cura/issues/5651#issuecomment-488048392 and the rest of these issues: https://github.com/ultimaker/cura/issues?utf8=%E2%9C%93&q=is%3Aissue+window_left+is%3Aopen+

There will be a companion PR for Cura to add a checkbox for the preference.


nallath:
This one got snowed under, sowwy.

CURA-6901

PR #524, opened 2019-09-21
fix glsl lies

rburema:
See https://github.com/Ultimaker/Cura/issues/6387

### DIFF
--- a/plugins/Tools/RotateTool/RotateTool.py
+++ b/plugins/Tools/RotateTool/RotateTool.py
@@ -230,7 +230,8 @@ class RotateTool(Tool):
     #
     #   \return type(Boolean)
     def getSelectFaceSupported(self):
-        return OpenGL.getInstance().getOpenGLShadingLanguageVersion() >= Version("1.50")
+        # Use a dummy postfix, since an equal version with a postfix is considered smaller normally.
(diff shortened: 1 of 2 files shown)


plugins/Tools/RotateTool/RotateTool.py
```diff
@@ -230,7 +230,8 @@ def getToolHint(self):
     #
     #   \return type(Boolean)
     def getSelectFaceSupported(self):
-        return OpenGL.getInstance().getOpenGLShadingLanguageVersion() >= Version("1.50")
+        # Use a dummy postfix, since an equal version with a postfix is considered smaller normally.
+        return Version(OpenGL.getInstance().getOpenGLVersion()) >= Version("4.1 dummy-postfix")
```
Ghostkeeper: Hold on, it seems that this function is not used anywhere in the code?

rburema:
Thanks for picking this up! Sorry about the pushiness.

> Hold on, it seems that this function is not used anywhere in the code?

The align-face-button is only visible when:
`        visible: UM.ActiveTool.properties.getValue("SelectFaceSupported");`
which doesn't show up in any search functions as you've noticed.

nallath:
No need to apologize, it needed to get done before the 4.3 :)

PR #528, opened 2019-09-29
UM.Controller: Adding variable for tool fallback

thopiekar:
Need to clear all tools on my scene and add my own "toolset". However, in this toolset there is no TranslateTool and I would like to fall back to a different tool.
Of course, it would be convenient to have also get* and set* functions, but I don't care about these right now.

### DIFF
--- a/UM/Controller.py
+++ b/UM/Controller.py
@@ -195,6 +195,17 @@ class Controller:
             self._input_devices[name].event.disconnect(self.event)
             del self._input_devices[name]
 
+    ##  Request the current fallbacl tool.
+    #   \return Id of the fallback tool
+    def getFallbackTool(self) -> str:
+        return self._fallback_tool
+
+    ##  Set the current active tool. The tool must be set by name.
+    #   \param tool The tools name which shall be used as fallback
+    def setFallbackTool(self, tool: str):
+        if self._fallback_tool is not tool:
+            self._fallback_tool = tool
+
     ##  Request tool by name. Returns None if no view is found.
     #   \param name \type{string} Unique identifier of tool (usually the plugin name)
     #   \return tool \type{Tool} if name was found, None otherwise.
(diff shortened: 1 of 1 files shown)


UM/Controller.py
```diff
@@ -265,7 +266,7 @@ def setActiveTool(self, tool: Optional[Union["Tool", str]]):
         from UM.Scene.Selection import Selection  # Imported here to prevent a circular dependency.
         if not self._active_tool and Selection.getCount() > 0:  # If something is selected, a tool must always be active.
             if "TranslateTool" in self._tools:
-                self._active_tool = self._tools["TranslateTool"]  # Then default to the translation tool.
+                self._active_tool = self._tools[self._fallback_tool]  # Then default to the translation tool.
```
fieldOfView: Shouldn't you also change the `if "TranslateTool" in self._tools:` above?

thopiekar: Great, thanks. Overlooked this one :smile:

thopiekar:
@Ghostkeeper @nallath Is there someone available to give it a review? :angel:

fieldOfView:
Looks like there is no way to set this _fallback_tool, and it is marked as private.

nallath:
If you don't care about changing the fallback tool, then why bother changing it? (I do agree it's a bit cleaner, but mostly because it's less hardcoded)

thopiekar:
Good point. Just added the missing functions.

UM/Controller.py
```diff
@@ -194,7 +195,18 @@ def removeInputDevice(self, name: str) -> None:
...
 
-    ##  Request tool by name. Returns None if no view is found.
+    ##  Request the current fallbacl tool.
+    #   \return Id of the fallback tool
+    def getFallbackTool(self) -> str:
+        return self._fallback_tool
+
+    ##  Set the current active tool. The tool must be set by name.
+    #   \param tool The tools name which shall be used as fallback
+    def setFallbackTool(self, tool: str):
```
nallath:
```suggestion
    def setFallbackTool(self, tool: str) -> None:
```
It needs to have full typing (so also indicate that nothing is returned)

thopiekar:
@nallath Any chance to get this merged soon? :wink:

PR #529, opened 2019-10-02
RotateTool: Check whether tool is enabled

thopiekar:
Probably the original code is doing the same check, but I believe that we want to check here simply whether the tool is enabled or not.
If that is the case (which I hope), then this might be better readable at least.

### DIFF
--- a/plugins/Tools/RotateTool/RotateTool.py
+++ b/plugins/Tools/RotateTool/RotateTool.py
@@ -75,7 +75,7 @@ class RotateTool(Tool):
             # Snap is "toggled back" when releasing the shift button
             self.setRotationSnap(not self._snap_rotation)
 
-        if event.type == Event.MousePressEvent and self._controller.getToolsEnabled():
+        if event.type == Event.MousePressEvent and self.getEnabled():
             # Start a rotate operation
             if MouseEvent.LeftButton not in event.buttons:
                 return False
--- a/plugins/Tools/RotateTool/RotateTool.py
+++ b/plugins/Tools/RotateTool/RotateTool.py
@@ -75,7 +75,7 @@ class RotateTool(Tool):
             # Snap is "toggled back" when releasing the shift button
             self.setRotationSnap(not self._snap_rotation)
 
-        if event.type == Event.MousePressEvent and self.getEnabled():
+        if event.type == Event.MousePressEvent and self._controller.getToolsEnabled() and self.getEnabled():
             # Start a rotate operation
             if MouseEvent.LeftButton not in event.buttons:
                 return False


fieldOfView:
Actually the original code checks if *all* tools are enabled. During some actions, *all* tools get disabled, and the rotate action should not be initiated when they are.

thopiekar:
Haven't tried, but I believe that this can fix the bug I see here.
Within the tool plugin here, I activate the face selection tool and for some reason it begins to rotate.
This time I skipped using any keybindings. Just simple clicks.

thopiekar:
@fieldOfView By the way, what is the definition of an "enabled" tool? Is a tool enabled, when you grab a handle? Or make a drag?

nallath:
Some tools can be conditionally enabled. Take a look at the PerObjectSetting tool. Since we only want to have it active if a single object is selected, we set it do disabled if more objects are selected.

thopiekar:
Ok, so per definition (as I can't find a clear statement in the sources).
 * What is a active tool? One that is currently in use (whose handle is visible)? Or one that is just selected in the toolbar?
 * Same for an enabled tool.

I guess the main problem is that I can't follow what you mean with enabled and active. (Among other things which seem to happen in the back of Cura around tools.)

nallath:
Enabled implies that the tool can / is allowed to do it's job. (eg; Can it be selected / active *at all*)
Active implies that it's the tool that is selected by the user.

Ghostkeeper:
This PR highlights a bit of a design flaw in my opinion. You can see this happening in Tool.py:
* The `getEnabled` function returns `self._enabled`.
* In the `__init__`, a signal is linked from `self._controller._onToolEnabledChanged` to `self._onToolEnabledChanged`.
* The `self._onToolEnabledChanged` function simply keeps `self._enabled` up to date.
* These are the only usages of `self._enabled` in the entire code base, Cura or Uranium. Nothing is overriding that behaviour.

So as far as I'm concerned, we have double data here and the `getEnabled` function should simply ask the controller whether it is enabled so we don't need that signal and there is no chance of these things getting out of sync.

Either way, I believe that this PR is correct regarding separation of concern. Just ask yourself whether you're enabled or not, and let Uranium's implementation of the tool handle the rest.

Ghostkeeper:
Ah, no, I was wrong. Controller doesn't keep track of which tools are enabled but only passes on the signal. That's also a bit weird in the design there.

So the getToolsEnabled is completely separate from whether the plug-in itself is enabled or not. AFAIK the getToolsEnabled determines whether the toolbar gets shown or not (e.g. it's gone when you have no models on the build plate). So technically it'd need to check both!

fieldOfView:
> AFAIK the getToolsEnabled determines whether the toolbar gets shown or not

No, it is whether the toolbar is enabled or not. The toolbar gets disabled during operations (eg when dragging a model).

Whether the toolbar gets shown or not purely depends on if the is "platform activity" (which means there's a model loaded).

Ghostkeeper:
Ah, that makes sense. But then the rotate tool still needs to be rotating even if `getToolsEnabled` is false, right?

fieldOfView:
Yes, but it should not *start* rotating, it should only continue rotating if it is already rotating.

plugins/Tools/RotateTool/RotateTool.py
```diff
@@ -75,7 +75,7 @@ def event(self, event):
             # Snap is "toggled back" when releasing the shift button
             self.setRotationSnap(not self._snap_rotation)
 
-        if event.type == Event.MousePressEvent and self._controller.getToolsEnabled():
+        if event.type == Event.MousePressEvent and self.getEnabled():
```
Ghostkeeper:
```suggestion
        if event.type == Event.MousePressEvent and self._controller.getToolsEnabled() and self.getEnabled():
```

thopiekar:
I'm glad to see that this conversation leaded to an improvement :smile:

thopiekar:
Ok, please forget about this PR. It really does no change to the problem I actually have.
Now I noticed the real reason why the rotation is being done.

Whenever I switch to face selection and I click on a surface, then the "layflat" procedure is done.
It doesn't matter, whether I have the RotationTool enabled or not - the feature is still enabled.

So my wild guess is, that there are signals connected to the layflat function still and these are only checking for the selection mode.

Could you, @nallath, be so kind and take a look into that?

Thanks!

rburema:
Hi @thopiekar 
Would this PR #543 help?
cheers, Remco.

thopiekar:
I'm on it: https://code.launchpad.net/~thopiekar/+recipe/uranium-master-daily

thopiekar:
@rburema The patch is working perfectly fine!!!

rburema:
Good to hear :-)

thopiekar:
Is it possible to get Linux and Windows builds of it?

rburema:
It's not merged yet, so we'll need to start a build manually. How soon do you need this?

thopiekar:
Well, the sooner the better. Additionally, I noticed an unusual behavior that when I click multiple times on a model, I'm not able to select it anymore. It keeps being deselected except I use "select all" from the right click menu.
I'm sure you are testing Cura good, but double check for this symptoms maybe.

rburema:
@thopiekar  I think I got the buildsystem to (eventually) give me individual builds of this branch without hecking up the nightlies. Could take a few hours. Remind me tomorrow if I don't get back to you.

rburema:
Check your mail :-)

PR #531, opened 2019-10-09
Controller: Adding setAllTools function

thopiekar:
Basically adds the missing piece for the get* command.
Apart from that I need it for a project.

### DIFF
--- a/UM/Controller.py
+++ b/UM/Controller.py
@@ -205,9 +205,14 @@ class Controller:
             return None
 
     ##  Get all tools
-    #   \return tools \type{dict}
+    #   \return tools \type{dict} All tools collected in a dictionary via tool name and it's instance
     def getAllTools(self) -> Dict[str, "Tool"]:
         return self._tools
+
+    ##  Set all tools
+    #   \param tools \type{dict} All tools collected in a dictionary via tool name and it's instance
+    def setAllTools(self, tools) -> None:
+        self._tools = tools
 
     ##  Add a Tool (transform object, translate object) if its not already added.
     #   \param tool \type{Tool} Tool to be added
--- a/UM/Controller.py
+++ b/UM/Controller.py
@@ -211,7 +211,7 @@ class Controller:
 
     ##  Set all tools
     #   \param tools \type{dict} All tools collected in a dictionary via tool name and it's instance
-    def setAllTools(self, tools) -> None:
+    def setAllTools(self, tools: dict) -> None:
         self._tools = tools
 
     ##  Add a Tool (transform object, translate object) if its not already added.


UM/Controller.py
```diff
@@ -205,10 +205,15 @@ def getTool(self, name: str) -> Optional["Tool"]:
...
 
     ##  Get all tools
-    #   \return tools \type{dict}
+    #   \return tools \type{dict} All tools collected in a dictionary via tool name and it's instance
     def getAllTools(self) -> Dict[str, "Tool"]:
         return self._tools
 
+    ##  Set all tools
+    #   \param tools \type{dict} All tools collected in a dictionary via tool name and it's instance
+    def setAllTools(self, tools) -> None:
```
ChrisTerBeke: The argument `tools` is missing typing.

UM/Controller.py
```diff
@@ -205,10 +205,15 @@ def getTool(self, name: str) -> Optional["Tool"]:
...
 
     ##  Get all tools
-    #   \return tools \type{dict}
+    #   \return tools \type{dict} All tools collected in a dictionary via tool name and it's instance
     def getAllTools(self) -> Dict[str, "Tool"]:
         return self._tools
 
+    ##  Set all tools
+    #   \param tools \type{dict} All tools collected in a dictionary via tool name and it's instance
+    def setAllTools(self, tools: dict) -> None:
```
ChrisTerBeke: Although now typed, I believe the correct typing is `Dict[str, Tool]`.

nallath:
```suggestion
    def setAllTools(self, tools: Dict[str, "Tool"]) -> None:
```

nallath:
Isn't it just easier to use the tool visibility for this? Eg; Say that you want to have a number of tools that are _only_ active when a certain view / stage is active.
You ensure that the default visibility of those tools is false (so they won't ever show up in the other stages / views / whatever). Once the stage /view that should have them visible activates, it stores the current visibility state of the tools, and sets the desired state (eg; Rotate tool invisible, new tool to visible). Once the stage gets de-activated again, you reset the visibility to the previously stored state.

UM/Controller.py
thopiekar: Oh, ok. So it would allow different kind of contents inside the dict.
But if we clearly define it as @nallath did for me, then it is clear for the typing that it must be a set of str : "Tool", right?

thopiekar:
@nallath Yes, it is an alternative. (I just didn't know it is there.)
However, the work is the same. As soon as the QML engine is loaded, I still need to run my loop and make all my tools invisible and every time I enter and exit my stage I need to apply the visibility of each tool still.
Sure it is cleaner, but work-wise it is the same.

So yes, I can implement it as you suggested, but as we all put already effort into this, why not merging it? Maybe it finds in another situation a useful usecase :)

What do you think?

fieldOfView:
> What do you think?

It is not how things work in Uranium (or Cura). There is no setter for a list of available profiles, there is no setter for a list of enabled settings, there is no setter for a list of plugins. All these things are loaded by a registry, and there are methods to enable/disable and/or hide/show them.

>  as we all put already effort into this, why not merging it? Maybe it finds in another situation a useful usecase :)

Paraphrasing Archer: Do you want coderot? Because that's how you get coderot.

As to your problem: you can also implement your own toolbar. Or subclass the Controller for your plugin. So many options...

nallath:
Because actually setting/removing them doesn't ensure that other signals that may or may not be connected to be disconnected. 

Apart from that, it's also a semantical difference; You don't want to remove the tools, you want to make them disabled/invisible for a certain moment (eg; They are still there, just not active). Actually changing them implies a lot more responsibility.

thopiekar:
@nallath So if I understand correctly, I should disable the tool as a plugin, right?
If so, then I'll need to do it via PluginRegistry and probably not by the plugin itself:
image
There is also no .setEnabled() function by the way.

nallath:
If you look at the ToolModel, you can see that it ignores tools that are set as invisible (aka; Those will not be shown in the toolbar)

fieldOfView:
FYI: Plugins cannot be disabled/reenabled "on the fly"; disabling a plugin requires a restart of Cura, like how it is done on the "Installed" tab of the Marketplace.

thopiekar:
@nallath Thanks! Got a blacklist with the cameratool and the selectiontool, so they won't get invisible and especially not visible. Another our_tools variable, which has a list of all tool ids which shall be explicitly shown. This is what I ended up with:
```
for tool_id in Application.getInstance().getController().getAllTools().keys():
            if tool_id not in self._tool_blacklist:
                visible = tool_id in self._our_toolset
                PluginRegistry.getInstance().getMetaData(tool_id).get("tool", {})["visible"] = visible
                Application.getInstance().getController().toolsChanged.emit()
```

Works good as far as I can see, but can imagine that I still miss something :smile:
Any concerns?

@fieldOfView I believed that all plugins can be disabled and enabled at any time and only whenever there is trouble shutting down a plugin a restart is required. Thank you for the hint!

PR #532, opened 2019-10-16
Performance improvements

Ghostkeeper:
These are performance improvements for Uranium applications w.r.t. the query system for the container registry.
* Cache intermediary results of queries.
* Increase cache size.

This belongs to this PR on Cura: https://github.com/Ultimaker/Cura/pull/6542

Implements CURA-6793.

### DIFF
--- a/UM/Settings/ContainerQuery.py
+++ b/UM/Settings/ContainerQuery.py
@@ -9,9 +9,6 @@ import functools
 if TYPE_CHECKING:
     from UM.Settings.ContainerRegistry import ContainerRegistry
 
-# The maximum amount of query results we should cache
-MaxQueryCacheSize = 10000
-
 ##  Wrapper class to perform a search for a certain set of containers.
 #
 #   This class is primarily intended to be used internally by
@@ -21,7 +18,7 @@ MaxQueryCacheSize = 10000
 #   \note Instances of this class will ignore the query results when
 #   comparing. This is done to simplify the caching code in ContainerRegistry.
 class ContainerQuery:
-    cache = collections.OrderedDict()  # type: collections.OrderedDict # To speed things up, we're keeping a cache of the container queries we've executed before.
+    cache = {}  # type: Dict[Tuple[Any, ...], ContainerQuery]  # To speed things up, we're keeping a cache of the container queries we've executed before.
 
     # If a field is provided in the format "[t1|t2|t3|...]", try to find if any of the given tokens is present in the
     # value. Use regex to do matching because certain fields such as name can be filled by a user and it can be string
@@ -77,7 +74,6 @@ class ContainerQuery:
             key_so_far += (key, value)
             if candidates is None and key_so_far in self.cache:
                 filtered_candidates = cast(List[Dict[str, Any]], self.cache[key_so_far].getResult())
-                self.cache.move_to_end(key_so_far)  # The cache entry was used, so make sure it doesn't fall off as an infrequently-used cache entry.
                 continue
 
             if isinstance(value, type):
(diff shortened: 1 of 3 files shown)


UM/Settings/ContainerQuery.py
```diff
@@ -1,14 +1,17 @@
...
-from typing import Any, cast, Dict, Iterable, List, Optional, Type, TYPE_CHECKING
-from UM.Logger import Logger
+from typing import Any, cast, Dict, List, Optional, Tuple, Type, TYPE_CHECKING
 import functools
 
 if TYPE_CHECKING:
     from UM.Settings.ContainerRegistry import ContainerRegistry
 
+# The maximum amount of query results we should cache
+MaxQueryCacheSize = 10000
```
nallath: Will setting it at 10k ever trigger the mem cleanup (if it doesn't, we might as well drop it all together, right?)

UM/Settings/ContainerQuery.py
```diff
@@ -56,50 +60,50 @@ def getResult(self) -> Optional[List[Dict[str, Any]]]:
...
-        self._result = list(filtered_candidates)
+                key_filter = functools.partial(self._matchDirect, property_name = key, value = value)
+            filtered_candidates = list(filter(key_filter, filtered_candidates))  # Execute this filter.
 
-    def __hash__(self) -> int:
-        return hash(self.__key())
+            if candidates is None:  # Only cache if we didn't pre-filter candidates.
+                cached_arguments = dict(zip(key_so_far[1::2], key_so_far[2::2]))
+                self.cache[key_so_far] = ContainerQuery(self._registry, ignore_case = self._ignore_case, **cached_arguments)  # Cache this query for the next time.
+                self.cache[key_so_far]._result = filtered_candidates
```
nallath: Accessing a protected member.

UM/Settings/ContainerQuery.py
```diff
@@ -56,50 +60,50 @@ def getResult(self) -> Optional[List[Dict[str, Any]]]:
...
-                return False
-        return True
-
     ##  Execute the actual query.
     #
     #   This will search the container metadata of the ContainerRegistry based
     #   on the arguments provided to this class' constructor. After it is done,
     #   the result can be retrieved with getResult().
     def execute(self, candidates: Optional[List[Any]] = None) -> None:
+        key_so_far = (self._ignore_case, )  # type: Tuple[Any, ...]
```
nallath: This little nugget needs a bit more documentation as to what it does and how it makes things better.

UM/Settings/ContainerQuery.py
nallath: The container query didn't give consent to its privates being touched.

UM/Settings/ContainerQuery.py
Ghostkeeper: I've added all of Ultimaker's printers and all of Creality's printers to initiate many, many queries and it didn't hit this limit. I'll remove the limit then.

UM/Settings/ContainerQuery.py
Ghostkeeper: This is allowed since it's a protected member of an instance of ContainerQuery. At least, C++ allows this.

UM/Settings/ContainerQuery.py
Ghostkeeper: Done here: e8a1a32017e031a8cffea91dc53695a32d59942e

UM/Settings/ContainerQuery.py
Ghostkeeper: Similar to above: Normally this is allowed since it's the same class. It's allowed in Java and C++ at least.

PR #534, opened 2019-10-21
Change SDK version to 7.0.0

ninovanhooff:
CURA-6858


Ghostkeeper:
When the SDK version is incremented with a major increment, deprecated functions need to be removed as well.

PR #535, opened 2019-10-22
Fix normals after mirroring.

rburema:
A mirror operaion was done via 'scale', but that inverts the normals, which might give some issues (for example, 'Make Overhang Printable' wouldn't work correctly).

part of CURA-6924

### DIFF
--- a/UM/Operations/MirrorOperation.py
+++ b/UM/Operations/MirrorOperation.py
@@ -28,6 +28,7 @@ class MirrorOperation(Operation.Operation):
 
     ##  Undo the operation.
     def undo(self):
+        self._node.invertNormals()
         self._node.setTransformation(self._old_transformation) #We stored the old transformation, so we can just restore that.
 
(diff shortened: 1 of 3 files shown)


UM/Operations/MirrorOperation.py
```diff
@@ -36,6 +36,7 @@ def redo(self):
             center = self._node.getPosition()
             self._node.setPosition(-center)
         self._node.scale(self._mirror, SceneNode.TransformSpace.World) #Then mirror around the origin.
+        self._node.invertNormals()  # Because the mirror is done via a scale operation, the normals where inverted.
```
LipuFei: @rburema I guess `invertNormals()` should also happen in `undo()`?

LipuFei: Added in 98d2380b

rburema: Yes, thanks.

PR #536, opened 2019-10-25
CURA-6557 Add CMake options to exclude plugins for installation

LipuFei:


cmake/UraniumPluginInstall.cmake
```diff
@@ -0,0 +1,98 @@
...
+#   - plugins/my_plugin/plugin.json
+#   - plugins/my_plugin/my_module/plugin.json
+# In this case, only "plugins/my_plugin" should be added via install().
+set(_no_install_plugin_list ${UM_NO_INSTALL_PLUGINS})
+# Sanitize the string so the comparison will be case-insensitive.
+string(STRIP   "${_no_install_plugin_list}" _no_install_plugin_list)
+string(TOLOWER "${_no_install_plugin_list}" _no_install_plugin_list)
+
+# WORKAROUND counterpart of what's in cura-build.
+string(REPLACE "," ";" _no_install_plugin_list "${_no_install_plugin_list}")
```
ninovanhooff: I'd say that if this is an essentials build, the UpdateChecker is implied to be in this list, right? Maybe add it here to prevent user mistakes. If that is too difficult, update the description for the essentials checkbox in Jenkins to say that UpdateChecker must be added to the list

LipuFei: This is not specific for any type of build. This is something generic. You can provide a list of plugins to exclude in a build via the CMake variable `UM_NO_INSTALL_PLUGINS` and that's it.

LipuFei:
The CI checks are fixed in master. I just did a rebase so the commits here will also pass.

cmake/mod_bundled_packages_json.py
```diff
@@ -0,0 +1,71 @@
...
+    """
+    json_file_list = []
+    for root, dir_names, file_names in os.walk(work_dir):
+        for file_name in file_names:
+            abs_path = os.path.abspath(os.path.join(root, file_name))
+            json_file_list.append(abs_path)
+    return json_file_list
+
+
+def remove_entries_from_json_file(file_path: str, entries: list) -> None:
```
nallath:
```suggestion
def remove_entries_from_json_file(file_path: str, entries: List[str]) -> None:
```
The documentation states that it's a list of strings, so might as well make the typing agree to that, right?

LipuFei: This is done on purpose. `List` requires the `typing` package. Because this is kinda a standalone script, I made it this way so it doesn't have extra dependencies.

Ghostkeeper: But the `typing` package is required to be built-in in Python since 3.5, so the same could be said for the `os` package or the `argparse` package.

PR #537, opened 2019-10-25
Switch the updates of the settingPropertyProvider to _updateDelayed

nallath:
I dusted off the good 'ol qml profiler to make things a bit faster.

Linked with https://github.com/Ultimaker/Cura/pull/6583

### DIFF
--- a/UM/Settings/Models/SettingPropertyProvider.py
+++ b/UM/Settings/Models/SettingPropertyProvider.py
@@ -65,7 +65,7 @@ class SettingPropertyProvider(QObject):
             self._stack.containersChanged.connect(self._containersChanged)
 
         self._validator = None
-        self._update()
+        self._updateDelayed()
         self.containerStackChanged.emit()
 
     ##  Set the containerStackId property.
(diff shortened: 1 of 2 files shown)


Ghostkeeper:
This change is causing a new QML warning to appear:

    2019-10-28 16:08:06,373 - WARNING - [MainThread] UM.Qt.QtApplication.__onQmlWarning [369]: file:///home/ruben/Projects/Cura/resources/qml/Settings/SettingExtruder.qml:78:9: Unable to assign [undefined] to int

This refers to the following line of code:

https://github.com/Ultimaker/Cura/blob/b360e0db3943fe1bb9d95e35c41e3630c7a06e40/resources/qml/Settings/SettingExtruder.qml#L78

This warning is output for all SettingExtruder and SettingCheckbox items every time you open the setting category (and a bunch of times when starting up). It tells me two things:
* Apparently during creation the value of a property is returned as `None`.
* Since the displayed extruder is still correct, the UI is apparently updated again when the 500ms timer expires. Or at least the Javascript is executed again (because I don't see it switch; maybe it's defaulting to 0 and that is correct).

nallath:
The warning is fixed in https://github.com/Ultimaker/Cura/commit/007add7fc2dbf70e7a0070d731bbe43a481f52a3

PR #538, opened 2019-10-30
Cache the results of the _findPlugin function

nallath:
When loading a single plugin, the function was called twice (once for actually loading it, once for populating the metadata). Since that is pretty wasteful, we can just cache the results the first time. This shouldn't cause any issues since plugin_id's are supposed to be unique.


nallath:
Closing in favor of #548

PR #539, opened 2019-11-01
Packages API speed improvements

ChrisTerBeke:
Goes with https://github.com/Ultimaker/Cura/pull/6615.

It would be great to have this in 4.4, but since we're in feature freeze I can understand why you'd want this on master instead.


ChrisTerBeke:
Closing in favor of https://github.com/Ultimaker/Uranium/pull/540.

PR #540, opened 2019-11-01
Packages API speed improvements

ChrisTerBeke:
Goes with https://github.com/Ultimaker/Cura/pull/6615.

It would be great to have this in 4.4, but since we're in feature freeze I can understand why you'd want this on master instead.

### DIFF
--- a/UM/PackageManager.py
+++ b/UM/PackageManager.py
@@ -92,6 +92,11 @@ class PackageManager(QObject):
     def packagesWithUpdate(self) -> Set[str]:
         return self._packages_with_update_available
 
+    ## Alternative way of setting the available package updates without having to check all packages in the cloud.
+    def setPackagesWithUpdate(self, packages: Set[str]):
+        self._packages_with_update_available = packages
+        self.packagesWithUpdateChanged.emit()
+
     def checkIfPackageCanUpdate(self, package_id: str) -> bool:
         available_versions = self._available_package_versions.get(package_id)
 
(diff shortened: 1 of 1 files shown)


UM/PackageManager.py
```diff
@@ -291,6 +296,24 @@ def getAllInstalledPackageIDs(self) -> Set[str]:
 
         return all_installed_ids
 
+    ## Get a list of tuples that contain the package ID and version.
+    #  Used by the Marketplace to check which packages have updates available.
+    def getAllInstalledPackageIdsAndVersions(self) -> List[Tuple[str, str]]:
+        package_ids_and_versions = []  # type: List[Tuple[str, str]]
+        all_installed_ids = self.getAllInstalledPackageIDs()
+        for package_id in all_installed_ids:
+            if package_id in self._bundled_package_dict:
```
nallath: A bundled package can be updated, so this check is not correct.

UM/PackageManager.py
```diff
@@ -291,6 +296,24 @@ def getAllInstalledPackageIDs(self) -> Set[str]:
...
+                continue
+            if package_id in self._application.getRequiredPlugins():
```
nallath: A required plugin is bundled by definition. So this check will never do anything

UM/PackageManager.py
ChrisTerBeke: I'll remove this check then.

UM/PackageManager.py
ChrisTerBeke: I'll remove this check then.

PR #542, opened 2019-11-04
Multiple tools per plugin

ninovanhooff:


UM/Controller.py
```diff
@@ -59,7 +59,7 @@ def __init__(self, application: "Application") -> None:
     #   \param name \type{string} Unique identifier of view (usually the plugin name)
     #   \param view \type{View} The view to be added
     def addView(self, view: View) -> None:
-        name = view.getPluginId()
+        name = view.getName()
```
ninovanhooff: name feels like display name to me, while "guid" might be more suitable here

UM/Controller.py
```diff
@@ -59,7 +59,7 @@ def __init__(self, application: "Application") -> None:
     #   \param name \type{string} Unique identifier of view (usually the plugin name)
```
ninovanhooff: Not a parameter

UM/Controller.py
```diff
@@ -116,7 +116,7 @@ def setActiveView(self, name: str) -> None:
     #   \param name \type{string} Unique identifier of stage (usually the plugin name)
```
ninovanhooff: Not a parameter

UM/Controller.py
fieldOfView: ... or id, since it is not globally unique.

nallath: It's kinda globally unique, since it's being constructed by combinding the plugin_id and the name. 
A guid is a bit annoying, since you do want to be able to use it as a predictable identifier in some other code (eg; setActiveView).

fieldOfView: A guid is a specifically formatted "thing". Don't call something that is not a guid a guid. Guid is just not applicable here. Other similar "kinda unique" properties to use for identification are called "id" in Cura. So why not this thing? A "name" in Cura is always the user-facing "pretty" label.

nallath: I've renamed it to be "just" id (so getId()). I guess that is different enough from getPluginId (which gives the id of the plugin the tool belongs to)

PR #543, opened 2019-11-14
Decouple direct connection from face selection to specific use.

rburema:


plugins/Tools/RotateTool/RotateTool.py
```diff
@@ -56,9 +56,10 @@ def __init__(self):
         self._iterations = 0
         self._total_iterations = 0
         self._rotating = False
-        self.setExposedProperties("ToolHint", "RotationSnap", "RotationSnapAngle", "SelectFaceSupported")
+        self.setExposedProperties("ToolHint", "RotationSnap", "RotationSnapAngle", "SelectFaceSupported", "SelectFaceToLayFlatMode")
         self._saved_node_positions = []
 
+        self._selectFaceMode = False
```
nallath: Code style; should be select_face_mode

plugins/Tools/RotateTool/RotateTool.py
```diff
@@ -264,6 +268,23 @@ def setRotationSnapAngle(self, angle):
             self._snap_angle = angle
             self.propertyChanged.emit()
 
+    ##  Wether the rotate tool is in 'Lay flat by face'-Mode.
+    #
+    #   \return (bool)
+    def getSelectFaceToLayFlatMode(self):
```
nallath: Missing mypy typing.

plugins/Tools/RotateTool/RotateTool.py
```diff
@@ -264,6 +268,23 @@ def setRotationSnapAngle(self, angle):
...
+        if not Selection.getFaceSelectMode():
+            self._selectFaceMode = False  # .. but not the other way around!
+        return self._selectFaceMode
+
+    ##  Set the rotate tool to/from 'Lay flat by face'-Mode.
+    #
+    #   \param (bool)
+    def setSelectFaceToLayFlatMode(self, select):
```
nallath: Missing mypy typing

rburema: Then we should do the typing of all the tools :-)

plugins/Tools/RotateTool/RotateTool.py
rburema: Then we should do the typing of all the tools :-)

plugins/Tools/RotateTool/RotateTool.py
rburema: Oops... will fix that right away.

plugins/Tools/RotateTool/RotateTool.py
nallath: Oh, I figured this would fall under the boy scouting rule.

rburema: Yes, you're right! Ok, will do. It just always looks so lopsided to me when _the entire rest of the file_ is not typed :'-D

nallath: Well, you could also do the rest ofcourse (and be a _really_ good boyscout)

rburema: ... the entire class proved a bit much...

plugins/Tools/RotateTool/RotateTool.py
Ghostkeeper: We've previously said that we'd add this typing boy-scout-style. When you edit something close by, add typing.

rburema: I think I remember* trying this, and then it gave more and more warnings (which would've tripped up the CI as well). So I added more typing, repeat. Eventually I had to conclude that it went way out of scope of the changes.

*) I'm not entirely sure it was for _this_ PR though.

PR #548, opened 2019-11-25
Speed up loading of plugins on start

fieldOfView:
This PR speeds up the loading of plugins on start. The original code contains recursion which I think nobody really understood and the folder cache variable did not work. Because of this, each successive loaded plugin would take longer to be located.

On my system, the "Loading plugins" stage took 47 seconds, which has now been reduced to 27 seconds. The original code was especially cumbersome because it traversed every .git and pycache folder for each plugin load, which on my system caused the _locate_plugin method to be call hundreds of thousands of time on launch.

Edit: this PR now includes the optimization in #538

### DIFF
--- a/UM/PluginRegistry.py
+++ b/UM/PluginRegistry.py
@@ -604,22 +604,11 @@ class PluginRegistry(QObject):
             return None
 
         if folder not in self._folder_cache:
-            sub_folders = []
-            for file in os.listdir(folder):
-                file_path = os.path.join(folder, file)
-                if os.path.isdir(file_path):
-                    entry = (file, file_path)
-                    sub_folders.append(entry)
-            self._folder_cache[folder] = sub_folders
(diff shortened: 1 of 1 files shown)


nallath:
I attempted to remedy this slightly different. Did you also have a look at https://github.com/Ultimaker/Uranium/pull/538?

fieldOfView:
Your code still uses _locatePlugin, which is basically broken (it builds a recursive cache, which it cannot use). I'll test with our optimisations combined.

fieldOfView:
Some profiling later, this is how long the plugin discovery and loading takes on my system:

* current master: 21.5 sec
* your fixes: 8.66 sec
* my fixes: 7.78 sec
* combined fixes: 7.69 sec

fieldOfView:
* latest optmisation: 7.44 sec

nallath:
I don't have that dramatic of an effect (from 1.9 sec to 1.4 sec) but it's still significantly faster.

PR #556, opened 2019-12-24
CURA-6823_improve_splash_screen_responsiveness

ninovanhooff:
Add some intermediate loading messages and make sure the messages are shown at the right moment


fieldOfView:
So now we have a “Loading interface” and a “Building UI”. How about “Loading theme” or “Setting up theme” instead?

nallath:
The loading of the theme takes 0.1 sec. We could add a message for that, but I don't think it will be readable.

PR #560, opened 2019-12-31
Only re-render the scene when it's actually needed

nallath:
Previously we would also re-render the entire scene if a QML object required some re-drawing. We can save a fair bit of time if we don't do this.

CURA-6247


rburema:
It now doesn't immediately change the scene on selecting other view-types in the simulation-view (like line-type -> material color or something), you always need to move first.

I've made a fix locally (to Cura, not Uranium) that I can commit, but I'm merging this first as it's a really small issue.

(In SimulationView.py in the Cura repo:)
~~~
    def setSimulationViewType(self, layer_view_type: int) -> None:
        if layer_view_type != self._layer_view_type:
            self._layer_view_type = layer_view_type
            self.currentLayerNumChanged.emit()

            scene = Application.getInstance().getController().getScene()
            scene.sceneChanged.emit(scene.getRoot())
~~~

PR #561, opened 2020-01-06
Fix restore window size if application is closed maximized

fieldOfView:
This PR fixes restoring the window to its "normal" size and location after closing it in maximized state. In addition to that, it also excludes the maximised/normal state of the window when the `restore_window_geometry` preference is unset.

Fixes https://github.com/Ultimaker/cura/issues/6904

Steps to reproduce issue that is fixed by this PR:
* load Cura and set the window to a custom size and position
* make sure the "Restore window position on start" option is checked
* restart Cura
-> windows is correctly placed in the custom position and size

* now maximize the window
* restart Cura
-> window is (correctly) maximized on startup

* now restore the window to not maximized
-> window (incorrectly!) does not restore to previous custom position and size

### DIFF
--- a/UM/Qt/Bindings/MainWindow.py
+++ b/UM/Qt/Bindings/MainWindow.py
@@ -239,11 +239,14 @@ class MainWindow(QQuickWindow):
 
     @pyqtSlot()
     def _onWindowGeometryChanged(self):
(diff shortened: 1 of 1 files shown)


Ghostkeeper:
Yup, works fine!

PR #565, opened 2020-01-19
CURA-6670 Fix windows/opengl/intel-drivers hang

rburema:
The most important thing for the fix itself is the 'keep in memory' thing, but we might want the other stuff just in case.

### DIFF
--- a/UM/Qt/QtApplication.py
+++ b/UM/Qt/QtApplication.py
@@ -134,7 +134,9 @@ class QtApplication(QApplication, Application):
         self.setStyle("fusion")
 
         self.setAttribute(Qt.AA_UseDesktopOpenGL)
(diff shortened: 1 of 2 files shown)


fieldOfView:
Now we have two contradicting preferences; `force_modern_opengl` and `force_legacy_opengl`. It "feels" like this could have been done more elegantly.

rburema:
@fieldOfView Yeah, that's because they weren't really designed as such, more like debugging aids that I decided to keep in because they might be useful someday in a similar situation. I'll make a note of it and try to amend that sometime this/next week or so in between the other stuff. Alternatively, feel free to fix it yourself of course :-)

rburema:
Hi @fieldOfView I've made the preferences (which are hidden from the user and only 'manually accessible' from the .cfg file) less confusing: see PR #573

PR #567, opened 2020-01-20
CURA-6627_Store_extra_data_project_files

ninovanhooff:
Created on behalf of Nallath

https://github.com/Ultimaker/Cura/pull/7005

### DIFF
--- a/UM/FileHandler/FileHandler.py
+++ b/UM/FileHandler/FileHandler.py
@@ -125,7 +125,7 @@ class FileHandler(QObject):
     # \param file_name The name of the file to load.
     # \param kwargs Keyword arguments.
     # \returns None if nothing was found
-    def readerRead(self, reader: FileReader, file_name: str, **kwargs: Any):
+    def readerRead(self, reader, file_name: str, **kwargs: Any):
         raise NotImplementedError("readerRead must be implemented by subclasses.")
 
     ##  Get a mesh writer object that supports writing the specified mime type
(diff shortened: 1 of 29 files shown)


UM/FileHandler/FileHandler.py
```diff
@@ -125,7 +125,7 @@ def addWriter(self, writer: "FileWriter") -> None:
     # \param file_name The name of the file to load.
     # \param kwargs Keyword arguments.
     # \returns None if nothing was found
-    def readerRead(self, reader: FileReader, file_name: str, **kwargs: Any):
+    def readerRead(self, reader, file_name: str, **kwargs: Any):
```
ninovanhooff: Why no typing?

nallath: The reason for this is that both the MeshFileHandlers and the WorkspaceFileHandlers inherit from this. But at that point, the WorkspaceFileHandler can't override the signature to accept a WorkspaceFileReader only, since it's not fully compatible.

This basicly left me with two options; Either remove this typing, or re-work the entire structure that it's in right now. After a lot of doubting, I decided to just remove the typing here. That being said, I'm open to suggestions as to how this can be resolved differently.

Ghostkeeper:
Hmm, seems my latest commit is causing an internal error in MyPy...?

rburema:
> Hmm, seems my latest commit is causing an internal error in MyPy...?

I thought `dict` is the structure itself, not the type and you should use `Dict` (not capital) when typing things?

PR #568, opened 2020-01-21
CURA-7090/Added function reEvaluateDismissedPackages() in the Package Manager

DonExo:
Added two functions to the Package Manager:
1. A function that marks a package as "dismissed" and puts it in a list called `dismissed` in the `packages.json` file in the user's local data folder.
2. A function that re-evaluates the dismissed incompatible packages, and if needed removes them from the dismissed_packages list.

Is being used in: https://github.com/Ultimaker/Cura/pull/7011

### DIFF
--- a/UM/PackageManager.py
+++ b/UM/PackageManager.py
@@ -345,17 +345,13 @@ class PackageManager(QObject):
     def getToRemovePackageIDs(self) -> Set[str]:
         return self._to_remove_package_set
 
-    def dismissPackage(self, package_id: str) -> None:
-        self._dismissed_packages.add(package_id)
+    def dismissAllIncompatiblePackages(self, incompatible_packages: List[str]) -> None:
+        self._dismissed_packages.update(incompatible_packages)
         self._saveManagementData()
+        Logger.debug("Dismissed Incompatible package(s): {}".format(incompatible_packages))
 
     def getDismissedPackages(self) -> List[str]:
         return list(self._dismissed_packages)
-
-    def removeFromDismissedPackages(self, package: str) -> None:
-        if package in self._dismissed_packages:
-            self._dismissed_packages.remove(package)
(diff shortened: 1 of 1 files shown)


UM/PackageManager.py
```diff
@@ -345,9 +345,26 @@ def getAllInstalledPackagesInfo(self) -> Dict[str, List[Dict[str, Any]]]:
...
+
     def getDismissedPackages(self) -> List[str]:
         return list(self._dismissed_packages)
 
+    def removeFromDismissedPackages(self, package: str) -> None:
+        if package in self._dismissed_packages:
+            self._dismissed_packages.remove(package)
+            Logger.debug("Removed package [%s] from the dismissed packages list" % package)
+
+    def reEvaluateDismissedPackages(self, data) -> None:
```
konskarm: Add type of "data"

UM/PackageManager.py
```diff
@@ -345,9 +345,26 @@ def getAllInstalledPackagesInfo(self) -> Dict[str, List[Dict[str, Any]]]:
...
+        dismissed_packages = self.getDismissedPackages()
+        for dismissed in dismissed_packages:
+            for item in data:
+                if item['package_id'] == dismissed:
```
konskarm:
```suggestion
                if item["package_id"] == dismissed:
```

UM/PackageManager.py
```diff
@@ -11,6 +11,9 @@
 
 from PyQt5.QtCore import pyqtSlot, QObject, pyqtSignal, QUrl, pyqtProperty
 
+from cura import ApplicationMetadata
```
ninovanhooff: Cura shouldn't be used from Uranium, right? Are you still working on this?

DonExo: That was a mistake, it has been deleted.
The `sdk_version` is now passed from the calling cura function.

PR #570, opened 2020-01-22
Do not call the finished callback when a HTTP request errors

ninovanhooff:
As a result of a request, either callback (success case) or
error_callback (failure case) should be called. Not both


LipuFei:
@ninovanhooff Looks good now. Feel free to merge if the tests pass.

PR #573, opened 2020-01-30
Make hidden OpenGL detect prefs. less confusing.

rburema:
During the windows bug fix, some hidden preferences where created that we decided to leave in. This makes them less ad-hoc.


fieldOfView:
👍 

Perhaps the `OpenGlVersionDetect` class could live in `OpenGLContext`?

Ghostkeeper:
Works well. Thanks!

PR #575, opened 2020-02-03
Cura 7024 duplicated machines version upgrade

nallath:


UM/VersionUpgradeManager.py
```diff
@@ -142,22 +144,22 @@ def registerCurrentVersion(self, version_info: Tuple[str, int], type_info: Any)
     #   The upgrade plug-ins must all be loaded at this point, or no upgrades
     #   can be performed.
     #
-    #   \return True if anything was upgraded, or False if it was already up to
-    #   date.
-    def upgrade(self) -> bool:
+    def upgrade(self):
```
nallath:
```suggestion
    def upgrade(self) -> None:
```
Mypy doesn't check the function otherwise.

UM/VersionUpgradeManager.py
```diff
@@ -176,7 +178,10 @@ def upgrade(self) -> bool:
     #   storage path.
     #   \param configuration_type The file type of the specified file.
     def upgradeExtraFile(self, storage_path: str, file_name: str, configuration_type: str) -> None:
-        self._upgrade_tasks.append(UpgradeTask(storage_path = storage_path, file_name = file_name, configuration_type = configuration_type))
+        self._upgrade_tasks.append(FileTask(storage_path = storage_path, file_name = file_name, configuration_type = configuration_type))
+
+    def upgradeExtraTask(self, task: CallableTask):
```
nallath:
```suggestion
    def upgradeExtraTask(self, task: CallableTask) -> None:
```

UM/VersionUpgradeManager.py
```diff
@@ -278,7 +283,7 @@ def _getFilesInDirectory(self, directory: str) -> Iterator[str]:
     ##  Gets all files that need to be upgraded.
     #
     #   \return A sequence of UpgradeTasks of files to upgrade.
-    def _getUpgradeTasks(self) -> Iterator[UpgradeTask]:
+    def _getUpgradeTasks(self) -> Iterator[FileTask]:
```
nallath: I think this entire function should be renamed. It's not getting Upgrade Tasks anymore after all. (Also, the documentation is outdated)

PR #576, opened 2020-02-04
Fixes simulation sliders yet again.

rburema:
If a mesh is drawn with a range, make sure the current range is applicable to that mesh. (But don't reset the range to the max of that mesh.)
What this does is fix the OpenGL crash that you can get when the max is over the index-buffer length -- but without messing up the sim. sliders.

Wait it turns out it _does_ have an internal ticket ... my bad: CURA-7161


rburema:
Closing this in favour of #577 (the same but then to 4.5 instead of master).

PR #580, opened 2020-02-19
Updated pt_BR strings for Cura 4.5

Patola:


Ghostkeeper:
Thanks for the continued support!

PR #584, opened 2020-02-28
CURA-7016 Fix Trust and tests thereof.

rburema:
Adding a public key to Cura revealed that fail fast crashes weren't happening when an untrusted plugin was about to be loaded. This fixes that. (Here of course since plugins are in Uranium as swell.)

(See also https://github.com/Ultimaker/Cura/pull/7182 )

### DIFF
--- a/UM/Trust.py
+++ b/UM/Trust.py
@@ -4,6 +4,7 @@
 import base64
 import json
 import os
+import sys
 from typing import Callable, Optional, Tuple
 
 # Note that we unfortunately need to use 'hazmat' code, as there apparently is no way to do what we want otherwise.
@@ -38,13 +39,10 @@ class TrustBasics:
 
     @staticmethod
     def defaultViolationHandler(message: str) -> None:
-        """Any (real) trust violation in the life code should go through this method in the end.
-
-        :return: The hash-algorithm used for the entire 'suite'.
-        """
+        """This violationHandler is called after any other handlers"""
 
         Logger.logException("e", message)
-        raise SystemExit(0x69)
+        sys.exit("Trust violation")
 
     @classmethod
     def getHashAlgorithm(cls):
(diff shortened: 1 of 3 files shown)


UM/Trust.py
```diff
@@ -36,6 +36,16 @@ class TrustBasics:
     __signature_filename_extension = ".signature"
     __root_signature_entry = "root_signature"
 
+    @staticmethod
```
ninovanhooff: This entire method confuses me. Probably the docstring is outdated. 

Why raise a SystemExit with an obscure hex arg instead of using sys.exit()? Will it be caught to allow some graceful cleanup / shutdown logic?

nallath: Should it even break as hard as it's doing right now? If a single file is not trusted, we shouldn't shut down the entire application (Because that would mean that if one file gets corrupted, the entire application doesn't start!). It would take a single plugin to be placed in the local folder to prevent someome from using cura.

This is also why our unit tests are failing; They simply expect that the function returns that the file is not trusted and that the program just continues from that point onwards.

tests/TestTrust.py
```diff
@@ -52,52 +53,70 @@ def test_signFileAndVerify(self, init_trust):
...
         trust_instance._public_key = None
         assert not trust_instance.signedFileCheck(filepath_signed)
-        assert violation_callback.call_count > 0
+        assert violation_callback.call_count == 3
         violation_callback.reset_mock()
         trust_instance._public_key = public_key
 
+        # Oh noes! Someone changed the file!
         with open(filepath_signed, "w") as file:
             file.write("\nPay 10 Golden Talents To Get Your Data Back Or Else\n")
```
ninovanhooff: What are golden talents and where do I get them? Need those nature pics restored ASAP

rburema: https://en.wikipedia.org/wiki/Talent_(measurement)

UM/Trust.py
rburema: The idea was to use another handler when Trust is used in another way (like for material _verification_ instead of _security_). Since we're not going to crash anymore at all the point is a bit moot.

rburema: Also for the unit-tests, so the tests wouldn't crash while running.

tests/TestTrust.py
rburema: https://en.wikipedia.org/wiki/Talent_(measurement)

rburema:
Love the extra attention to detail peeps! You've given this an extra bit of shine :-)

PR #585, opened 2020-02-28
Remember the last used output device between sessions

fieldOfView:
This PR adds a preference to the OutputDeviceManager to remember which outputdevice is selected between sessions.

https://github.com/Ultimaker/Cura/issues/7168

### DIFF
--- a/UM/Application.py
+++ b/UM/Application.py
@@ -181,7 +181,7 @@ class Application:
         self._preferences.addPreference("general/disabled_plugins", "")
 
         self._controller = Controller(self)
-        self._output_device_manager = OutputDeviceManager(self)
+        self._output_device_manager = OutputDeviceManager()
 
         self._operation_stack = OperationStack(self._controller)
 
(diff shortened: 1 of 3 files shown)


fieldOfView:
`TestOutputDeviceManager` will have to be fixed.

I'm not sure if there is a better way to access the Application object (which is needed to reach the Preferences object and to delay restoring the active output device until after the output device plugins have loaded). If there is, the signature of OutputDevice() would not have to change.

fieldOfView:
Hmm, simple as this PR is, it needs some more work; UM.Application does not have a signal I can use to wait until all outputdeviceplugins have loaded, and we don't want to tie the OutputDeviceManager to UM.QTApplication.

Perhaps this should live in the OutputDeviceManagerProxy, which is already tied to Qt.

fieldOfView:
Changed to a much cleaner implementation in `OutputDeviceManagerProxy`

rburema:
Devs: Refer to internal ticket CURA-7335 for this PR.

PR #587, opened 2020-03-03
Prevent code injection in setting functions

nallath:
This makes the setting function a whole lot more paranoid! It contains both a white and a black list. All setting calls must be on the white list and a number of strings are blacklisted. 

CURA-7169

### DIFF
--- a/UM/Settings/SettingFunction.py
+++ b/UM/Settings/SettingFunction.py
@@ -230,6 +230,8 @@ class _SettingExpressionVisitor(ast.NodeVisitor):
         # extra sure that certain calls are *not* made!)
         if node.s in self._blacklist:
             raise IllegalMethodError(node.s)
+        if node.s.startswith("_"):
+            raise IllegalMethodError(node.s)
 
         if node.s not in self._knownNames and node.s not in dir(builtins):  # type: ignore #AST uses getattr stuff, so ignore type of node.s.
             self.keys.add(node.s)  # type: ignore
@@ -240,6 +242,8 @@ class _SettingExpressionVisitor(ast.NodeVisitor):
         # extra sure that certain calls are *not* made!)
         if node.value in self._blacklist:
             raise IllegalMethodError(node.value)
+        if node.s.startswith("_"):
+            raise IllegalMethodError(node.s)
         if isinstance(node.value, str) and node.value not in self._knownNames and node.value not in dir(builtins):  # type: ignore #AST uses getattr stuff, so ignore type of node.value.
             self.keys.add(node.value)  # type: ignore
 
--- a/UM/Settings/SettingFunction.py
+++ b/UM/Settings/SettingFunction.py
@@ -3,16 +3,12 @@
 
 import ast
 import builtins  # To check against functions that are built-in in Python.
-import math  # Imported here so it can be used easily by the setting functions.
-import uuid  # Imported here so it can be used easily by the setting functions.
-import base64  # Imported here so it can be used easily by the setting functions.
-import hashlib  # Imported here so it can be used easily by the setting functions.
 from types import CodeType
 from typing import Any, Callable, Dict, FrozenSet, NamedTuple, Optional, Set, TYPE_CHECKING
 
+from UM.Logger import Logger
 from UM.Settings.Interfaces import ContainerInterface
 from UM.Settings.PropertyEvaluationContext import PropertyEvaluationContext
-from UM.Logger import Logger
 
 if TYPE_CHECKING:
     from typing import FrozenSet
@@ -28,10 +24,10 @@ def _debug_value(value: Any) -> Any:
 
 
 class SettingFunction:
-    """
-    This class is used to evaluate Python codes (or you can call them formulas) for a setting's property. If a setting's
-    property is a static type, e.g., a string, an int, a float, etc., its value will just be interpreted as it is, but
-    when it's a Python code (formula), the value needs to be evaluated via this class.
+    """Evaluates Python formulas for a setting's property.
+
+    If a setting's property is a static type, e.g. a string, an int, a float, etc., its value will just be interpreted
+    as it is, but when it's a Python code (formula), the value needs to be evaluated via this class.
     """
     def __init__(self, expression: str) -> None:
         """Constructor.
--- a/UM/Settings/SettingFunction.py
+++ b/UM/Settings/SettingFunction.py
@@ -268,7 +268,8 @@ class _SettingExpressionVisitor(ast.NodeVisitor):
         "encode",
         "b64encode",
         "digest",
-        "md5"
+        "md5",
+        "radians"
     }  # type: Set[str]
 
     _blacklist = {
--- a/UM/Settings/SettingFunction.py
+++ b/UM/Settings/SettingFunction.py
@@ -316,3 +316,27 @@ class _SettingExpressionVisitor(ast.NodeVisitor):
         "__pycache__",
         "__file__"
     }  # type: Set[str]
+
+    _allowed_builtins = {
+        "bool",
+        "True",
+        "False",
+        "None",
+        "float",
+        "int",
+        "str",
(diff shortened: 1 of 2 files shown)


UM/Settings/SettingFunction.py
```diff
@@ -194,39 +194,88 @@ def visit_Name(self, node: ast.Name) -> None:  # [CodeStyle: ast.NodeVisitor req
...
+
     def visit_Str(self, node: ast.AST) -> None:
         """This one is used before Python 3.8 to visit string types.
         
         visit_Str will be marked as deprecated from Python 3.8 and onwards.
         """
+        # The blacklisting is done just in case (All function calls should be whitelisted. The blacklist is to make
+        # extra sure that certain calls are *not* made!)
+        if node.s in self._blacklist:
+            raise IllegalMethodError(node.s)
```
BagelOrb: You are blacklisting only constant values.

Shouldn't you also blacklist similarly in `visit_Call` ?
It seems to me we should check all of the possible [classnames](See https://docs.python.org/3/library/ast.html#node-classes) and override as many as possible using `visit_<classname>`.

i.e. ExtSlice, SetComp, Starred, etc?

BagelOrb: GeneratorExp?

UM/Settings/SettingFunction.py
```diff
@@ -194,39 +194,153 @@ def visit_Name(self, node: ast.Name) -> None:  # [CodeStyle: ast.NodeVisitor req
...
+        "floor"
         "debug",
         "sum",
         "len",
         "uuid",
         "hashlib",
-        "base64"
+        "base64",
+        "uuid3",
+        "NAMESPACE_DNS",
```
BagelOrb: Why is this one needed?

UM/Settings/SettingFunction.py
```diff
@@ -194,39 +194,153 @@ def visit_Name(self, node: ast.Name) -> None:  # [CodeStyle: ast.NodeVisitor req
...
+        "lambda",
+        "__getattribute__",
+        "__setattr__",
+        "find_module",
+        "__pycache__",
+        "__file__"
     }  # type: Set[str]
+
+    _allowed_builtins = {
+        "bool",
```
BagelOrb: How about tuple?
How about sorted?
Hwo about range?
I could imagine use cases for those in setting functions.

UM/Settings/SettingFunction.py
```diff
@@ -194,39 +194,153 @@ def visit_Name(self, node: ast.Name) -> None:  # [CodeStyle: ast.NodeVisitor req
             self.values.add(node.id)
             self.keys.add(node.id)
```
BagelOrb: Please document these.

fieldOfView:
All of a sudden I have an urge to import python modules in setting functions. Quickly, now it is still possible!

UM/Settings/SettingFunction.py
```diff
@@ -194,39 +194,88 @@ def visit_Name(self, node: ast.Name) -> None:  # [CodeStyle: ast.NodeVisitor req
...
         "os",
+        "delattr"
+        "getattr"
+        "dir"
+        "open",
+        "write",
+        "compile",
         "import",
         "__import__",
+        "_",  # Because you can use this guy to create a number of the other blacklisted strings
```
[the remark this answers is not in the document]
nallath: You could, yeah. But it does make it a lot harder to use.

UM/Settings/SettingFunction.py
```diff
@@ -194,39 +194,88 @@ def visit_Name(self, node: ast.Name) -> None:  # [CodeStyle: ast.NodeVisitor req
...
+    def visit_ListComp(self, node):
+        # This is mostly because there is no reason to use it at the moment. It might also be an attack vector.
+        raise IllegalMethodError("List Comprehension is not allowed")
+
+    def visit_Attribute(self, node: ast.Attribute) -> None:
+        if node.attr not in self._knownNames:
+            raise IllegalMethodError(node.attr)
+        self.visit(node.value)  # Go a step deeper
+
+    def generic_visit(self, node):
```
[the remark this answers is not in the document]
nallath: It is not. If this method is not overriden, the others don't get visited.

UM/Settings/SettingFunction.py
nallath: Because of the PPA

UM/Settings/SettingFunction.py
nallath: I'm not quite sure what there is to document.

BagelOrb: Please document the fields `keys` and `values`. What are these concepts referring to?

Just copy the description from the class into a comment on the actual fields.

BagelOrb: I don't understand the difference. What setting keys are used by the expression, but not needed to execute the function?! Or the other way around..?

UM/Settings/SettingFunction.py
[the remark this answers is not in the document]
BagelOrb: That's very strange.
Then what does this method of the base class do?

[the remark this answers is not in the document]
BagelOrb: Also: can you document why to override this function and move this function to either above the list of all the visit_<classname> functions?

UM/Settings/SettingFunction.py
nallath: You can't blacklist call like that, because it doesn't have an ID to check for.

tests/Settings/TestSettingFunction.py
```diff
@@ -40,6 +40,10 @@ def setting_function_good(request):
     "(",                                                                # Syntax error.
     "[x for x in (1).__class__.__base__.__subclasses__() if x.__name__ == '_ImportLockContext'][0]().__enter__.__globals__['__builtins__']['__import__']('os').system(\"echo omgzomg\")",  # Obsfucated call to system
     "[x for x in (1).__class__.__base__.__subclasses__() if x.__name__ == '_ImportLockContext'][0]().__enter__.__globals__['__' + 'builtins__']['__import__']('os').system(\"echo omgzomg\")"   #Another obsfucated call to system
+    "'_'", # This string is not allowed
+    "import sys",  # We don't allow importing
+    "'a_a'[1]",  # Trying to circumvent the protection against underscores
+    "{x:x for x in [1,2,3,4]}"  # We don't allow dict comprehension
```
fieldOfView: Why don't we allow dict comprehension? Seems arbitrary to disallow that.

nallath: One of the hacks from the pen testing guys used list comprehensions to get past our defences. You can use the same trick with dict comprehensions. The main reason for it is that you can use it to filter results by using the "if" option of a comprehension.

fieldOfView:
Will string manipulations such as `.upper()` and `.lower()`, `.startswith()` and `endswidth()` and the use of `in` (eg `if "test" in list_value`) still be allowed?

nallath:
`"if "test" in list_value"` -> Should work
As for upper, lower, startswith and endswith, they are not in the whitelist, so they don't work. That being said, I don't expect that there will be much risk in adding those. Do you have a specifc reason for wanting them?

fieldOfView:
> "if "test" in list_value" -> Should work

That is the important one for me now. I am using the `in` keyword here: https://github.com/Ultimaker/Cura/commit/a7e23ba99e67b8f6e10c2edee88efcfe01257baa

No specific reason for upper, lower, startswith and endswith, yet.

Ghostkeeper:
I don't want to find out later that we can't add a bugfix to the PPA or something because we didn't add `lower()` to the whitelist. So if we're confident that it doesn't add risk then I'd rather add these functions to the whitelist. I can see the use case of `lower` in particular for checking cases case-insensitively in combination with `in`. We've changed the casing of g-code flavours in the past.

PR #592, opened 2020-03-24
Fix visiting non-string constants in Python 3.8

fieldOfView:
This PR fixes checking non-string constants in Python 3.8. #587 introduced a regression because it incorrectly assumes all constants are strings.

Fixes https://github.com/Ultimaker/Cura/issues/7335


thopiekar:
This brings Cura back to life!
Big thanks @fieldOfView !

image

nallath:
CURA-7322

PR #593, opened 2020-03-24
Add helper class to open URLs from QML

konskarm:
Added a helper class that restricts the schemes allowed when opening
a URL. For the time being, only http and https schemes are allowed to
be opened from QML files.

Related Cura PR: https://github.com/Ultimaker/Cura/pull/7344

CURA-7310


UM/Qt/Bindings/Utilities.py
```diff
@@ -16,34 +16,35 @@ class UrlUtil(QObject):
     """
     Helper class used to open URLs from QML.
     """
-    valid_uri_schemes = ["http", "https"]
+    __valid_uri_schemes = ("http", "https")
 
     def __init__(self, parent = None):
         super().__init__(parent)
 
     @pyqtSlot(str, list)
```
fieldOfView: If the function returns a bool, the pyqtSlot should specify a Result

UM/Qt/Bindings/Utilities.py
```diff
@@ -16,34 +16,35 @@ class UrlUtil(QObject):
...
         """
         Checks whether the target_url has an allowed scheme and, if it does, it opens the URL. If the target_url has a
         disallowed or invalid scheme, then it logs an error. This function can be called inside QML files.
 
         :param target_url: The URL string to be opened e.g. 'https://example.org'
         :param schemes: A list of the schemes that are allowed to be opened e.g. ['http', 'https']
-        :return: None
+        :return: True if the URL opens successfully, False if an invalid scheme is used
         """
         allowed_schemes = set()
```
fieldOfView: `allowed_schemes = list(set(self.__valid_uri_schemes) & set(schemes))`

konskarm:
Thanks for the review @fieldOfView . I am refactoring the code and adding tests, so it is still a WIP. I will ask for your review again once I am finished with those.

fieldOfView:
Isn't it codestyle to have a class in a file with the same name? This is a class named `UrlUtil` in a file named `Utilities.py`. Talking about codestyle, I thought there was a rule not to use abbreviations. So the class name would have to be `UrlUtilities`.

PR #595, opened 2020-04-01
CURA 7333 Add internal variable to keep track of object idx

konskarm:
The ReadMeshJob is called for each object in the 3mf file individually.
Therefore, when reloading a 3mf file, the ReadMeshJob has to be aware
of which object it is reloading. To do that, it keeps a track of it
through the new internal variable object_to_be_reloaded.

This variable is read back in cura in order to request the object at
that specific index.

CURA-7333


konskarm:
Merge this along with https://github.com/Ultimaker/Cura/pull/7384

PR #604, opened 2020-04-22
Speed up getting the cfg version by using regex instead of configparser

nallath:
This drastically speeds up the upgrader. On a regular boot, it took ~1.8
sec when upgrading, vs ~0.05s when upgrading with this fix (do note that
it's running with a profiler, so reality is about ~4x as fast on both sides)

See also: https://github.com/Ultimaker/Cura/pull/7559

### DIFF
--- a/UM/VersionUpgrade.py
+++ b/UM/VersionUpgrade.py
@@ -1,6 +1,6 @@
-# Copyright (c) 2016 Ultimaker B.V.
+# Copyright (c) 2020 Ultimaker B.V.
 # Cura is released under the terms of the LGPLv3 or higher.
-
+import re
 from UM.PluginObject import PluginObject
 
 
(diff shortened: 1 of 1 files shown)


UM/VersionUpgrade.py
```diff
@@ -19,6 +19,23 @@ def __init__(self):
...
+
+        In all config files that concern this version upgrade, the version number is stored in general/version, so get
+        the data from that key.
+
+        :param serialised: The contents of a config file.
+        :return: The version number of that config file.
+        """
+
+        format_version = int(self._version_regex.search(serialised).groups()[-1])
+        setting_version = int(self._setting_version_regex.search(serialised).groups()[-1])
```
Ghostkeeper, line 37: We need to have a fallback that sets the `setting_version` to 0 if it's not present?

nallath: Isn't the profile just invalid if it's not there?

Ghostkeeper: No, profiles from before Cura 2.6 didn't have any `setting_version`.

The profile was previously marked as invalid if the normal `version` isn't there. However with your new change you also made that fall back to 0. I guess that's fine too. The profile would then get dropped rather than marked as corrupt, since there is no upgrade path from file version 0 to the most recent version.

Ghostkeeper:
The version upgrade still works correctly. My (sample size 3) upgrade from 4.5 to 4.7 now took on average 0.0268s instead of 0.0450s. I don't know why it would be taking 1.8s for you, since it's only checking the files in your configuration folder, which is a couple of dozen files. But it is indeed a speed improvement by not creating a couple of dozen extra ConfigParsers.

PR #608, opened 2020-05-09
View.GL: Correcting equality check of numbers

thopiekar:
First this will get rid of the messages below and second it is the correct way to check the equality of numbers.
Here is the difference:
```
>>> x = int(1000.0)
>>> x is 1000
False
>>> x = 1000
>>> x is 1000
False
>>> x == 1000
True
```

```
  if common_path is "" or not common_path.startswith(install_prefix):
/usr/lib/python3/dist-packages/UM/View/GL/ShaderProgram.py:340: SyntaxWarning: "is" with a literal. Did you mean "=="?
  elif type(value) is list and type(value[0]) is list and len(value[0]) is 4:
/usr/lib/python3/dist-packages/UM/View/GL/ShaderProgram.py:342: SyntaxWarning: "is" with a literal. Did you mean "=="?
  elif type(value) is list and len(value) is 2:
/usr/lib/python3/dist-packages/UM/View/GL/ShaderProgram.py:344: SyntaxWarning: "is" with a literal. Did you mean "=="?
  elif type(value) is list and len(value) is 3:
/usr/lib/python3/dist-packages/UM/View/GL/ShaderProgram.py:346: SyntaxWarning: "is" with a literal. Did you mean "=="?
  elif type(value) is list and len(value) is 4:
/usr/lib/python3/dist-packages/UM/View/GL/ShaderProgram.py:348: SyntaxWarning: "is" with a literal. Did you mean "=="?
  elif type(value) is list and type(value[0]) is list and len(value[0]) is 2:
```

### DIFF
--- a/UM/View/GL/ShaderProgram.py
+++ b/UM/View/GL/ShaderProgram.py
@@ -337,15 +337,15 @@ class ShaderProgram:
         elif type(value) is Color:
             self._shader_program.setUniformValue(uniform,
                 QColor(value.r * 255, value.g * 255, value.b * 255, value.a * 255))
-        elif type(value) is list and type(value[0]) is list and len(value[0]) is 4:
+        elif type(value) is list and type(value[0]) is list and len(value[0]) == 4:
             self._shader_program.setUniformValue(uniform, self._matrixToQMatrix4x4(Matrix(value)))
-        elif type(value) is list and len(value) is 2:
+        elif type(value) is list and len(value) == 2:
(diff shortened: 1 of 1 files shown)


Ghostkeeper:
I'm not getting those warnings. Strange. But indeed what you're saying makes sense in theory, and the shaders still seem to work correctly in practice. Thanks!

thopiekar:
No problem :)

PR #610, opened 2020-05-15
Workaround: Naughty graphics drivers fall back to legacy.

rburema:
Recently, some graphics drivers showed up that only support glsl 400 while reportedly being 4.1 version OpenGL. This should not be possible, since the minimum for OpenGL 4.1 is glsl 410, but well, here we are.

### DIFF
--- a/UM/View/GL/OpenGLContext.py
+++ b/UM/View/GL/OpenGLContext.py
@@ -153,6 +153,7 @@ class OpenGLContext:
                 gl = cls.detect_ogl_context.versionFunctions(gl_profile) # type: Any #It's actually a protected class in PyQt that depends on the requested profile and the implementation of your graphics card.
 
                 gpu_type = "Unknown"  # type: str
+                glsl_version = "Unknown"  # type: str
 
                 result = None
                 if gl:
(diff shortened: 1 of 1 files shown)


rburema:
Nope this won't work ... at all. It conflicts with the _other_ Intel workaround where we can't delete graphics contexts :'-(

fieldOfView:
Aww...

Ghostkeeper:
I would normally delete the branch. However maybe you want to keep this one in for a while?

rburema:
@Ghostkeeper I don't know, when can we be sure that everyone has updated their 'old' intel drivers? (The ones that crash a program when you delete an OpenGL context.)

Otherwise this approach makes no sense.

Ghostkeeper:
It takes a while before people can update their Intel drivers, because some people are not able to update to the latest drivers but only to the latest drivers that are approved by their OEM. For instance, Lenovo could have a years' delay or so before they release those fixed drivers for their own laptops, and only then will Windows indicate that there is an update available.

rburema:
That's about what I expected, yes.

On the other hand, I guess it's not much harm to keep the branch around until the next cleanup though. (Not that I would be against deleting it per-se. It can be recreated without too much work if needed.)

PR #611, opened 2020-05-21
Cura 7222 get coordinates in camera tool

jellespijker:
CURA-7222 describe that the camera-plugin should be able to read the state vector of the camera, manipulate, store and retrieve. The state vector of the camera in Euler space, consists of its position [x, y, z] and [roll, pitch, yaw], when the camera is in orthographic projection this is expanded with a zoom factor.
In order to accommodate the camera-plugin the following camera properties where exposed to the proxy:
- get/set Position
- get/set Orientation
- get/set Projection
- get/set Perspective


jellespijker:
Due to a rewrite of the cura-camera-plugin the PR is no longer relevant. I'm closing it.

PR #612, opened 2020-05-25
Add widgets to Rotate by 90 degrees

fieldOfView:
This PR adds 3 pairs of arrow widgets to the Rotate ToolHandle, to rotate objects by exactly 90 degrees. To do so, it adds a way for ToolHandle instances to register additional widgets/controls. In this PR, the arrow widgets can still be dragged like the axes, but when clicked they rotate the model by +/- 90 degrees around the relevant axis.

image

This feature covers a large chunk of the usecases for a panel with degree value boxes (https://github.com/Ultimaker/Cura/issues/2779) without the added can of worms that would open.


nallath:
CURA-7468

nallath:
I really like this :D

PR #614, opened 2020-06-02
Create an UnclickableItem component that consumes clicks

konskarm:
The UnclickableItem component has a MouseArea in order
to consume unwanted clicks inside the item. This is primarily
used in order to avoid accidentally zooming or twisting the scene
while hovering over components that reside above the scene, such as
the MainWindowHeader in Cura.

Created during Turbo Testing and Tooling.


Ghostkeeper:
Is it maybe worth the time to use this component for a few things that need it, before we merge it in?

PR #618, opened 2020-06-05
Reset Qt Network manager on unknown network error

ninovanhooff:
Also fix a spammy internal Qt error because we called error() too many
times.

CURA-7437

### DIFF
--- a/UM/TaskManagement/HttpRequestManager.py
+++ b/UM/TaskManagement/HttpRequestManager.py
@@ -398,12 +398,14 @@ class HttpRequestManager(TaskManager):
         # We do nothing if the request was aborted or and error was detected because an error callback will also
         # be triggered by Qt.
         reply = request_data.reply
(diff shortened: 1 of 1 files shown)


UM/TaskManagement/HttpRequestManager.py
```diff
@@ -393,8 +398,9 @@ def _onRequestFinished(self, request_data: "HttpRequestData") -> None:
         # We do nothing if the request was aborted or and error was detected because an error callback will also
         # be triggered by Qt.
         reply = request_data.reply
-        if reply is not None and reply.error() != QNetworkReply.NoError:
-            if reply.error() == QNetworkReply.OperationCanceledError:
+        reply_error = reply.error() # error() must only be called once
```
ChrisTerBeke, line 401: You now call reply.error before checking if reply is not None, this could result in a crash?

PR #624, opened 2020-06-24
CURA-7106 speedup multiple objects on build plate

nallath:
CURA-7106

needs https://github.com/Ultimaker/Cura/pull/7971

### DIFF
--- a/UM/View/RenderBatch.py
+++ b/UM/View/RenderBatch.py
@@ -286,21 +286,6 @@ class RenderBatch:
             self._shader.enableAttribute("a_uvs", "vector2f", offset)
             offset += mesh.getVertexCount() * 2 * 4
 
-        for attribute_name in mesh.attributeNames():
-            attribute = mesh.getAttribute(attribute_name)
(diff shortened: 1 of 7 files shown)


UM/View/RenderBatch.py
```diff
@@ -247,7 +247,6 @@ def _renderItem(self, item: Dict[str, Any]):
             normal_matrix = Matrix(transformation.getData())
             normal_matrix.setRow(3, [0, 0, 0, 1])
             normal_matrix.setColumn(3, [0, 0, 0, 1])
-            normal_matrix.invert()
```
ninovanhooff, line 250: I would expect that removing an invert has some serious behavioral change.
Was the invert placed there in error?

nallath: The invert was never needed. As far as I've been able to find from the code, we just care about the general angle here (not the actual direction). This invert was also the most expensive part of the calculations here, so this is probably one of the most influential changes wrt speed (especially when you render a lot of items at once) in this PR.

PR #625, opened 2020-06-26
Replace configparser with a faster alternative

nallath:
When reading the metadata, we use the python config parser to do so. But the config parser has a lot of overhead (and functionalities that we dont actually need / use).

In order to fix this, i've made a lightweight alternative called FastConfigParser that uses two regexes to handle the complete parsing. This makes the loading of metadata a little more than twice as fast (From 2.5 seconds to 1.2 seconds for all metadata on boot).


UM/FastConfigParser.py
```diff
@@ -0,0 +1,49 @@
...
+
+    @staticmethod
+    def _convertValue(value: str) -> supported_data:
+        """
+        There could also be floats / ints in the data, so we try to convert these first.
+        :param value:
+        :return: The converted value if it could be casted to an int / float, the original if not.
+        """
+        try:
+            return int(value)
```
Ghostkeeper, line 36: The original `configparser` leaves them as strings. If this is a drop-in replacement we can also leave it as strings since the consumer must parse it anyway then.

UM/FastConfigParser.py
```diff
@@ -0,0 +1,49 @@
...
+supported_data = Union[int, float, str]
+
+
+class FastConfigParser:
+    """
+    This class is to replace the much slower configparser provided by Python itself.
+    It's probably nowhere near as robust and supports only a fraction of the functionality of the real deal.
+    """
+    header_regex = re.compile("\[(.*)\]\n([^\[]*)")
+    key_value_regex = re.compile("([^=\n ]+) *= *([^\n]+)")
```
Ghostkeeper, line 14: I believe we are using line continuation of the config files. See the [example here](https://docs.python.org/3/library/configparser.html#supported-ini-file-structure) under "Multiline Values". It's used if you have multiple post-processing scripts in one printer.

nallath: The current fastConfigParser only works for parsing InstanceContainers (which won't have multi-line post-processing anyway).

UM/FastConfigParser.py
nallath: Hmm, I don't know why I changed that. Oh well, easy enough to remove it.

UM/FastConfigParser.py
fieldOfView: It would be great if it were documented what major features are not supported, instead of "supports only a fraction of the functionality", so (plugin-?)developers know whether this is safe to use or the real deal should be used instead.

nallath: Fair enough. So far it supports the most basic functionality of the config parser (So comments are likely not working).

Most of the profiles are just header seperated key value pairs, so that is what the config parser is able to load (and able to load very fast).

Ghostkeeper: Also with the start and end g-code.

nallath: Should be fixed now! I'm adding some tests to make extra sure that all of this works.

UM/Settings/InstanceContainer.py
```diff
@@ -611,8 +611,7 @@ def deserializeMetadata(cls, serialized: str, container_id: str) -> List[Dict[st
         """
 
         serialized = cls._updateSerialized(serialized)  # Update to most recent version.
-        parser = configparser.ConfigParser(interpolation = None)
-        parser.read_string(serialized)
+        parser = FastConfigParser(serialized)
```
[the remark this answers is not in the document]
nallath: It's constructing the metadata because it needs to do so anyway. Then it puts the id and the type in it (and after that, whatever it got from the file) I don't see how any work is wasted.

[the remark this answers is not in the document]
ninovanhooff: Right, I thought is was overwritten but now see it was appended

PR #626, opened 2020-06-28
CICD: Lint Python code for undefined names

cclauss:


Ghostkeeper:
I'd prefer such linting to be executed via CTest so that we can run them locally.

I'm also not entirely sure if this is really going to cooperate nicely with things like our [setting property system](https://github.com/Ultimaker/Uranium/blob/master/UM/Settings/InstanceContainer.py#L316) and duck-typing-based scene nodes. We use `hasattr` in quite a few places throughout our code base.

cclauss:
> return key in self._instances and hasattr(self._instances[key], property_name)

`return hasattr(self._instances.get(key), property_name)`

cclauss:
The issues raised do not seem to be related to `hasattr()`.

nallath:
The issues seem to be caused by some very very old benchmarking scripts that, to my knowledge, are not used at all (They also seem to not have been touched in years...)

cclauss:
Undefined names are good ways to find dead code because they have potential to raise NameError if the code is actually run.

Ghostkeeper:
> The issues raised do not seem to be related to `hasattr()`.

What I mean is indeed not specifically related to `hasattr`, but our setting property system creates dynamic setting properties. We can ask for `self.description` without `description` ever being explicitly set. Apparently this linter doesn't check for that though.

Where a name was undefined in typing, please only import the name `if TYPE_CHECKING`, like we do here for instance:

https://github.com/Ultimaker/Cura/blob/master/cura/CuraApplication.py#L119-L120

And please connect this via CTest rather than as a separate workflow, or we'd need to run many separate commands in order to run all tests locally!

PR #628, opened 2020-07-01
Cura 6253 fix remove container that requires to load a container first

jellespijker:
If a container is partial (only metadata) loaded it shouldn't have to
load the container before removal. The emitted RemovedContainer signal
expects a container to be passed along with it. This will be an
EmptyInstanceContainer containing the metadata of the container.

### DIFF
--- a/UM/Settings/ContainerRegistry.py
+++ b/UM/Settings/ContainerRegistry.py
@@ -16,6 +16,7 @@ from UM.Logger import Logger
 from UM.MimeTypeDatabase import MimeType, MimeTypeDatabase
 from UM.PluginRegistry import PluginRegistry #To register the container type plug-ins and container provider plug-ins.
 from UM.Resources import Resources
+from UM.Settings.EmptyInstanceContainer import EmptyInstanceContainer
 from UM.Settings.ContainerFormatError import ContainerFormatError
 from UM.Settings.ContainerProvider import ContainerProvider
 from UM.Settings.constant_instance_containers import empty_container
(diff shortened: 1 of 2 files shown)


UM/Settings/ContainerRegistry.py
```diff
@@ -441,12 +442,10 @@ def removeContainer(self, container_id: str) -> None:
...
-                # has been deleted, but due to lazy loading, it hasnt even been loaded yet. The issues is that in order
-                # to notify the rest of the code, we need to actually *have* the container. So we need to load it
-                # in order to remove it...
-                provider = self.source_provider.get(container_id)
-                if provider:
-                    container = provider.loadContainer(container_id)
+                # has been deleted, but due to lazy loading, it hasn't been loaded yet. The issues is that in order
+                # to notify the rest of the code, we need to actually *have* the container. So an empty instance 
+                # container is created, which is emitted with the containerRemoved signal and contains the metadata
+                container = EmptyInstanceContainer(container_id)
```
konskarm, line 448: The metadata of the old container should be transferred to the new EmptyInstanceContainer, because some functions that are activated by the containerRemoved signal take actions based on the container's metadata (e.g. see CloutOutputDeviceManager.py:456 or IntentcategoryModel:95).

jellespijker: Looked through our own codebase and imo the branches that are executed with a request to non-existing metadata, such as: `type` don't break when this key isn't present. That being said It could be that an external plugin uses the metadata. So just to be safe I created a commit in which the metadata is copied to the EmptyInstanceContainer.

Btw, the CloudOutputDeviceManager.py:456 will never request the metadata because it will first check if its a `GlobalStack` instance which it is not.

PR #629, opened 2020-07-02
CURA-7540 wrong setting visibility after project import

nallath:
Fixes https://github.com/Ultimaker/Cura/issues/7967

### DIFF
--- a/UM/Settings/Models/SettingDefinitionsModel.py
+++ b/UM/Settings/Models/SettingDefinitionsModel.py
@@ -213,7 +213,6 @@ class SettingDefinitionsModel(QAbstractListModel):
 
     def setExpanded(self, expanded: List[str]) -> None:
         """Set the expanded property"""
-
         new_expanded = set()
         for item in expanded:
             if item == "*":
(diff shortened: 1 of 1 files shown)


UM/Settings/Models/SettingDefinitionsModel.py
```diff
@@ -291,26 +298,42 @@ def _getDefinitionsByKey(self, key: str) -> List["SettingDefinition"]:
...
 
     #@deprecated("Use collapseRecursive instead.", "4.5")  # Commented out because these two decorators don't work together.
     @pyqtSlot(str)
     def collapse(self, key: str) -> None:
         return self.collapseRecursive(key)
 
-    ##  Hide the children of a specified SettingDefinition and all children of those settings as well.
     @pyqtSlot(str)
-    def collapseRecursive(self, key: str) -> None:
+    def collapseRecursive(self, key: str, *, emit_signal: bool = True) -> None:
```
konskarm, line 336: A little bit of documentation is required as to why the boolean trap is created here.

nallath: It's actually to prevent the boolean trap. I'll add the documentation.

PR #631, opened 2020-08-16
Updated pt_BR strings for 4.7

Patola:


Ghostkeeper:
That was a false positive. It was translated as `Inicializando gestor de pacotes...` so that seems to be fine.

PR #633, opened 2020-08-26
Fix key value pairs in comments being parsed as seperate values

nallath:
Fixes Ultimaker/Cura#8235
Fixes Ultimaker/Cura#8230


tests/Settings/config_parser_files/weird_values.cfg
```diff
@@ -6,6 +6,11 @@ even_more_weirdness = [yay!]
 woah_weird = =(200 if the_value != "whatever" or more_weirdness == "derp" else 0)
 lesser_equal = =60 if 1 <= 2 else 30
 greater_equal = =20 if 1 >= 2 else "yay"
+key_value_in_string = G1
+	; speed_z = {speed_z}
```
Ghostkeeper, line 10: Use spaces instead of tab.

```suggestion
    ; speed_z = {speed_z}
```

Ghostkeeper (approved):

Ghostkeeper:
Oh, it seems that my suggestion to replace the tab with spaces causes the test to fail :rofl:

nallath:
Well, that's why we have automatic running tests!

nallath:
It also needed to change the test (since that was checking for a tab instead of spaces as well)

Ghostkeeper:
It's not working correctly yet. I tried this start g-code:
```
;layer height={layer_height}
;infill={infill_sparse_density}
=just a value
bla
just_key=
and that concludes the test
```

Then I restart Cura and it leaves just this:
```
;layer height={layer_height}
```

And it also puts this in the log about it:
```
2020-08-27 10:54:51,899 - WARNING - [MainThread] UM.Settings.InstanceContainer.setProperty [376]: Tried to set value of setting 	;infill that has no instance in container Yoda_settings or its definition Ultimaker 3
2020-08-27 10:54:51,899 - WARNING - [MainThread] UM.Settings.InstanceContainer.setProperty [376]: Tried to set value of setting 	 that has no instance in container Yoda_settings or its definition Ultimaker 3
2020-08-27 10:54:51,899 - WARNING - [MainThread] UM.Settings.InstanceContainer.setProperty [376]: Tried to set value of setting 	just_key that has no instance in container Yoda_settings or its definition Ultimaker 3
```

Ghostkeeper:
Hmm the change of base doesn't seem to work. I'll cherry-pick them to 4.7 instead.

PR #636, opened 2020-09-06
OBJ files can have negative indices.

### LINKED ISSUE #635: OBJ file with "negative vertex reference numbers" does not load
**Application version**
Cura 4.7.0

**Platform**
Windows Version 10.0.18363 Build 18363

**Printer**
not relevant

**Reproduction steps**
1. Loaded OBJ file "Box FM.obj" (attached as a TXT file)

**Screenshot(s)**
Just a Warning message: "The selected model was too small to load."

**Actual results**
The above warning message.

**Expected results**
The OBJ model should load, a simple 20x20x5 box.

**Log file**
n/a

**Additional information**
This is perhaps an obscure feature of the OBJ file format; but for completeness would likely be easy enough to support.
Faces (lines beginning with 'f') may use "negative vertex reference numbers". (See http://www.martinreddy.net/gfx/3d/OBJ.spec). In this file, for example, a triangular face is represented by the sequence:
v 0 0 5
v 0 20 5
v 20 0 5
f -1 -2 -3
where the face references the previous 3 vertices with negative indexes.
I suspect this feature is not supported, resulting in the failure to load the file.
FWIW, Fusion 360 and the Windows 3D Viewer both load the file successfully.

[Box FM.txt](https://github.com/Ultimaker/Cura/files/5178532/Box.FM.txt)

rburema:
Should fix #635

Not tested yet, so this is a draft. Feel free to consider it non-draft if another dev has tested it.

### DIFF
--- a/plugins/FileHandlers/OBJReader/OBJReader.py
+++ b/plugins/FileHandlers/OBJReader/OBJReader.py
@@ -18,7 +18,7 @@ class OBJReader(MeshReader):
 
     def _toAbsoluteIndex(self, max, data):
         """ Handle negative indices (those are relative to the position, so -2 is the second one before the face). """
-        return [index if index > 0 else (1 + max) - index for index in data]
+        return [index if index > 0 else 1 + max + index for index in data]
 
     def _read(self, file_name):
         scene_node = None
--- a/plugins/FileHandlers/OBJReader/OBJReader.py
+++ b/plugins/FileHandlers/OBJReader/OBJReader.py
@@ -16,6 +16,10 @@ class OBJReader(MeshReader):
         super().__init__()
         self._supported_extensions = [".obj"]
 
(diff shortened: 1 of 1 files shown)


nallath:
Sorry, doesn't work :(. I've loaded the model as provided in https://github.com/Ultimaker/Uranium/issues/635 in blender and in Cura.

Without the pull request: "Model is too small to load"
With the pull request:
NegativeIndexBoxCura

Box in blender:
BlenderNegateiveIndexBox

rburema:
@nallath Turns out subtracting negative values makes a number larger, not smaller. Who knew! Anyway I turned the `-` into a `+` and now it works :-)

nallath (approved):

nallath:
CURA-7708

PR #638, opened 2020-09-21
Fix file watcher on windows for removable devices

awiegel:
Fixes file watcher on windows when loading a file from a removable device (USB/SD-Card/...). Crash happens because QEventLoop is expected to be active before adding a path to the file watcher and the removable device seems to not have one ready. Manually create QEventLoop to fix this issue (QEventLoop gets created anyways in a later stage).

Benefit: File watcher can be used again on windows for reloading files automatically on change.


rburema:
Hi @awiegel Sorry it took so long to get back. Seems like we can accept this without alteration (pending final QA/etc. approval). Thanks for the contribution!

PR #644, opened 2020-10-06
CURA-7736 Don't allow files to be outside of plugin-root

rburema:


UM/PluginRegistry.py
```diff
@@ -172,6 +173,65 @@ def _savePluginData(self) -> None:
...
+            current_dir = worklist.pop()
+
+            # If the directory under scrutiny is a signed folder, bundled, or in the ignore list, it's ok:
+            has_signature_file = os.path.isfile(os.path.join(current_dir, TrustBasics.getSignaturesLocalFilename()))
+            is_bundled = self._isPathInLocation(install_prefix, current_dir)
+            in_ignore_list = os.path.basename(current_dir) in plugin_path_ignore_list
+            if has_signature_file or is_bundled or in_ignore_list:
+                continue
+
+            # Otherwise, for all files in the current directory:
```
Ghostkeeper, line 207: Redundant documentation.

UM/PluginRegistry.py
```diff
@@ -172,6 +173,65 @@ def _savePluginData(self) -> None:
...
+
+        # Put the parent-root on the work-list:
+        worklist = [abs_path]
+        while worklist:
+            current_dir = worklist.pop()
+
+            # If the directory under scrutiny is a signed folder, bundled, or in the ignore list, it's ok:
+            has_signature_file = os.path.isfile(os.path.join(current_dir, TrustBasics.getSignaturesLocalFilename()))
+            is_bundled = self._isPathInLocation(install_prefix, current_dir)
+            in_ignore_list = os.path.basename(current_dir) in plugin_path_ignore_list
```
Ghostkeeper, line 203: Doesn't this ignore list break the security?

Entries on the ignore list are implicitly trusted here. One of the entries of the ignore list is `__pycache__`. The `__pycache__` folder overwrites any contents of the `.py` files of the parent folder if the cache is present and written more recently than the `.py` files with the same name. So by trusting `__pycache__` it is effectively blindly trusting code without checking the signature.

Correct me if I'm wrong here. Maybe I've misunderstood the ignore list.

Ghostkeeper (changes requested):

UM/PluginRegistry.py
rburema: Take a good look. The security as provided by the Trust object, is not worse for this, and the plugins are still checked by the Trust object. This part only checks whether these types of files should be allowed to live outside of the normal file-hierarchy. That is, outside of the plugin folders themselves.

_Inside_ of the plugin folders, that Trust object still checks everything but '__pycache__' (and even that only escapes because it's explicitly deleted before the check actually happens), and therefore still needs a signature. Even if it was a '.git' or 'tests' folder.

They're only implicitly trusted in the sense that we don't expect there to be references to '.git', 'test' (or indeed '__pycache__') in  the actual signed plugin. Something we should catch when signing a plugin anyway, because something weird is going on if you reference your own '.git', 'test' or '__pycache__' folder in production code.

So, as you can probably (hopefully, unless _I'm_ mistaken) see, security is stronger compared to the old situation. _However_ it would be stronger still if I did indeed remove that check anyway, and there's not _that_ much reason to keep it. It does however mean that you don't get to have _any_ extraneous files in the general 'plugin' area, not even those ones.

UM/PluginRegistry.py
rburema: Fixed: 95ca32d9

UM/PluginRegistry.py
rburema: I hope that the comment before made some sense, but in the end I decided that it was right to remove the check _anyway_. Devs for the enterprise (or that want the added security) will just have to learn to live without stray '.git', 'test' or '\_\_pycache\_\_' folders in their plugin locations. (That is to say, without a plugin 'around' them.)

Should you have changed opinion because of the explanation, feel free to revert the (latest) commit that fixes this comment: b404ca7d

PR #647, opened 2020-10-13
Introduce Deprecated class

konskarm:
So that deprecated classes can inherit from that, since the deprecated decorator cannot be used on classes.


nallath:
Can't you just use the deprecated on the constructor?

PR #649, opened 2020-10-20
Make approximate circle accept how many segments it should use

nallath:
In some cases, the 8 segments option is far too rough.

CURA-7643

### DIFF
--- a/UM/Math/Polygon.py
+++ b/UM/Math/Polygon.py
@@ -3,6 +3,7 @@
 from typing import Optional, Tuple, List, Union
 
 import numpy
+import math
 import scipy.spatial
 
 from UM.Logger import Logger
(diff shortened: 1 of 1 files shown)


konskarm (approved):

UM/Math/Polygon.py
```diff
@@ -22,16 +23,13 @@ def approximatedCircle(radius):
...
-            [radius, 0],
-            [radius * 0.707, -radius * 0.707],
-            [0, -radius],
-            [-radius * 0.707, -radius * 0.707]
-        ], numpy.float32))
+        step = 2 * math.pi / num_segments
+
+        points = []
+        for i in range(0, num_segments):
+            points.append([radius * -math.cos(i * step), radius * math.sin(i * step)])
```
Ghostkeeper, line 30: Not really performance-sensitive here, but for next time, Numpy supports vector operations like so:

```
points = numpy.arange(0, 2 * math.pi, step)
points = numpy.stack((numpy.cos(points), numpy.sin(points)))
```

PR #652, opened 2020-10-25
Updated pt_BR strings for Cura 4.8.

Patola:


Ghostkeeper:
Thanks! We'll review the translations in a week or so, when most of the other translations should also be in.

Devs, see our translation ticket for 4.8: CURA-7782.

Ghostkeeper (approved):

PR #657, opened 2020-12-04
Fix showing expanded/collapsed indicator arrow for setting categories

fieldOfView:
This PR fixes the expanded/collapsed indicator arrow for setting categories in setting views. It was broken in 905e8eb5b44fa6d16310aad80bec7ced32455b02. Reverting that change may have a performance hit, so some profiling may be in order.

Fixes https://github.com/Ultimaker/Cura/issues/8283, CURA-7692

### DIFF
--- a/UM/Settings/Models/SettingDefinitionsModel.py
+++ b/UM/Settings/Models/SettingDefinitionsModel.py
@@ -77,6 +77,8 @@ class SettingDefinitionsModel(QAbstractListModel):
 
         self.destroyed.connect(self._onDestroyed)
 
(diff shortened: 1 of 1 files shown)


Ghostkeeper:
Developers, see issue CURA-7692.

fieldOfView:
More than two months on, could this get some love? I almost started fixing this a second time, until I remembered that I fixed it before.

nallath:
I found a solution that is a bit in between, It's a bit slower, but not as bad as it was before the commit that introduced it all.

PR #658, opened 2020-12-05
Adding mirror property to RecolorImage

AmrBarghouthi:
This to makes it possible to flip the image in RecolorImage which is need it for some icons when translating Apps to support Right to Left languages

### DIFF
--- a/UM/Qt/qml/UM/RecolorImage.qml
+++ b/UM/Qt/qml/UM/RecolorImage.qml
@@ -12,6 +12,7 @@ Item
     property alias source: img.source
     property alias color: overlay.color
     property alias sourceSize: img.sourceSize
+    property alias mirror: img.mirror
 
     Image
     {
@@ -20,6 +21,7 @@ Item
         visible: false
         sourceSize.width: parent.width
         sourceSize.height: parent.height
+        layer.enabled: true
     }
 
     ColorOverlay {


AmrBarghouthi:
I added a function to indicate the language direction to be used in QML files 
example use is adding the following to row layout elements parent
```
LayoutMirroring.enabled: catalog.i18nIsRightToLeft()
LayoutMirroring.childrenInherit: true
```

Ghostkeeper:
I'm a bit confused as to why you would want to mirror images in right-to-left scripts?

And the solution of adding a translated string may not be very robust either. It would be better to get this from the metadata of the translation. I think `gettext.info()` can read that.

AmrBarghouthi:
@Ghostkeep image mirroring is going to be for stuff like arrows 

as for metadata of the translation I will look into it

AmrBarghouthi:
I update it the i18nIsRightToLeft function to get the direction from language metadata
thanks @Ghostkeeper for the suggestion

Ghostkeeper:
The only directional arrows that we have are really the folding arrows in the setting category headers. But those are not really related to any text, are they?

AmrBarghouthi:
yes but it's not only for arrows 
for example 

these icons look way better when mirrored in right to left application  
anyway it is a feature of Uranium that doesn't really do any harm if added

UM/Qt/qml/UM/RecolorImage.qml
```diff
@@ -20,6 +21,7 @@ Item
         visible: false
         sourceSize.width: parent.width
         sourceSize.height: parent.height
+        layer.enabled: true
```
Ghostkeeper, line 24: I'll change this to `layer.enabled: mirror` to save some rendering performance. These images are used a lot (e.g. with every setting) so performance is important.

Ghostkeeper (approved):

Ghostkeeper:
I can't seem to be able to get this to work.

I added this property to the Dutch translation .po files (all 4):

    "Language-Direction: rtl\n"

I then recompiled the .po files to .mo files and printed the `isRightToLeft()` function from QML:

`print(catalog.i18nIsRightToLeft());`

However it still returns `false`.

Setting the mirror property works, but setting it to depend on being right-to-left never mirrors the image.

PR #660, opened 2020-12-24
CURA-7868: Introduce the file provider plugin type

konskarm:
CURA-7868

### DIFF
--- a/UM/FileProvider.py
+++ b/UM/FileProvider.py
@@ -1,12 +1,13 @@
-# Copyright (c) 2020 Ultimaker B.V.
+# Copyright (c) 2021 Ultimaker B.V.
 # Uranium is released under the terms of the LGPLv3 or higher.
-from PyQt5.QtCore import pyqtSignal
+
+from PyQt5.QtCore import pyqtSignal, QObject
 
 from UM.PluginObject import PluginObject
 from typing import Optional
 
 
-class FileProvider(PluginObject):
+class FileProvider(PluginObject, QObject):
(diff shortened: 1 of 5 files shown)


UM/FileProvider.py
```diff
@@ -0,0 +1,77 @@
...
+    in the Open File(s) submenu"""
+
+    def __init__(self) -> None:
+        super().__init__()
+        self._menu_item_display_text = None  # type: Optional[str]
+        self._menu_item_name = None  # type: Optional[str]
+        self._shortcut = None  # type: Optional[str]
+        self._enabled = True
+
+    def setMenuItemDisplayText(self, display_text: str) -> None:
```
Ghostkeeper, line 27: Rather than using setters and getters, which are public and not intended to be overridden by subclasses, it is simpler to just have a public field. That's what they're for, after all.

Ghostkeeper (changes requested):

Ghostkeeper:
For consistency of the interface, it would be good to provide a default plug-in that provides the local file dialogue. This means that the action needs to be removed from the Cura actions screen, yes, but it'll be re-added through that plug-in then.

UM/FileProvider.py
Ghostkeeper: Fixed here:
* https://github.com/Ultimaker/Uranium/pull/660/commits/ce0d2728cb04f987078ffb0f5d5d0ffbef1373a7
* https://github.com/Ultimaker/cura-private-data/pull/66/commits/46902a45a7173b5e857736c2943147eb561dc123

PR #662, opened 2021-01-08
Open remote files MVP

rburema:
Digital Factory!

Internal ticket: CURA-7864

See also:
https://github.com/Ultimaker/Cura/pull/9089
https://github.com/Ultimaker/cura-private-data/pull/68


rburema:
Tests fail, but doesn't seem to be related to this code?

PR #667, opened 2021-01-29
Replace non-existant property visible with isVisible.

Ghostkeeper:
Turns out isVisible (the QtApplication function) was never called, so _main_window.visible was never called, so no-one realized that this latter one didn't exist. But _main_window _also_ has a isVisible (function not property) so now that is called by the QtApplication isVisible (property) instead.


Ghostkeeper (commented):
Looks good to me.

PR #668, opened 2021-01-29
[CURA-7995] Fix from/to remote menu order w.r.t. normal flow

rburema:
Which is done by adding a 'priority' field to FileProvider (at least what Uranium is concerned).

### DIFF
--- a/UM/FileProvider.py
+++ b/UM/FileProvider.py
@@ -40,3 +40,7 @@ class FileProvider(PluginObject, QObject):
     def run(self) -> None:
         """Call function associated with the file provider"""
         raise NotImplementedError
+
+    def getPriority(self) -> int:
+        """Where it should be sorted in lists, or which should be tried first."""
+        return -1
--- a/UM/FileProvider.py
+++ b/UM/FileProvider.py
@@ -37,10 +37,11 @@ class FileProvider(PluginObject, QObject):
         If the provider is not enabled, it should not be displayed in the interface.
         """
 
+        self.priority = 0
+        """
+        Where it should be sorted in lists, or which should be tried first.
+        """
+
     def run(self) -> None:
         """Call function associated with the file provider"""
         raise NotImplementedError
-
-    def getPriority(self) -> int:
-        """Where it should be sorted in lists, or which should be tried first."""
-        return -1
--- a/UM/OutputDevice/OutputDeviceManager.py
+++ b/UM/OutputDevice/OutputDeviceManager.py
@@ -7,10 +7,10 @@ from UM.Signal import Signal, signalemitter
 from UM.Logger import Logger
 from UM.PluginRegistry import PluginRegistry
 
-from .ProjectOutputDevice import ProjectOutputDevice
 
 if TYPE_CHECKING:
     from UM.OutputDevice.OutputDevice import OutputDevice
+    from UM.OutputDevice.ProjectOutputDevice import ProjectOutputDevice
     from UM.OutputDevice.OutputDevicePlugin import OutputDevicePlugin
 
 
@@ -124,7 +124,7 @@ class OutputDeviceManager:
     activeDeviceChanged = Signal()
     """Emitted whenever the active device changes."""
 
-    def getOutputDevices(self) -> ValuesView[OutputDevice]:
+    def getOutputDevices(self) -> ValuesView["OutputDevice"]:
         """Get a list of all registered output devices.
 
         :return: :type{list} A list of all registered output devices.
@@ -132,7 +132,7 @@ class OutputDeviceManager:
 
         return self._output_devices.values()
 
-    def getProjectOutputDevices(self) -> ValuesView[ProjectOutputDevice]:
+    def getProjectOutputDevices(self) -> ValuesView["ProjectOutputDevice"]:
         """Get a list of all registered output devices.
 
         :return: :type{list} A list of all registered output devices.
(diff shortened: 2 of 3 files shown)


fieldOfView:
In most other places where order is affected, the term "weight" is used instead of "priority".

UM/FileProvider.py
```diff
@@ -39,4 +39,8 @@ def __init__(self) -> None:
 
     def run(self) -> None:
         """Call function associated with the file provider"""
-        raise NotImplementedError
\ No newline at end of file
+        raise NotImplementedError
+
+    def getPriority(self) -> int:
```
konskarm, line 44: We can make the `priority` a property of the FileProvider to be consistent with the other properties.

UM/Qt/Bindings/FileProviderModel.py
```diff
@@ -59,9 +60,12 @@ def update(self) -> None:
                     "name": plugin_id,
                     "displayText" : file_provider.menu_item_display_text,
                     "fileProvider": file_provider,
-                    "shortcut": file_provider.shortcut
+                    "shortcut": file_provider.shortcut,
+                    "priority": file_provider.getPriority(),
                 })
 
+        self.sort(lambda x: -float(x["priority"]))
```
konskarm, line 67: Is there a specific reason to convert it into a float?

UM/Qt/Bindings/FileProviderModel.py
```diff
@@ -46,7 +46,8 @@ def update(self) -> None:
             "name"         : "LocalFileProvider",
             "displayText"  : "From Disk",
             "fileProvider" : None,  # it's not loaded via a plugin, so its FileProvider is empty
-            "shortcut"     : "Ctrl+O"
+            "shortcut"     : "Ctrl+O",
+            "priority"     : 0,
```
konskarm, line 50: The "From Disk" option should always appear first, so we should increase the priorty of the LocalFileProvider.
```suggestion
            "priority"     : 99,
```

UM/Qt/Bindings/FileProviderModel.py
```diff
@@ -59,9 +60,12 @@ def update(self) -> None:
                     "name": plugin_id,
                     "displayText" : file_provider.menu_item_display_text,
                     "fileProvider": file_provider,
-                    "shortcut": file_provider.shortcut
+                    "shortcut": file_provider.shortcut,
+                    "priority": file_provider.getPriority(),
```
konskarm, line 64: Convert the priority into a property of the FileProvider
```suggestion
                    "priority": file_provider.priority,
```

konskarm:
> In most other places where order is affected, the term "weight" is used instead of "priority".

This is actually consistent with the OutputDevices, which also use priorties.

fieldOfView:
Tools, Stages, Views, Definitions, Qualities, Intents and SettingVisibilityPresets all use "weights".  It would be nice if there were a code standard.

Ghostkeeper:
True. But I recognise the difficulty in setting code standards for this sort of terminology. There are too many such words to standardise them, and we'd never remember them. I'll accept the variable name as it is.

UM/OutputDevice/OutputDeviceManager.py
```diff
@@ -14,13 +14,20 @@
...
+    Values:
+     * NO       = 0: The plugin can't add a device 'manually' (or at least not with the given parameters).
+     * POSSIBLE = 1: The plugin will try to add the (specified) device 'manually', unless another plugin has priority.
+     * PRIORITY = 2: The plugin has determined by the specified parameters that it's responsible for adding this device and thus has priority.
+     If this fails, the plugins that replied 'POSSIBLE' will be tried.
+     NOTE: This last value should be used with great care!
+    """
+    NO = 0,
+    POSSIBLE = 1,
+    PRIORITY = 2
```
Ghostkeeper, line 30: I liked the original formatting better, honestly. This duplicates the entire class in documentation. Since it is quite subjective I'm also fine with leaving it like this though.

UM/Qt/Bindings/FileProviderModel.py
```diff
@@ -46,7 +46,8 @@ def update(self) -> None:
             "name"         : "LocalFileProvider",
             "displayText"  : "From Disk",
             "fileProvider" : None,  # it's not loaded via a plugin, so its FileProvider is empty
-            "shortcut"     : "Ctrl+O"
+            "shortcut"     : "Ctrl+O",
+            "priority"     : 99,  # Assign a high value to make sure it appears on top in the File->Open File(s) submenu
```
Ghostkeeper, line 50: This default makes no sense to me. All file providers get this priority unless they override it, right? Why would all of them need to appear on top?

Nobody is

Ghostkeeper (changes requested):

UM/Qt/Bindings/FileProviderModel.py
konskarm: FileProviders get a priority of 0 by default https://github.com/Ultimaker/Uranium/blob/9b3082238711ace62496cfc8e2669395005dcb46/UM/FileProvider.py#L40. That's why the 99 is given only to the LocalFileProvider item of the FileProviderModel.

UM/OutputDevice/OutputDeviceManager.py
konskarm: I did not find any standard in Python on how to document Enums. I'll just revert it to the way it was.

nallath:
It should be weight! Yes it's hard to standardise things, but in this case _we noticed it_.

rburema:
@fieldOfView  We've chosen to accept this for now, but only because there are output device plugins in the wild too, which means that those should remain priority unless we want a breaking change, in which case we keep it like this here (for now) since the code here is similar to that of output devices.

We'll probably look at it again when we have to make a breaking change anyway.

PR #672, opened 2021-02-25
Fix reading OBJ for vertex/no-texture/normal case.

rburema:
Just make sure the face always has 9 data points, but load the signalling invalid ones when that part is missing.

fixes Ultimaker/Cura#9284

_Update_: Made internal jira ticket CURA-8049 so this'll show up in our changelog for when it's merged & QA'd.


Ghostkeeper (approved):
I think it's good like this too, but it's hard to reason about so I'll give this a thorough test with various types of Wavefront OBJ files...

rburema:
> I think it's good like this too, but it's hard to reason about so I'll give this a thorough test with various types of Wavefront OBJ files...

We'd probably set it up differently if we where to implement it now. Hopefully this is the last change needed to get it finally working for all cases. (I also hope I didn't jinx it with that statement...)

fieldOfView:
Alternatively, we could have TrimeshReader handle OBJs.

rburema:
@fieldOfView I could be wrong, but skimming the code in trimesh I don't think it can handle 'relative' indexing (which is when the indices in the faces are set to negative).

PR #676, opened 2021-03-10
CURA-8050: Remove project output devices from the list when they get disabled

konskarm:
Previously, enabling/disabling a project output device would not reflect the change in the actual output devices list.

Now, if a project output device gets disabled, it will be automatically removed from the output_devices list if it was added there in the first place. When it gets enabled again, the device is readded to this list, making it available to be selected from the OutputDevicesActionButton in the bottom right.

CURA-7903

### DIFF
--- a/UM/OutputDevice/OutputDeviceManager.py
+++ b/UM/OutputDevice/OutputDeviceManager.py
@@ -247,12 +247,10 @@ class OutputDeviceManager:
             self.resetActiveDevice()
         return True
 
-    def addProjectOutputDevice(self, device: "ProjectOutputDevice", add_to_output_devices: bool = False) -> None:
+    def addProjectOutputDevice(self, device: "ProjectOutputDevice") -> None:
         """Add and register a project output device.
 
         :param device: The output device to add.
-        :param add_to_output_devices: Boolean to determine whether the device should also be added to the _output_devices
-                                      list. If that happens, the device will appear as an option also in the OutputDevicesActionButton
 
         :note Does nothing if a device with the same ID as the passed device was already added.
         """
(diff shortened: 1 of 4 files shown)


UM/OutputDevice/OutputDeviceManager.py
```diff
@@ -247,12 +247,10 @@ def removeOutputDevice(self, device_id: str) -> bool:
             self.resetActiveDevice()
         return True
 
-    def addProjectOutputDevice(self, device: "ProjectOutputDevice", add_to_output_devices: bool = False) -> None:
+    def addProjectOutputDevice(self, device: "ProjectOutputDevice") -> None:
```
rburema, line 250: Did/does any other ProjectOuputDevice use this parameter, or was it added originally because of the DF one?

plugins/LocalFileOutputDevice/LocalFileOutputDevice.py
```diff
@@ -24,8 +24,8 @@
 class LocalFileOutputDevice(ProjectOutputDevice):
     """Implements an OutputDevice that supports saving to arbitrary local files."""
 
-    def __init__(self, parent = None):
-        super().__init__(device_id = "local_file", parent = parent)
+    def __init__(self, add_to_output_devices: bool = True, parent = None):
```
rburema, line 27: Other plugins (not ours) that implement ProjectOutputDevice will have to do this as well, then? It's an API break in any case.

Not necessarily a deal-breaker, but we should be careful. Let's ask peoples opinions on this during eCCB.

rburema: Never mind, I was made aware that it has a default. (which should have been clear from the default being right in my face in the first place :-p )

UM/OutputDevice/OutputDeviceManager.py
rburema: Ok, turns out the parameter wasn't present in 4.8 anyway, nothing to see here!

rburema (approved):

rburema:
Seriously 'CURA-8050_Ensure_that_disabled_project_output_devices_are_removed_from_the_OutputDevicesActionButton'?

...

I'm glad github has that 'copy branchname' feature :-p

Ghostkeeper:
Git's tab completion is a great help too ;)

PR #684, opened 2021-04-02
Fix mypy warnings caused by pythong & numpy upgrade

nallath:


konskarm (approved):

konskarm:
We accept for now that the HttpRequestManager tests will be skipped, since they produce Segfaults when writting their results using the junitxml.

PR #686, opened 2021-04-04
Add missing Czech translations for v4.9

sustmi:


Ghostkeeper:
Thanks for the update! You were fast.

We'll review these translations before the 4.9 stable version is released (just whether the syntax is okay, we don't really have Czech expertise). We'll wait a bit until most of the translations are in, so we can process them in bulk. That makes it a bit easier for us. So just a few more days.

jellespijker:
Looks good @sustmi, I will merge them and re-enable the translations again. Thank you for your help.

PR #689, opened 2021-04-11
Updated pt_BR translation

Patola:
Added new labels and fixed some verb declinations


jellespijker:
Looks good @Patola, thank you for your help. It is appreciated.

PR #691, opened 2021-04-14
[CURA-8005] Crashes reported, but cause unknown: defensive code + extra log.

rburema:
277 crashes reported against 4.8 according to sentry -- but no indication as to what the _original_ matrix was that went in that causes this


nallath (approved):

Ghostkeeper:
I've reproduced this crash about 50%-ish when working with the automated Settings Guide screenshot tool, when it creates top-down screenshots. It's conceivable that it also reproduces with the top-down view button.

rburema:
Part if the reason I made this change was also so that we get some more info in the logs in case something does happen. So if it happens again, you should at least be able to see what matrix went in (in the logs), so we can actually attempt fix the underlying cause.

PR #693, opened 2021-04-19
Central File Storage

Ghostkeeper:
This is a pull request that adds new functionality to Uranium: Central File Storage. This functionality allows the programmer to store a file in a central location that is not dependent on the version number of the application he's running. It is a requirement for using plug-ins with large files, as the version upgrade system would copy those plug-ins each time the user upgrades the application. Using this central file storage, the programmer can move a big file out of the plug-in directory, reducing the file size of the plug-in directory.

There is a new class `CentralFileStorage` which collects two methods: `store` and `retrieve`. The `store` function stores a file in the central location, physically moving it away from the directories that get copied for each version. The `retrieve` function checks the file against a hash and then gives the path where the file is stored if that hash check is successful.

Requirements (from the ticket):
* There is a new function in Uranium to store big files in a shared location. This function takes a file path, a file version and an ID to reference the file by.
* If the file doesn't exist, the function exits cleanly without error. This way the plug-in can call the storing function simply at start-up without needing to check if this is the first time it was ran.
* The storage function needs to move the file to a central location which is not copied when the version upgrade runs. This central location needs to be part of the "data" storage location (not "config"). So in Windows it needs to be in %APPDATA%\Roaming\cura; in Linux it needs to be in ~/.local/share/cura; in MacOS it needs to be in ~/Library/Application Support/cura. This will be a sub-folder next to all of the folders for each Cura version.
* The storage function needs to rename the file to something which includes the file ID as well as the version number.
* The storage function needs to check if there is already a file with that file ID and version number in the central location. If there is and the files are the same, it simply deletes the new file and retains the old one. If there is and the files are NOT the same, it needs to raise an error.
* There needs to be a retrieval function which returns the file path of a file with a certain file ID and version number.
* This retrieval function needs to accept a file hash as well. Before returning the file path, it needs to check the file against that hash and if they don't match raise an error.

Implements CURA-8162.

### DIFF
--- a/UM/CentralFileStorage.py
+++ b/UM/CentralFileStorage.py
@@ -6,6 +6,7 @@ import os  # To remove duplicate files.
 import os.path  # To re-format files with their proper file extension but with a version number in between.
 import shutil  # To move files in constant-time.
 
+from UM.Logger import Logger
 from UM.Resources import Resources  # To get the central storage location.
 from UM.Version import Version  # To track version numbers of these files.
 
@@ -36,6 +37,7 @@ class CentralFileStorage:
         :param version: A version number for the file.
         """
         if not os.path.exists(file_path):
+            Logger.debug(f"{file_id} {str(version)} was already stored centrally.")
             return
         storage_path = cls._getFilePath(file_id, version)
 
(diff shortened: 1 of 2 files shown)


jellespijker (approved):
Looks good

PR #694, opened 2021-04-28
Cura 8010 icon size

jellespijker:
This PR adds new logic to the getIcons function allowing an icon to be stored in a subfolder. For instance different sizes. Such as is the standard within the UX department.

Since the used icons are renamed according to the conversion table and practice of the UXers this is also updated in the code.

Also, check PR ultimaker/cura#9716

### DIFF
--- a/UM/Qt/Bindings/Theme.py
+++ b/UM/Qt/Bindings/Theme.py
@@ -23,7 +23,7 @@ class Theme(QObject):
         self._engine = engine
         self._styles = None  # type: Optional[QObject]
         self._path = ""
-        self._icons = {}  # type: Dict[str, QUrl]
+        self._icons = {}  # type: Dict[str, Dict[str, QUrl]]
         self._images = {}  # type: Dict[str, QUrl]
 
         # Workaround for incorrect default font on Windows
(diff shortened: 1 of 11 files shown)


jellespijker:
tests are failing.
Will take a look in a minute

fieldOfView:
> For instance different sizes.

*Scalable* Vector Graphics.
Color me confused...

jellespijker:
Yes, UXers are strange....

Just talked with @konskarm , I will change them to: small, default and big

UM/Qt/Bindings/Theme.py
```diff
@@ -116,10 +116,14 @@ def getSize(self, size) -> QSizeF:
...
     @pyqtSlot(str, result = "QUrl")
-    def getIcon(self, icon_name: str) -> QUrl:
-        if icon_name in self._icons:
-            return self._icons[icon_name]
+    def getIcon(self, icon_name: str, px: str = "24") -> QUrl:
+        if px in self._icons:
+            if icon_name in self._icons[px]:
+                return self._icons[px][icon_name]
+            elif icon_name in self._icons["icons"]:  # Retrieve the "old" icon from the base icon folder
+                return self._icons["icons"][icon_name]
```
konskarm, line 126: I am not in favor of naming icons by their size (12, 24, 30 etc). I think it's best to use terms like small, default, medium, large and set the size using the sourceSize of the images.

konskarm (changes requested):

UM/Qt/Bindings/Theme.py
jellespijker: Automatically setting them to the source size, might be a bit overkill imo. I agree with the renaming to: small, default, large

jellespijker:
@fieldOfView Not yet used, but now you could simplify or add more detail to an icon depending on the size. Notice the detail in the printers at the back
image

fieldOfView:
Thanks for the explanation. The Cura PR made it more understandable since it came with a picture. Pictures help ;-)

PR #695, opened 2021-05-05
Add system to deprecate icons in the theme

nallath:
CURA-8010

### DIFF
--- a/UM/Qt/Bindings/Theme.py
+++ b/UM/Qt/Bindings/Theme.py
@@ -4,6 +4,7 @@
 import json
 import os
 import sys
+import warnings
 from typing import Dict, Optional, List
 
 from PyQt5.QtCore import QObject, pyqtProperty, pyqtSignal, QCoreApplication, QUrl, QSizeF
@@ -24,6 +25,7 @@ class Theme(QObject):
         self._styles = None  # type: Optional[QObject]
         self._path = ""
         self._icons = {}  # type: Dict[str, Dict[str, QUrl]]
+        self._deprecated_icons = {} # type: Dict[str, Dict[str, str]]
         self._images = {}  # type: Dict[str, QUrl]
 
         # Workaround for incorrect default font on Windows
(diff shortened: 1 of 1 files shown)


UM/Qt/Bindings/Theme.py
```diff
@@ -243,6 +253,16 @@ def load(self, path: str, is_first_call: bool = True) -> None:
...
                     self._icons[detail_level][name] = QUrl.fromLocalFile(os.path.join(base_path, icon))
 
+            deprecated_icons_file = os.path.join(iconsdir, "deprecated_icons.json")
+            if os.path.isfile(deprecated_icons_file):
+                try:
+                    with open(deprecated_icons_file, encoding="utf-8") as f:
+                        data = json.load(f)
+                        for icon in data:
+                            self._deprecated_icons[icon] = data[icon]
+                except:
```
Ghostkeeper, line 263: Probably better to just catch `EnvironmentError` here and `JSONDecodeError`. That way if we make a mistake like a `TypeError` or something, it would crash and tell us about it.

It's also a good idea to log the errors then, if we catch them.

UM/Qt/Bindings/Theme.py
```diff
@@ -135,6 +137,14 @@ def getIcon(self, icon_name: str, detail_level: str = "default") -> QUrl:
         elif icon_name in self._icons["icons"]:  # Retrieve the "old" icon from the base icon folder
             return self._icons["icons"][icon_name]
 
+        if icon_name in self._deprecated_icons:
+            new_icon = self._deprecated_icons[icon_name]["new_icon"]
+            warning = "The icon {old_icon_name} is deprecated. Please use {new_icon} instead.".format(old_icon_name = icon_name, new_icon = new_icon)
```
Ghostkeeper, line 142: We can actually use modern string formatting nowadays, new in Python 3.6!

```suggestion
            warning = f"The icon {icon_name} is deprecated. Please use {new_icon} instead."
```

Ghostkeeper (changes requested):

UM/Qt/Bindings/Theme.py
nallath: I keep forgetting this. Good call!

konskarm (approved):

PR #698, opened 2021-05-26
Update SettingDefinitionsModel.py

YuvalZilber:
_isAnyDescendantVisible didn't respect the "enabled" property of the child if defined.
it is of type "SettingFunction" and pass the if statement even if set to "False".


rburema:
Devs, please see internal ticket CURA-8288 as a back-ref.

rburema:
Thanks for your contribution @YuvalZilber !

rburema:
I'm sorry, I made a mistake while reviewing this. It seems this way throws errors when calling the formula, if not all the settings in the formula can be resolved.

To solve this you'd need information that is not available to the UM library (the active machine). Which could've been given as the optional argument to the evaluation call (but can't because that information would be hard to get at).

As such, we've decided to revert this for now.

PR #699, opened 2021-05-27
Revert "Update Icons"

nallath:
Because we simply don't have time to properly fix all of this for 4.10, we are reverting these changes for now. This revert will be reverted for the release after this. 

CURA-8267

https://github.com/Ultimaker/Cura/pull/9916


Ghostkeeper (approved):
Something is weird about how the icon names apparently changed in Uranium but the icons themselves are not in Uranium. More of an architectural problem and out of scope here. Just... interesting.

PR #701, opened 2021-05-31
Fix central storage with signed plugins

nallath:
CURA-8270

### DIFF
--- a/UM/PluginRegistry.py
+++ b/UM/PluginRegistry.py
@@ -627,7 +627,7 @@ class PluginRegistry(QObject):
                     final_location = loc
 
         # Move data (if any) to central storage
-        central_storage_file = os.path.join(final_location, plugin_id, Trust.getCentralStorageFilename())
+        central_storage_file = os.path.join(final_location, plugin_id, TrustBasics.getCentralStorageFilename())
         if os.path.exists(central_storage_file):
             try:
                 with open(central_storage_file, "r", encoding = "utf-8") as file_stream:
--- a/UM/PluginRegistry.py
+++ b/UM/PluginRegistry.py
@@ -626,23 +626,14 @@ class PluginRegistry(QObject):
                     highest_version = current_version
                     final_location = loc
 
-
         # Move data (if any) to central storage
         central_storage_file = os.path.join(final_location, plugin_id, "central_storage.json")
         if os.path.exists(central_storage_file):
             try:
                 with open(central_storage_file, "r", encoding="utf-8") as file_stream:
                     self._handleStoringFiles(file_stream.read(), os.path.join(final_location, plugin_id))
-
             except:
-                print("NOPE NOPE")
                 pass
-
-        '''if "files_for_central_storage" in plugin_metadata:
-            Logger.log("d", f"Moving files to central storage for {plugin_id}")
-            for file_path, file_id, version_string, hash in plugin_metadata["files_for_central_storage"]:
-                CentralFileStorage.store(file_path, file_id, Version(version_string))'''
-
         try:
             file, path, desc = imp.find_module(plugin_id, [final_location])
         except Exception:
--- a/UM/PluginRegistry.py
+++ b/UM/PluginRegistry.py
@@ -698,12 +698,15 @@ class PluginRegistry(QObject):
         """
         try:
             file_manifest = json.loads(file_data)
-        except json.decoder.JSONDecodeError:
+        except (json.decoder.JSONDecodeError, UnicodeDecodeError):
             Logger.logException("e", "Failed to parse central_storage.json")
             return
 
         for file_to_move in file_manifest:
-            CentralFileStorage.store(os.path.join(plugin_path, file_to_move[0]), file_to_move[1], Version(file_to_move[2]))
+            try:
+                CentralFileStorage.store(os.path.join(plugin_path, file_to_move[0]), file_to_move[1], Version(file_to_move[2]))
+            except (TypeError, IndexError):
+                Logger.logException("w", "Unable to move file to central storage")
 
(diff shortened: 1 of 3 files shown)


UM/PluginRegistry.py
```diff
@@ -624,6 +625,15 @@ def _findPlugin(self, plugin_id: str) -> Optional[types.ModuleType]:
                 if current_version > highest_version:
                     highest_version = current_version
                     final_location = loc
+
+        # Move data (if any) to central storage
+        central_storage_file = os.path.join(final_location, plugin_id, "central_storage.json")
+        if os.path.exists(central_storage_file):
+            try:
+                with open(central_storage_file, "r", encoding="utf-8") as file_stream:
```
Ghostkeeper, line 633: Code style: Spaces around binary operators.
```suggestion
                with open(central_storage_file, "r", encoding = "utf-8") as file_stream:
```

UM/PluginRegistry.py
```diff
@@ -678,6 +688,23 @@ def _locatePlugin(self, plugin_id: str, folder: str) -> Optional[str]:
 
         return None
 
+    def _handleCentralStorage(self, file_data: str, plugin_path: str) -> None:
+        """
+        Plugins can indicate that they want certain things to be stored in a central location.
+        In the case of a signed plugin, this is done by means of the central_storage.json file.
```
Ghostkeeper, line 694: It's also done this way in the case of a non-signed plug-in, right?

UM/Trust.py
```diff
@@ -386,6 +393,12 @@ def signedFolderCheck(self, path: str) -> bool:
 
         try:
             json_filename = os.path.join(path, TrustBasics.getSignaturesLocalFilename())
+            storage_filename = os.path.join(path, TrustBasics.getCentralStorageFilename())
+
+            storage_json = None
+            if os.path.exists(storage_filename):
+                with open(storage_filename, "r", encoding="utf-8") as data_file:
```
Ghostkeeper, line 400: Code style: Spaces around binary operators.
```suggestion
                with open(storage_filename, "r", encoding = "utf-8") as data_file:
```

UM/Trust.py
```diff
@@ -435,6 +448,20 @@ def signedFolderCheck(self, path: str) -> bool:
                         if os.path.islink(dir_full_path) and not self._follow_symlinks:
                             Logger.log("w", "Directory symbolic link '{0}' will not be followed.".format(dir_full_path))
 
+                # Check if the files moved to storage are still correct.
+                if storage_json:
+                    for entry in storage_json:
+                        try:
+                            # If this doesn't raise an exception, it's correct.
+                            CentralFileStorage.retrieve(entry[1], entry[3], Version(entry[2]))
+                        except:
```
Ghostkeeper, line 457: It would be better to catch the specific exceptions that you expect here. That way, if we change the code in the future, we'll get a correct error message when we've made a mistake.
```suggestion
                        except (EnvironmentError, IOError):
```

UM/Trust.py
```diff
@@ -435,6 +448,20 @@ def signedFolderCheck(self, path: str) -> bool:
...
+                            self._violation_handler("Centrally stored file '{0}' didn't match with checksum or it could not be found".format(entry[1]))
```
Ghostkeeper, line 458: Maybe use the string formatting again?

```suggestion
                            self._violation_handler(f"Centrally stored file '{entry[1]}' didn't match with checksum or it could not be found")
```

Ghostkeeper (changes requested):

Ghostkeeper:
It would also be good to document the file format of the JSON file, perhaps in the class documentation of the `CentralFileStorage` class. Like with an example.

UM/PluginRegistry.py
```diff
@@ -624,6 +625,15 @@ def _findPlugin(self, plugin_id: str) -> Optional[types.ModuleType]:
                 if current_version > highest_version:
                     highest_version = current_version
                     final_location = loc
+
+        # Move data (if any) to central storage
+        central_storage_file = os.path.join(final_location, plugin_id, "central_storage.json")
```
Ghostkeeper, line 630: Should use the file name from the `Trust` class?
```suggestion
        central_storage_file = os.path.join(final_location, plugin_id, Trust.getCentralStorageFilename())
```

Ghostkeeper (changes requested):
Additional problems found during testing:

1. The files are only moved to a central location when the files are verified by the Trust class. This gives trouble for a plug-in developer, since they must then effectively implement their side twice: A JSON file for the case when the files are verified by the Trust class, and a manual move for the case when the files are not verified. It would be more elegant to always read the JSON file, and have that be the canonical way to move files to a central storage for everyone, regardless of whether a plug-in is signed or not.
2. You'll probably want to catch `UnicodeDecodeError` from parsing the JSON file, and `TypeError` and `IndexError` when accessing the elements from it.

konskarm (commented):
I think that the name of the filed to-be-moved should not be altered with the version appended at the end, as this can be problematic when moving a library to the central storage. 

Maybe we can consider appending the version at the top level folder, if an entire folder is moved to the central storage.

Ghostkeeper:
Indeed, but that is out of scope for this PR (CURA-8270).

UM/PluginRegistry.py
nallath: You don't _have_ to do it via the central_storage.json file in that case, but yeah. I'll update the documentation

nallath:
I'm not quite sure what you mean by point one. The central_storage.json will work regardless if the plugin is signed or not.

Ghostkeeper (approved):
Changes look good.

I can't find any more what I did when testing this issue yesterday where I found that it only worked in Enterprise. I'll try it again.

PR #702, opened 2021-06-02
CURA-8261: Allow central storage to store folders

konskarm:
Plugins that make use of the central storage can now move an entire folder without having to mention each and every file in their `central_storage.json`. So now, when an entire folder is move, the `central_storage.json` can look like this:

```json
[
    [
        "my_folder",
        "my_folder",
        "1.0.0",
        "aa9125bc0121eaaa09125989832532c9c718a9b9e9ddd123"
    ]
]
```

CURA-8261

### DIFF
--- a/UM/CentralFileStorage.py
+++ b/UM/CentralFileStorage.py
@@ -62,9 +62,9 @@ class CentralFileStorage:
         if os.path.exists(storage_path):  # File already exists. Check if it's the same.
             if os.path.getsize(path) != os.path.getsize(storage_path):  # As quick check if files are the same, check their file sizes.
                 raise FileExistsError(f"Central file storage already has an item (file or directory) with ID {path_id} and version {str(version)}, but it's different.")
-            new_file_hash = cls._hashItem(path)
-            stored_file_hash = cls._hashItem(storage_path)
-            if new_file_hash != stored_file_hash:
+            new_item_hash = cls._hashItem(path)
+            stored_item_hash = cls._hashItem(storage_path)
+            if new_item_hash != stored_item_hash:
                 raise FileExistsError(f"Central file storage already has an item (file or directory) with ID {path_id} and version {str(version)}, but it's different.")
             if os.path.isfile(path):
                 os.remove(path)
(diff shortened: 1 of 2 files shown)


UM/CentralFileStorage.py
```diff
@@ -34,59 +35,62 @@ class CentralFileStorage:
...
         """
         Store a new file into the central file storage. This file will get moved to a storage location that is not
         specific to this version of the application.
 
         If the file already exists, this will check if it's the same file. If the file is not the same, it raises a
         `FileExistsError`. If the file is the same, no error is raised and the file to store is simply deleted. It is a
         duplicate of the file already stored.
-        :param file_path: The path to the file to store in the central file storage.
-        :param file_id: A name for the file to store.
+        :param path: The path to the file to store in the central file storage.
```
nallath, line 46: That's not quite correct. You can store a file _or_ a folder now right? (make sure to update that in all other params as well)

UM/CentralFileStorage.py
```diff
@@ -34,59 +35,62 @@ class CentralFileStorage:
...
+        storage_path = cls._getFilePath(path_id, version)
 
         if os.path.exists(storage_path):  # File already exists. Check if it's the same.
-            if os.path.getsize(file_path) != os.path.getsize(storage_path):  # As quick check if files are the same, check their file sizes.
-                raise FileExistsError(f"Central file storage already has a file with ID {file_id} and version {str(version)}, but it's different.")
-            new_file_hash = cls._hashFile(file_path)
-            stored_file_hash = cls._hashFile(storage_path)
+            if os.path.getsize(path) != os.path.getsize(storage_path):  # As quick check if files are the same, check their file sizes.
+                raise FileExistsError(f"Central file storage already has a file with ID {path_id} and version {str(version)}, but it's different.")
+            new_file_hash = cls._hashPath(path)
```
nallath, line 63: Perhaps rename this to "new_item_hash" as well (since it might not be a file but a dir?)

nallath (changes requested):

UM/CentralFileStorage.py
konskarm: Good catch, fixed.

nallath (approved):

PR #703, opened 2021-06-04
Disable central storage for bundled plugins

nallath:
CURA-8284


UM/CentralFileStorage.py
```diff
@@ -5,101 +5,126 @@
...
+        ["relative/path/to/dir", "relative/path/to/dir/in/storage", "1.0.1, "124986978cb21e"]
     ]
     """
 
+    # In some cases a plugin might ask for files to be moved, but it's not needed (since the plugin is actually bundled
+    # In order to ensure that the same API can still be used, we store those situations.
+    _unmoved_files = {}  # type: Dict[str, str]
+
     @classmethod
-    def store(cls, file_path: str, file_id: str, version: Version = Version("1.0.0")) -> None:
```
Ghostkeeper, line 37: You renamed `file_id` to `path_id`, but I think you misunderstood that parameter. The file ID is only a name of the item. It's not necessarily the same as the file name or file path (since files might be renamed in later versions).

A better name for it would be `item_id`, since that is how you are calling it in the documentation now.

UM/CentralFileStorage.py
```diff
@@ -5,101 +5,126 @@
...
 
     A brief example of such a file:
     [
-        ["files/VeryLargeFileToStore.big", "very_large_file", "1.0.1", "abcdefghijklmnop"]
+        ["files/VeryLargeFileToStore.big", "very_large_file", "1.0.1", "abcdefghijklmnop"],
+        ["relative/path/to/dir", "relative/path/to/dir/in/storage", "1.0.1, "124986978cb21e"]
     ]
     """
 
+    # In some cases a plugin might ask for files to be moved, but it's not needed (since the plugin is actually bundled
```
Ghostkeeper, line 39:
```suggestion
    # In some cases a plugin might ask for files to be moved, but it's not needed (since the plugin is actually bundled)
```

UM/CentralFileStorage.py
```diff
@@ -5,101 +5,126 @@
...
+        :param sha256_hash: A SHA-256 hash of the item (file or directory) you expect to receive from the central storage.
+        :param version: The version number of the item (file or directory) to retrieve.
+        :return: A path to the location of the centrally stored item (file or directory).
+        :raises FileNotFoundError: There is no item (file or directory) stored with that name and version.
+        :raises IOError: The hash of the item (file or directory) is incorrect. Opening this item could be a security
+        risk.
         """
-        storage_path = cls._getFilePath(file_id, version)
+        if path_id + str(version) in cls._unmoved_files:
+            return cls._unmoved_files[path_id + str(version)]
```
Ghostkeeper, line 106: This is a security risk, because it is returning a path here without checking the hash. An attacker could replace the file in this position with a malicious file, and it would be trusted.

I suggest you modify the `_getItemPath` function to include this case, and then execute the rest of the function as normal.

Ghostkeeper (changes requested):

UM/CentralFileStorage.py
nallath: That's not true. If a file is not moved it is done because it's bundled (and thus in the same location as where Cura is installed). We skip all checks for bundled plugins, so we should also be able to skip this as far as I can see.

Ghostkeeper (approved):

UM/CentralFileStorage.py
Ghostkeeper: As discussed, this is indeed not an attack vector if the `store` function is used correctly.

Ghostkeeper:
I tested this by adding the following `central_storage.json` file to the Settings Guide:

```
[
	["example.png", "example", "1.0.0", "666332308da6ada9eac67d70fc3810d3f85e63450323c6f13c9db6aba5e19769"]
]
```

The Settings Guide in my computer is located in the source folder for Cura (in its `/plugins` folder among all the other bundled plug-ins). This would normally make the example.png file move to the central storage, and indeed it does on the Master branch, but not on this branch. So it is indeed correctly not moving files to central storage for bundled plug-ins.

I also tested moving the Settings Guide to the normal `plugins` folder in the configuration folder and performing the same test. It then did correctly move the `example.png` file to central storage.

So this works correctly according to those tests!

PR #704, opened 2021-06-08
CURA-8291: Verbose plugin loading on startup

konskarm:
This PR changes the following:

- To make sure that the splash screen's loading wheel is not stuck while plugins are being unpacked, the unzipping of a downloaded plugin by the PackageManager now happens file-by-file. After each extracted file, the package manager calls the `processEvents()` function, which ensures that the interface gets updated (and the loading wheel keeps turning instead of getting stuck).
- Same thing now happens while the CentralFileStorage is hashing files, because hashing can take a long time and give the idea that the splash screen is stuck.
- The splash screen now displays a message regarding which plugin is currently being unpacked and which plugin is being loaded. This way big plugins don't give the idea that the startup is stuck.

Requires https://github.com/Ultimaker/Cura/pull/9948.

CURA-8291

### DIFF
--- a/UM/CentralFileStorage.py
+++ b/UM/CentralFileStorage.py
@@ -6,6 +6,7 @@ import os  # To remove duplicate files.
 import os.path  # To re-format files with their proper file extension but with a version number in between.
 import shutil  # To move files in constant-time.
 from typing import List, Tuple, Dict
+from PyQt5.QtCore import QCoreApplication
 
 from UM.Logger import Logger
 from UM.Resources import Resources  # To get the central storage location.
@@ -148,6 +149,7 @@ class CentralFileStorage:
             while len(contents) > 0:
                 hasher.update(contents)
                 contents = f.read(block_size)
+        QCoreApplication.processEvents()  # Process events to allow the interface to update
         return hasher.hexdigest()
 
     @classmethod
--- a/UM/CentralFileStorage.py
+++ b/UM/CentralFileStorage.py
@@ -41,6 +41,8 @@ class CentralFileStorage:
     # In order to ensure that the same API can still be used, we store those situations.
     _unmoved_files = {}  # type: Dict[str, str]
 
+    _is_enterprise_version = False
+
     @classmethod
     def store(cls, path: str, path_id: str, version: Version = Version("1.0.0"), move_file: bool = True) -> None:
         """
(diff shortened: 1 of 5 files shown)


UM/CentralFileStorage.py
```diff
@@ -148,6 +149,7 @@ def _hashFile(cls, file_path: str) -> str:
             while len(contents) > 0:
                 hasher.update(contents)
                 contents = f.read(block_size)
+        QCoreApplication.processEvents()  # Process events to allow the interface to update
```
nallath, line 152: Alternatively; you could also call the self._application.showSplashMessage, since that ensures that processEvents is being called.

UM/PluginRegistry.py
```diff
@@ -352,6 +352,10 @@ def isBundledPlugin(self, plugin_id: str) -> bool:
         self._bundled_plugin_cache[plugin_id] = is_bundled
         return is_bundled
 
+    # Indicates that a specific plugin is currently being loaded. If the plugin_id is empty, it means that no plugin
+    # is currently being loaded.
+    pluginLoadingInProgress = pyqtSignal(str, arguments = ["plugin_id"])
```
nallath, line 357: It's a signal, so it should notify a change, not a state.

So this should be changed to pluginLoadStarted or something. If you want something that keeps track of the state, i would argue that it's a property ("pluginLoadingInProgress" makes more sense as a property)

UM/Qt/QtApplication.py
```diff
@@ -186,6 +186,12 @@ def initialize(self) -> None:
         Logger.log("i", "Initializing version upgrade manager ...")
         self._version_upgrade_manager = VersionUpgradeManager(self)
 
+    def _displayLoadingPluginSplashMessage(self, plugin_id: Optional[str]) -> None:
+        message = i18nCatalog("uranium").i18nc("@info:progress", "Loading plugins...")
+        if plugin_id:
+            message = f"{i18nCatalog('uranium').i18nc('@info:progress', 'Loading plugin')} {plugin_id}..."
```
nallath, line 192: Again; i'm not really sure if i'm a fan of this way of getting handling the translations.

I personally think that `i18nCatalog('uranium').i18nc('@info:progress', 'Loading plugin {plugin_id}').format(plugin_id=plugin_id)` is much more readable.

nallath (changes requested):
See comments

rburema (approved):

PR #706, opened 2021-06-15
[CURA-8320] Update SDK/API from 7.5.0 to 7.6.0 for 4.10

rburema:
Will amend this description with the changes shortly, as well as (attempt to) mail our plugin contributors.

In Uranium (so excluding Cura, see the other PR for that):

**new classes, their methods**

- Uranium/UM/CentralFileStorage
  _This class stores and retrieves items (files or directories) stored in a location central to all versions of the application. Its purpose is to provide a way for plug-ins to store very big items (files or directories) without having those big items copied to the new resource directory at every version upgrade._
    - `[classmethod] def store(cls, path: str, path_id: str, version: Version = Version("1.0.0"), move_file: bool = True) -> None`
	  _Store a new item (file or directory) into the central file storage. This item will get moved to a storage location that is not specific to this version of the application._
	- `[classmethod] def retrieve(cls, path_id: str, sha256_hash: str, version: Version = Version("1.0.0")) -> str`
	  _Retrieve the path of a previously stored item (file or directory)._
	- `[classmethod] def setIsEnterprise(cls, is_enterprise: bool) -> None`
	  _Some things central storage does will only (not) need to happen for enterprise editions of Cura. As of writing, this is only used to do a security check against the hash-values of the files._

**added method signatures**

- in Uranium/UM/FileHandler/FileHandler:
  `resolveAnySymlink(file_name: str) -> str`
  _Some linked files (.lnk in Windows for example) aren't always automatically resolved to their base. This method resolves to what they're pointing at, given the link filename._

- in Uranium/UM/Trust/Trust:
  `[classmethod] TrustBasics.getCentralStorageFilename(cls) -> str`
  _See 'CentralFileStorage' in 'new classes and methods'._

**added optional parameters and/or typechecking**

- in Uranium/UM/Bindings/Theme: Added 'detail_level' to
  `Theme.getIcon [getIcon(self, icon_name: str, detail_level: str = "default") -> QUrl]`
  _detail_level: The level of detail of the icon. Choice between "low, "default", "medium", "high"._

- in Uranium/UM/View/GL/ShaderProgram: Added typechecking to
  `ShaderProgram.setUniformValue [setUniformValue(self, name: str, value: Union[Vector, Matrix, Color, List[float], List[List[float]], float, int], **kwargs: Any) -> None]`

See also: https://github.com/Ultimaker/Cura/pull/9982

### DIFF
--- a/tests/Settings/ContainerTestPlugin/plugin.json
+++ b/tests/Settings/ContainerTestPlugin/plugin.json
@@ -1,5 +1,5 @@
 {
     "name": "TestContainerPlugin",
-    "api": 7,
+    "api": "7.6.0",
     "version": "1.0.0"
 }
--- a/plugins/ConsoleLogger/plugin.json
+++ b/plugins/ConsoleLogger/plugin.json
@@ -3,6 +3,6 @@
     "author": "Ultimaker B.V.",
     "version": "1.0.1",
     "description": "Outputs log information to the console.",
-    "api": "7.5.0",
+    "api": "7.6.0",
     "i18n-catalog": "uranium"
 }
--- a/plugins/FileHandlers/OBJReader/plugin.json
+++ b/plugins/FileHandlers/OBJReader/plugin.json
@@ -3,6 +3,6 @@
     "author": "Ultimaker B.V.",
     "version": "1.0.1",
     "description": "Makes it possible to read Wavefront OBJ files.",
-    "api": "7.5.0",
+    "api": "7.6.0",
     "i18n-catalog": "uranium"
 }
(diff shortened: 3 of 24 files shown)


tests/Settings/ContainerTestPlugin/plugin.json
```diff
@@ -1,5 +1,5 @@
 {
     "name": "TestContainerPlugin",
-    "api": 7,
+    "api": "7.6.0",
```
Ghostkeeper, line 3: This is incorrect. The `"api"` field needs to be an integer.

The field is basically deprecated, replaced by `"supported_sdk_versions"`. But we can't remove them since otherwise Cura 3.6 and earlier would accept it as a plug-in with fitting API.

```suggestion
    "api": 7,
```

Ghostkeeper (changes requested):

tests/Settings/ContainerTestPlugin/plugin.json
rburema: Thanks! It's good to pick up tasks that don't always fall to me so now and then :-)

PR #709, opened 2021-06-19
Updated pt_BR strings for Cura 4.10

Patola:


jellespijker:
Looks good @Patola , thanks very much appreciated

PR #712, opened 2021-07-23
CURA-7278: Differentiate warning messages in cura

konskarm:
Now the messages contain a message type, which can be one of the following:
1. POSITIVE
2. NEUTRAL (default)
3. WARNING
4. NEGATIVE

The message types determine what icon will appear next to the title of the message. 

image
image

Requires https://github.com/Ultimaker/Cura/pull/10175.

CURA-7278

### DIFF
--- a/UM/Message.py
+++ b/UM/Message.py
@@ -22,9 +22,15 @@ class Message(QObject):
         ALIGN_LEFT = 2
         ALIGN_RIGHT = 3
 
+    class MessageType:
+        CONFIRMATION = 0
+        INFORMATION = 1
+        WARNING = 2
+        FAILURE = 3
+
     def __init__(self, text: str = "", lifetime: int = 30, dismissable: bool = True, progress: float = None,
                  title: Optional[str] = None, parent=None, use_inactivity_timer: bool = True, image_source: str = "",
-                 image_caption: str = "", option_text: str = "", option_state: bool = True) -> None:
+                 image_caption: str = "", option_text: str = "", option_state: bool = True, message_type: int = MessageType.INFORMATION) -> None:
 
         """Class for displaying messages to the user.
         Even though the lifetime can be set, in certain cases it can still have a lifetime if nothing happens with the
@@ -75,6 +81,8 @@ class Message(QObject):
 
         self._actions = []  # type: List[Dict[str, Union[str, int]]]
         self._title = title
+
+        self._message_type = message_type
 
         self.actionTriggered.connect(self._onActionTriggered)
 
@@ -195,6 +203,14 @@ class Message(QObject):
 
     def getImageCaption(self) -> str:
         return self._image_caption
+
+    def getMessageType(self) -> int:
+        """
+        Gets the type of the message.
+        The message gets a different icon according to its type.
+        :return: The type of the message (CONFIRMATION, INFORMATION, WARNING, FAILURE)
+        """
+        return self._message_type
 
     def setText(self, text: str) -> None:
         """Changes the text on the message.
--- a/UM/Message.py
+++ b/UM/Message.py
@@ -12,7 +12,6 @@ from UM.Signal import Signal, signalemitter
 class Message(QObject):
     """Class for displaying messages to the user."""
 
-
     class ActionButtonStyle:
         DEFAULT = 0
         LINK = 1
@@ -23,14 +22,14 @@ class Message(QObject):
         ALIGN_RIGHT = 3
 
     class MessageType:
-        CONFIRMATION = 0
-        INFORMATION = 1
+        POSITIVE = 0
+        NEUTRAL = 1
         WARNING = 2
-        FAILURE = 3
+        ERROR = 3
 
     def __init__(self, text: str = "", lifetime: int = 30, dismissable: bool = True, progress: float = None,
                  title: Optional[str] = None, parent=None, use_inactivity_timer: bool = True, image_source: str = "",
-                 image_caption: str = "", option_text: str = "", option_state: bool = True, message_type: int = MessageType.INFORMATION) -> None:
+                 image_caption: str = "", option_text: str = "", option_state: bool = True, message_type: int = MessageType.NEUTRAL) -> None:
 
         """Class for displaying messages to the user.
         Even though the lifetime can be set, in certain cases it can still have a lifetime if nothing happens with the
@@ -208,7 +207,7 @@ class Message(QObject):
         """
         Gets the type of the message.
         The message gets a different icon according to its type.
-        :return: The type of the message (CONFIRMATION, INFORMATION, WARNING, FAILURE)
+        :return: The type of the message (POSITIVE, NEUTRAL, WARNING, ERROR)
         """
         return self._message_type
 
--- a/UM/Message.py
+++ b/UM/Message.py
@@ -51,9 +51,9 @@ class Message(QObject):
         really, it's up to the QML to handle that).
         :param progress: Is there any progress to be displayed? if -1, it's seen
         as indeterminate.
-        :param message_type: Defines the type of message (POSITIVE, NEUTRAL, WARNING, ERROR, default: NEUTRAL).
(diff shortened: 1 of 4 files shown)


UM/Message.py
```diff
@@ -38,14 +43,17 @@ def __init__(self, text: str = "", lifetime: int = 30, dismissable: bool = True,
...
         :param title: Phrase that will be shown above the message.
         :param image_source: an absolute path where an image can be found to be
         displayed (QUrl.toLocalFile()) can be used for that.
         :param image_caption: Text to be displayed below the image (or anywhere
         really, it's up to the QML to handle that).
         :param progress: Is there any progress to be displayed? if -1, it's seen
         as indeterminate.
+        :param message_type: Defines the type of message (POSITIVE, NEUTRAL, WARNING, ERROR, default: NEUTRAL).
+                            According to the type, an icon appears next to the message title. The NEUTRAL messages
+                            contain no icon.
```
Ghostkeeper, line 56: Indentation is inconsistent with the rest of the params here.

Ghostkeeper: Should we document where to get this from? In the type it just says "int", and this documentation doesn't mention that it should be a member of the `MessageType` class.

UM/Qt/qml/UM/MessageStack.qml
```diff
@@ -91,16 +92,122 @@ ListView
...
             }
 
-            height: childrenRect.height
+            height: Math.max(messageTypeIcon.height, messageTitle.height)
+            Item
+            {
+                id: messageTypeIcon
+                visible: messageIcon.source != ""
+                height: UM.Theme.getSize("small_button_icon").height
+                width: visible ? UM.Theme.getSize("small_button_icon").height : 0
```
Ghostkeeper, line 101: This is not a button, nor should it align with any button. It seems appropriate to use `small_button_icon` here, then?

UM/Qt/qml/UM/MessageStack.qml
```diff
@@ -91,16 +92,122 @@ ListView
...
+                    },
+                    State
+                    {
+                        name: "warning"
+                        when: model.message_type == 2
+                        PropertyChanges
+                        {
+                            target: messageIcon
+                            source: UM.Theme.getIcon("Warning", "low")
+                            color: UM.Theme.getColor("message_warning_icon")
```
Ghostkeeper, line 160: We already have a "warning" colour. Should we not enforce that colour to be used for all warnings, so that the warning colours are consistent everywhere?
Rather than hard-coding specific theme entries for specific details in the interface, the design philosophy is to have re-usable theme entries as much as possible.

UM/Qt/qml/UM/MessageStack.qml
```diff
@@ -91,16 +92,122 @@ ListView
...
+                        }
+                        PropertyChanges
+                        {
+                            target: messageIconBackground
+                            color: UM.Theme.getColor("message_warning_icon_background")
```
Ghostkeeper, line 165: Rather than hard-coding specific theme entries for specific details in the interface, the design philosophy is to have re-usable theme entries as much as possible.
A better theme entry could be something like `warning_contrast` or `warning_background`. Although in this case it appears to be more like a foreground than a background.

UM/Qt/qml/UM/MessageStack.qml
```diff
@@ -91,16 +92,122 @@ ListView
...
+                    },
+                    State
+                    {
+                        name: "error"
+                        when: model.message_type == 3
+                        PropertyChanges
+                        {
+                            target: messageIcon
+                            source: UM.Theme.getIcon("Cancel", "low")
+                            color: UM.Theme.getColor("message_error_icon")
```
Ghostkeeper, line 176: We already have an "error" colour. Should we not enforce that colour to be used for all errors, so that the error colours are consistent everywhere?
Rather than hard-coding specific theme entries for specific details in the interface, the design philosophy is to have re-usable theme entries as much as possible.

UM/Qt/qml/UM/MessageStack.qml
```diff
@@ -91,16 +92,122 @@ ListView
...
+                [
+                    State
+                    {
+                        name: "positive"
+                        when: model.message_type == 0
+                        PropertyChanges
+                        {
+                            target: messageIcon
+                            source: UM.Theme.getIcon("Check", "low")
+                            color: UM.Theme.getColor("message_success_icon")
```
Ghostkeeper, line 129: Rather than hard-coding specific theme entries for specific details in the interface, the design philosophy is to have re-usable theme entries as much as possible.
So perhaps a better name for this entry would be to use simply `success`, to fit in with the similar status colour names "warning" and "error".

UM/Qt/qml/UM/MessageStack.qml
```diff
@@ -91,16 +92,122 @@ ListView
...
+                        }
+                        PropertyChanges
+                        {
+                            target: messageIconBackground
+                            color: UM.Theme.getColor("message_success_icon_background")
```
Ghostkeeper, line 134: Rather than hard-coding specific theme entries for specific details in the interface, the design philosophy is to have re-usable theme entries as much as possible.
A better theme entry could be something like `success_contrast` (to indicate that it should contrast with the "success" theme entry) or `success_background`. Although in this case it appears more like a foreground than a background.

UM/Qt/qml/UM/MessageStack.qml
```diff
@@ -91,16 +92,122 @@ ListView
...
+                        PropertyChanges
+                        {
+                            target: messageIcon
+                            source: UM.Theme.getIcon("Cancel", "low")
+                            color: UM.Theme.getColor("message_error_icon")
+                        }
+                        PropertyChanges
+                        {
+                            target: messageIconBackground
+                            color: UM.Theme.getColor("message_error_icon_background")
```
Ghostkeeper, line 181: Rather than hard-coding specific theme entries for specific details in the interface, the design philosophy is to have re-usable theme entries as much as possible.
A better theme entry could be something like `error_contrast` or `error_background`. Although in this case it appears to be more like a foreground than a background.

UM/Qt/qml/UM/MessageStack.qml
```diff
@@ -91,16 +92,122 @@ ListView
...
+                ]
+            }
+
+
+            Label
+            {
+                id: messageTitle
+                // Account for the left and right margins of the titleBar since they are not automatically accounted
+                // for using the width - x
+                width: parent.width - x - 2 * UM.Theme.getSize("default_margin").width
```
Ghostkeeper, line 193: Rather than calculating the width, it seems easier and more maintainable to use a `RowLayout` here instead of a `Row`.

Ghostkeeper (changes requested):

UM/Message.py
konskarm: Fixed in 6366138fdc2e7672e14a95b8559bffccba197039.

UM/Qt/qml/UM/MessageStack.qml
konskarm: Changed to it's own theme entry.

UM/Qt/qml/UM/MessageStack.qml
konskarm: I missed the already existing warning color. It is now using that instead of defining its own.

UM/Qt/qml/UM/MessageStack.qml
konskarm: I missed the already existing error color. It is now using that instead of defining its own.

UM/Qt/qml/UM/MessageStack.qml
konskarm: There was no success color so I added it as a generic one. The message is also using that now.

UM/Qt/qml/UM/MessageStack.qml
konskarm: The titleBar was changed to a RowLayout.

PR #713, opened 2021-07-23
[SEC-240] Add Pre-Storage Check To Trust

rburema:
In enterprise scenario's:

Before moving the files, run a quicker security check on the central storage file itself, since it could be potentially used to copy files from an elevated account (even if the plugin itself isn't run yet).

To prevent this, two things where done;
- one is that it's checked that the file that's being moved is (to) _within_ the plugin folder.
- The other is a quick 'pre-check' that doesn't scan the whole folder but only checks the integrity of the central storage file (and the manifest). This should, (and now _can_) be done _before_ moving however.

P.S. Internal ticket is also cloned as CURA-8407

### DIFF
--- a/UM/Trust.py
+++ b/UM/Trust.py
@@ -1,4 +1,4 @@
-# Copyright (c) 2020 Ultimaker B.V.
+# Copyright (c) 2021 Ultimaker B.V.
 # Uranium is released under the terms of the LGPLv3 or higher.
 
 import base64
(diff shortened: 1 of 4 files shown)


UM/Trust.py
```diff
@@ -384,6 +384,81 @@ def _verifyFile(self, filename: str, signature: str) -> bool:
...
+        since the central storage system (which otherwise runs first) copies files, and the copying of files itself
+        can be an attack. Note that right after copying, a full check can be done, so the files themselves don't have
+        to be checked yet (since that happens in the full check after copying).
+
+        Shared pools of versioned items ('central storage') are used if a folder contains a 'central storage file'.
+        (See the CentralFileStorage class for details.) The 'canonical' version of a folder as far as the Trust system
+        is concerned, is the one where the items are already in central storage. Otherwise there would either be a whole
+        range of 'acceptable answers' (and that's harder to test against) or, there would always be a need to verify
+        a situation that shouldn't be that frequent (items need to be copied to central storage). This creates a problem
+        in that the central storage mechanism needs to run _first_, however. The central storage has it's own security
```
Ghostkeeper, line 410:
```suggestion
        in that the central storage mechanism needs to run _first_, however. The central storage has its own security
```

UM/Trust.py
```diff
@@ -384,6 +384,81 @@ def _verifyFile(self, filename: str, signature: str) -> bool:
...
+        try:
+            # Check if the central storage file exist, if not, then the system won't copy anything in any case.
+            central_storage_filename = os.path.join(path, TrustBasics.getCentralStorageFilename())
+            if not os.path.exists(central_storage_filename):
+                Logger.log("i", f"No central storage file for unbundled folder '{path}'.")
+                return True
+
+            # Open the file containing signatures (just reading the json is negligible compared to the verify or store):
+            json_filename = os.path.join(path, TrustBasics.getSignaturesLocalFilename())
+            with open(json_filename, "r", encoding="utf-8") as data_file:
```
Ghostkeeper, line 431: Code style: Spaces around binary operators.
```suggestion
            with open(json_filename, "r", encoding = "utf-8") as data_file:
```

UM/Trust.py
```diff
@@ -384,6 +384,81 @@ def _verifyFile(self, filename: str, signature: str) -> bool:
...
+
+        try:
+            # Check if the central storage file exist, if not, then the system won't copy anything in any case.
+            central_storage_filename = os.path.join(path, TrustBasics.getCentralStorageFilename())
+            if not os.path.exists(central_storage_filename):
+                Logger.log("i", f"No central storage file for unbundled folder '{path}'.")
+                return True
+
+            # Open the file containing signatures (just reading the json is negligible compared to the verify or store):
+            json_filename = os.path.join(path, TrustBasics.getSignaturesLocalFilename())
```
Ghostkeeper, line 430: Could we give these a bit more descriptive variable names? It's quite a blur right now. Something like:

`json_filename` -> `signatures_filename` or `signatures_path`
`data_file` -> `signatures_file`
`json_data` -> `signatures_data` or `signatures_contents`
`signatures_json` -> `root_signature_category` (it's not JSON any more at this point, just a `dict`).
`name_in_data` -> `central_storage_filename`

Feel free to come up with different names of course. But the point is that the variables describe what they contain, not just the data type. That would make this segment a lot easier to understand.

UM/Trust.py
```diff
@@ -452,17 +522,22 @@ def signedFolderCheck(self, path: str) -> bool:
...
+
                             # If a directory was moved, add all the files in that directory to the file_count. For
                             # individual files mentioned in the central_storage.json increment the file_count by 1.
                             if os.path.isdir(central_storage_path):
                                 file_count += sum([len(files) for _, _, files in os.walk(central_storage_path)])
                             elif os.path.isfile(central_storage_path):
                                 file_count += 1
-                        except (EnvironmentError, IOError):
-                            self._violation_handler(f"Centrally stored file '{entry[1]}' didn't match with checksum or it could not be found")
+                        except (EnvironmentError, IOError, OSError):
```
Ghostkeeper, line 538: `OSError` is an alias of `EnvironmentError` so this change does nothing.

Ghostkeeper (changes requested):

UM/Trust.py
rburema: Resolved in 6f7dadba

UM/Trust.py
rburema: Resolved in 6f7dadba

Ghostkeeper (approved):
Better. Thanks! I'll try testing it now. Rather complex to test though!

Issue #664, opened 2021-01-13
Profile Rendering


marcus-bigrep:
I deal with big data and some times preview rendering is here quite slow < 1 FPS.
Thus, I have now done some profiling and results are strange.
Take attached 3mf and code adjustments at the bottom of the issue. Then slice the sample and rendering the preview of the gcode.
The log states that (preview) rendering takes <0.3sec but I get >1sec if I manually stop the time with a stop watch (from "start rendering" to "end rendering"). In particular, >1sec is the real rendering time.
Do you know where the time is spent?

https://github.com/Ultimaker/Uranium/blob/master/UM/Qt/Bindings/MainWindow.py
```
    import time

    def _render(self):
        tstart = time.perf_counter()
        Logger.log("w", "start rendering")

        if self._full_render_required:
            renderer = self._app.getRenderer()
            view = self._app.getController().getActiveView()
            renderer.beginRendering()
            view.beginRendering()
            renderer.render()
            view.endRendering()
            renderer.endRendering()
            self._full_render_required = False
            self.renderCompleted.emit()
        else:
            self._app.getRenderer().reRenderLast()

        tend = time.perf_counter()
        Logger.log("w", "end rendering %s", str(tend - tstart))

```

[UMS5_BigRepManifold.zip](https://github.com/Ultimaker/Uranium/files/5808861/UMS5_BigRepManifold.zip)

nallath:
I don't really understand how you timed this. When profiling code like this, I'd recommend using something like https://github.com/pyutils/line_profiler since it shows per line how much time was spent on it.

marcus-bigrep:
Thanks for the advice, I'll do it as well.
Though, I don't think it is about digging deeper in this function.
It seems that there is something in the message handling and Qt which takes a lot of time. And this is a big black hole, I don't understand.
In particular, it is not about this code. This code is fast.

_**Bottom line of this investigation is:**_
Log message is performed asynchronously and has a latency of >0.7sec.
Especially, the question is what happens in the 0.7sec? Do you have an idea?

nallath:
I'm not sure what part of the messaging that could be. It has to be something with opengl since the rendering is so much faster if you only display a few lines. 

But i do have the feeling that you might be on to something here that I missed the last few times that I tried profiling. Problem is; I don't know what I missed...

marcus-bigrep:
Just why I stumbled about it. I hacked in some caching of VBOs.
[...]

PR #714, opened 2021-07-26
Graphic Buffers Update

bremco:
The highly interdependent Issues this fixes:
- The Vertex Array Object (VAO) for a render wasn't actually being (re)used, but regenerated each time. Now it's cached.
- While (otherwise) vertex- and index-buffers where already cached for normal rendering, the index-buffer for a mesh was recreated each time only a _part_ of the mesh (a range) needed to be drawn (for example, Layer and Simulation view in Cura). (NB: This was due to a workaround on our side, caused by an only _partial_ implementation of a specific bit of the OpenGL API by the Python interface we're using.) This is now fixed.

Complications:
- In order to show 'ranged'/partial meshes, implementing programs (like Cura) must now have shaders that take a draw-range uniform, which can be compared against the newly introduced 'vertex-index' attribute. (As we can't use the OpenGL API due to partial implementation, and we can't use the `gl_primitiveID` built-in due to support of legacy shaders.) See also https://github.com/Ultimaker/Cura/pull/10188, **the front-end implementation of this PR.**

See also this related discussion https://github.com/Ultimaker/Uranium/issues/664 earlier.


nallath:
Do you also have before / after profiling? I'm curious what the gains are :)

bremco:
@nallath Right, was one of the things I meant to do as well :-)
Not sure I can get that done today though.

(P.S. Even if there are no true gains I'd argue this is more correct, but that's harder to justify from a 'risk' perspective. Also technically upping the SDK with my shader shenanigans.)

UM/View/RenderBatch.py
```diff
@@ -234,43 +229,7 @@ def render(self, camera: Optional[Camera]):
...
-            index_buffer = OpenGL.getInstance().createIndexBuffer(mesh)
-        else:
-            # glDrawRangeElements does not work as expected and did not get the indices field working..
-            # Now we're just uploading a clipped part of the array and the start index always becomes 0.
-            index_buffer = OpenGL.getInstance().createIndexBuffer(
-                mesh, force_recreate=True, index_start = self._render_range[0], index_stop = self._render_range[1])
-        if index_buffer is not None:
-            index_buffer.bind()
-
+    def _setMeshAttributes(self, mesh: Any):
```
nallath, line 232:
```suggestion
    def _setMeshAttributes(self, mesh: Any) -> None:
```
Partially typed functions are not checked by type checking.

UM/View/RenderBatch.py
```diff
@@ -302,21 +261,77 @@ def _renderItem(self, item: Dict[str, Any]):
...
-                else:
-                    self._gl.glDrawElements(self._render_mode, mesh.getFaceCount(), self._gl.GL_UNSIGNED_INT, None)
-            else:
-                if self._render_mode == self.RenderMode.Triangles:
-                    self._gl.glDrawRangeElements(self._render_mode, self._render_range[0], self._render_range[1], self._render_range[1] - self._render_range[0], self._gl.GL_UNSIGNED_INT, None)
-                else:
-                    self._gl.glDrawElements(self._render_mode, self._render_range[1] - self._render_range[0], self._gl.GL_UNSIGNED_INT, None)
-        else:
-            self._gl.glDrawArrays(self._render_mode, 0, vertex_count)
+    def _vertexBuffersSetup(self, mesh: Any):
```
nallath, line 264:
```suggestion
    def _vertexBuffersSetup(self, mesh: Any) -> None:
```

UM/View/RenderBatch.py
```diff
@@ -302,21 +261,77 @@ def _renderItem(self, item: Dict[str, Any]):
...
+            index_buffer.bind()
+
+        self._setMeshAttributes(mesh)
+
+        # Cache and return:
+        mesh.setCachedUserValue(self._shader.getReferenceKey(), vao)
+        vao.release()
+        return vao
+
+    def _renderItem(self, item: Dict[str, Any]):
```
nallath, line 296:
```suggestion
    def _renderItem(self, item: Dict[str, Any]) -> None:
```

bremco:
Lesson for next time: Run the tests _after_ I've merged with main as well.

bremco:
(... actually I don't think I did run the tests. I was confused with last week, when I had to set them up again on my work lappy.)

In any case, please try again, should pass now!

UM/View/RenderBatch.py
```diff
@@ -302,21 +261,79 @@ def _renderItem(self, item: Dict[str, Any]):
...
-                else:
-                    self._gl.glDrawElements(self._render_mode, mesh.getFaceCount(), self._gl.GL_UNSIGNED_INT, None)
-            else:
-                if self._render_mode == self.RenderMode.Triangles:
-                    self._gl.glDrawRangeElements(self._render_mode, self._render_range[0], self._render_range[1], self._render_range[1] - self._render_range[0], self._gl.GL_UNSIGNED_INT, None)
-                else:
-                    self._gl.glDrawElements(self._render_mode, self._render_range[1] - self._render_range[0], self._gl.GL_UNSIGNED_INT, None)
-        else:
-            self._gl.glDrawArrays(self._render_mode, 0, vertex_count)
+    def _vertexBuffersSetup(self, mesh: Any) -> Optional[QOpenGLVertexArrayObject]:
```
fieldOfView, line 264: `Any` is a bit lazy here, when you obviously expect `MeshData`

bremco: Oops! Fixed in https://github.com/Ultimaker/Uranium/pull/714/commits/8611188f88ec11dbef37bcd40dd96f79d17540cd

There's a joke about lazy evaluation in there ... _but I just couldn't be bothered._

bremco:
I still didn't do a proper benchmark, but playing around a bit with some printouts of the average time of the render of the SimulationView (where I expect most of the time-win will be) it seems that the newer code is about 20% faster.

UM/View/RenderBatch.py
```diff
@@ -234,43 +229,7 @@ def render(self, camera: Optional[Camera]):
...
-            index_buffer = OpenGL.getInstance().createIndexBuffer(mesh)
-        else:
-            # glDrawRangeElements does not work as expected and did not get the indices field working..
-            # Now we're just uploading a clipped part of the array and the start index always becomes 0.
-            index_buffer = OpenGL.getInstance().createIndexBuffer(
-                mesh, force_recreate=True, index_start = self._render_range[0], index_stop = self._render_range[1])
-        if index_buffer is not None:
-            index_buffer.bind()
-
+    def _setMeshAttributes(self, mesh: Any) -> None:
```
fieldOfView, line 232:
```suggestion
    def _setMeshAttributes(self, mesh: MeshData) -> None:
```

bremco:
When trying to benchmark this, I noticed `lineMeshVertexCount` (the entire layer version, not the polygon one), which I now use in SimulationPass, is quite a slow function compared to what was there before.

I need to set up some caching or something, or this PR will feel incomplete (even though it's already a bit faster). A lot of SimilationPass could be imporved, but let's try to keep it a bit smaller for now.

rburema:
Closing this in favour of https://github.com/Ultimaker/Uranium/pull/721 due to laptop & account related shenanigans.

PR #717, opened 2021-07-27
[SEC-240] Fix items that would be copied outside of storage location.

rburema:
Also refactor those sorts of test into the signed folder pre-storage check, where they belong. (Otherwise a part of that security was handled by the plugin system.) Also amend unit-tests to take into account these situations. While doing that, found an oversight in the `isPathInLocation` functionality, which is by necessity now also in Trust.

SEC-240 / CURA-8407

### DIFF
--- a/UM/CentralFileStorage.py
+++ b/UM/CentralFileStorage.py
@@ -69,8 +69,8 @@ class CentralFileStorage:
                 cls._unmoved_files[full_identifier] = path
             return
 
-        if not os.path.exists(cls.centralStorageLocation()):
-            os.makedirs(cls.centralStorageLocation())
+        if not os.path.exists(cls.getCentralStorageLocation()):
+            os.makedirs(cls.getCentralStorageLocation())
         if not os.path.exists(path):
             Logger.debug(f"{path_id} {str(version)} was already stored centrally or the provided path is not correct")
             return
@@ -129,10 +129,10 @@ class CentralFileStorage:
         :return: A path to store such an item.
         """
         item_name = item_id + "." + str(version)
-        return os.path.join(cls.centralStorageLocation(), item_name)
+        return os.path.join(cls.getCentralStorageLocation(), item_name)
 
     @classmethod
-    def centralStorageLocation(cls) -> str:
+    def getCentralStorageLocation(cls) -> str:
         """
         Gets a directory to store things in a version-neutral location.
         :return: A directory to store things centrally.
--- a/UM/CentralFileStorage.py
+++ b/UM/CentralFileStorage.py
@@ -69,8 +69,8 @@ class CentralFileStorage:
                 cls._unmoved_files[full_identifier] = path
             return
 
-        if not os.path.exists(cls._centralStorageLocation()):
-            os.makedirs(cls._centralStorageLocation())
+        if not os.path.exists(cls.centralStorageLocation()):
+            os.makedirs(cls.centralStorageLocation())
         if not os.path.exists(path):
             Logger.debug(f"{path_id} {str(version)} was already stored centrally or the provided path is not correct")
             return
@@ -129,10 +129,10 @@ class CentralFileStorage:
         :return: A path to store such an item.
         """
         item_name = item_id + "." + str(version)
-        return os.path.join(cls._centralStorageLocation(), item_name)
+        return os.path.join(cls.centralStorageLocation(), item_name)
 
     @classmethod
-    def _centralStorageLocation(cls) -> str:
+    def centralStorageLocation(cls) -> str:
         """
         Gets a directory to store things in a version-neutral location.
         :return: A directory to store things centrally.
--- a/UM/PluginRegistry.py
+++ b/UM/PluginRegistry.py
@@ -730,7 +730,7 @@ class PluginRegistry(QObject):
             full_path = os.path.join(plugin_path, file_to_move[0])
             try:
                 CentralFileStorage.store(full_path, file_to_move[1], Version(file_to_move[2]), move_file = not is_bundled_plugin)
-            except (FileExistsError, TypeError, IndexError):
+            except (FileExistsError, TypeError, IndexError, OSError):
                 Logger.logException("w", f"Can't move file {file_to_move[0]} to central storage for '{plugin_path}'.")
         return True
 
--- a/UM/PluginRegistry.py
+++ b/UM/PluginRegistry.py
@@ -9,7 +9,6 @@ import stat  # For setting file permissions correctly;
 import time
 import types
 import zipfile
-from pathlib import Path
 from typing import Any, Callable, Dict, List, Optional, Tuple, TYPE_CHECKING
 
 from PyQt5.QtCore import QCoreApplication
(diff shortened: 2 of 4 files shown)


UM/CentralFileStorage.py
```diff
@@ -129,10 +129,10 @@ def _getItemPath(cls, item_id: str, version: Version) -> str:
         :return: A path to store such an item.
         """
         item_name = item_id + "." + str(version)
-        return os.path.join(cls._centralStorageLocation(), item_name)
+        return os.path.join(cls.centralStorageLocation(), item_name)
 
     @classmethod
-    def _centralStorageLocation(cls) -> str:
+    def centralStorageLocation(cls) -> str:
```
nallath, line 135: Nitpicking: Shouldn't it be getCentralStorageLocation? We tend to format properties like "centralStorageLocation" and functions as "getCentralStorageLocation"

UM/Trust.py
```diff
@@ -285,6 +285,16 @@ def removeCached(path: str) -> bool:
             Logger.logException("e", "Removal of pycache for unbundled path '{0}' failed.".format(path))
         return False
 
+    @staticmethod
+    def isPathInLocation(location: str, path: str) -> bool:
```
nallath, line 289: It's a public function now, so i think this has more need for documentation now.

UM/Trust.py
```diff
@@ -451,9 +461,29 @@ def signedFolderPreStorageCheck(self, path: str) -> bool:
                     self._violation_handler(f"Manifest '{manifest_path}' is not properly self-signed in '{path}'.")
                     return False
 
-                # Otherwise, as far as this quick pre check is concerned, there is nothing wrong:
-                Logger.log("i", f"Central storage file signed correctly for '{path}'.")
-                return True
+            # Check if the central storage file doesn't contain files that would be moved outside the plugin folder,
+            #   or files that would be moved to outside of the central storage location:
```
nallath, line 465:
```suggestion
            #  or files that would be moved to outside of the central storage location:
```

UM/Trust.py
```diff
@@ -285,6 +285,16 @@ def removeCached(path: str) -> bool:
             Logger.logException("e", "Removal of pycache for unbundled path '{0}' failed.".format(path))
         return False
 
+    @staticmethod
+    def isPathInLocation(location: str, path: str) -> bool:
+        try:
+            canonical_location = os.path.normpath(location)
```
nallath, line 291: Since we're now at python 3.8, pathlib has much better support. Perhaps we should consider using that more (it at the very least has a nicer interface / convenience functions!). It's probably a bit out of scope for this issue though.

UM/Trust.py
```diff
@@ -451,9 +461,29 @@ def signedFolderPreStorageCheck(self, path: str) -> bool:
                     self._violation_handler(f"Manifest '{manifest_path}' is not properly self-signed in '{path}'.")
                     return False
 
-                # Otherwise, as far as this quick pre check is concerned, there is nothing wrong:
-                Logger.log("i", f"Central storage file signed correctly for '{path}'.")
-                return True
+            # Check if the central storage file doesn't contain files that would be moved outside the plugin folder,
+            #   or files that would be moved to outside of the central storage location:
+            with open(central_storage_filename, "r", encoding = "utf-8") as central_storage_file:
+                central_storage_list = json.loads(central_storage_file.read())
```
nallath, line 467: The json could be malformatted. Not sure if we should catch that (and log it) or just accept that the exception bubbles up into a critical error.

tests/TestTrust.py
```diff
@@ -226,3 +244,15 @@ def test_keyIOFails(self):
     def test_signNonexisting(self):
         private_key, public_key = TrustBasics.generateNewKeyPair()
         assert TrustBasics.getFileSignature("file-not-found", private_key) is None
+
+    def test_isPathInLocation(self):
```
nallath, line 248: I think this test should be re-written with parametrize. Best way to do that would be to have two different tests; One that asserts truth, other that asserts false.

Although it's a bit more code to parametrize it, i personally prefer it, since it means that all the asserts are executed, even if an earlier one fails.

nallath (changes requested):
I haven't checked this myself, but did we also check if the coverage of the trust tests is 100%? If not, we should probably add a task for that ASAP (and perhaps even enforce that in the CI!)

UM/Trust.py
rburema: It is caught, see the try-statement on line 432.

UM/CentralFileStorage.py
rburema: It's basically a read-only property. I didn't name it as such, I just removed the 'private to class' marker ;-)

UM/Trust.py
rburema: Fair enough!

UM/Trust.py
rburema: On purpose, to show that the sentence continues. Don't mind it lined out though, but then you'd want to remove both spaces, not just one.

tests/TestTrust.py
rburema: Sure, I'll start on that when I'm done with these unit-tests.

rburema:
> I haven't checked this myself, but did we also check if the coverage of the trust tests is 100%? If not, we should probably add a task for that ASAP (and perhaps even enforce that in the CI!)

They used to be if not 100%, at least very close to that, before the Central Storage was added. I put a few more there now to shore it up at least a little bit. We should indeed make another task, as there may be a few code-paths that are still missing.

UM/CentralFileStorage.py
rburema: Fixed in https://github.com/Ultimaker/Uranium/pull/717/commits/81250f7af3c809cf213a6464016698bfeb9c0c12

UM/Trust.py
rburema: Fixed in https://github.com/Ultimaker/Uranium/pull/717/commits/81250f7af3c809cf213a6464016698bfeb9c0c12

UM/Trust.py
rburema: Fixed in https://github.com/Ultimaker/Uranium/pull/717/commits/81250f7af3c809cf213a6464016698bfeb9c0c12

tests/TestTrust.py
rburema: Fixed in https://github.com/Ultimaker/Uranium/pull/717/commits/81250f7af3c809cf213a6464016698bfeb9c0c12

UM/Trust.py
jellespijker: I feel your pain and yes out-of-scope for this ticket. But nice to keep in mind for future reference.

UM/PluginRegistry.py
```diff
@@ -737,16 +728,10 @@ def _handleCentralStorage(self, file_data: str, plugin_path: str, is_bundled_plu
...
-            # Check if the central storage file didn't contain any files-to-copy outside of the plugin directory:
-            if self._check_if_trusted and not is_bundled_plugin and not self._isPathInLocation(plugin_path, full_path):
-                Logger.log("w", f"Suspect central file location '{full_path}' for file in '{plugin_path}'.")
-                return False
-
             try:
                 CentralFileStorage.store(full_path, file_to_move[1], Version(file_to_move[2]), move_file = not is_bundled_plugin)
-            except (TypeError, IndexError):
-                Logger.logException("w", f"Can't move file {file_to_move[1]} to central storage for '{plugin_path}'.")
+            except (FileExistsError, TypeError, IndexError):
```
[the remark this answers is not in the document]
jellespijker: Will add it myself since this is the only remark I have

jellespijker (approved):
Looks good.
Single remark which I chose to implement myself so that it can be moved to QA

PR #719, opened 2021-07-30
[CURA-8412] Make creation of shader program return None on failure.

rburema:
Shaders are already treated (in frontend program) as if they where None on failure. Now actually make it so.

This is part of a bugfix that came to our attention as 'If the slice button is clicked while the context-menu is also open on that same button, Cura crashes'. Suspect it may fix more than just that. For the rest of that more specific fix, see: https://github.com/Ultimaker/Cura/pull/10210

### DIFF
--- a/UM/View/GL/OpenGL.py
+++ b/UM/View/GL/OpenGL.py
@@ -1,4 +1,4 @@
-# Copyright (c) 2019 Ultimaker B.V.
+# Copyright (c) 2021 Ultimaker B.V.
 # Uranium is released under the terms of the LGPLv3 or higher.
 
 import sys
@@ -6,7 +6,7 @@ import ctypes   # type: ignore
 
 from PyQt5.QtGui import QOpenGLVersionProfile, QOpenGLContext, QOpenGLFramebufferObject, QOpenGLBuffer
 from PyQt5.QtWidgets import QMessageBox
-from typing import Any, TYPE_CHECKING, cast
+from typing import Any, TYPE_CHECKING, cast, Optional
 
 from UM.Logger import Logger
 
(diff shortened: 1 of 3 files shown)


UM/View/GL/OpenGL.py
```diff
@@ -199,7 +199,10 @@ def createShaderProgram(self, file_name: str) -> ShaderProgram:
             # If the loading failed, it could be that there is no specific shader for this version.
             # Try again without a version nr to get the generic one.
             if version_string != "":
-                shader.load(file_name, version = "")
+                try:
+                    shader.load(file_name, version = "")
+                except InvalidShaderProgramError:
+                    return None
```
konskarm, line 205: I think we should log the InvalidShaderProgramError message here.

UM/View/GL/ShaderProgram.py
```diff
@@ -127,6 +130,9 @@ def setVertexShader(self, shader):
 
         if not self._shader_program.addShaderFromSourceCode(QOpenGLShader.Vertex, shader):
             Logger.log("e", "Vertex shader failed to compile: %s", self._shader_program.log())
+            return False
+
+        return True
```
konskarm, line 135: The return type should be reflected in the typing

UM/View/GL/ShaderProgram.py
```diff
@@ -138,13 +144,19 @@ def setFragmentShader(self, shader):
 
         if not self._shader_program.addShaderFromSourceCode(QOpenGLShader.Fragment, shader):
             Logger.log("e", "Fragment shader failed to compile: %s", self._shader_program.log())
+            return False
+
+        return True
```
konskarm, line 149: See similar comment above.

UM/View/GL/ShaderProgram.py
```diff
@@ -138,13 +144,19 @@ def setFragmentShader(self, shader):
...
 
     def setGeometryShader(self, shader):
         if not self._shader_program:
             self._shader_program = QOpenGLShaderProgram()
 
         if not self._shader_program.addShaderFromSourceCode(QOpenGLShader.Geometry, shader):
             Logger.log("e", "Geometry shader failed to compile: %s", self._shader_program.log())
+            return False
+
+        return True
```
konskarm, line 159: See similar comment above.

konskarm (changes requested):

PR #724, opened 2021-08-13
Fix the rotation direction of the 90 degree rotation arrows

fieldOfView:
This PR fixes the rotation direction of the 90 degree rotation arrows in the Rotate Tool. Currently the arrows in both directions rotate the model in the same direction. They should operate in opposite directions.

Fixes https://github.com/Ultimaker/Cura/issues/10279


fieldOfView:
I think I broke the behavior here: fa210f34c3db42720bcd083066c5fc39a7a00c1a. It is fixed with this PR.

nallath (approved):

nallath:
For our reference CURA-8481

nallath:
I've also cherry-picked this into the 4.11 branch (It is in feature freeze, but not yet in bug fix freeze)

PR #725, opened 2021-08-17
Fixed the _blacklist variable

YuvalZilber:
Fixed _blacklist variable to include '__class__', '__globals__', 'hasattr' as 3 seperate strings and not '__class____globals__hasattr' as 1 combined string


nallath (approved):
Nice catch! We might need to backport this to our LTS versions as it's a security fix. I will discuss this with the rest of the team.

rburema:
Thanks!

It took a bit longer since this caused some discussion regarding cherry-picking it to another branch as well. That may still happen, but at least this isn't kept open anymore.

PR #726, opened 2021-08-18
[CURA-8491] Update SDK 7.6 -> 7.7 because of API changes.

rburema:
URANIUM:

* Added optional 'message_type' argument to Message.__init__
* Added Message.MessageType enum-class
* Added Message.getMessageType(self) -> MessageType
  Messages can have types now, this determines the sort of icon they get for now as well.

* Added TrustBasics.isPathInLocation(location: str, path: str) -> bool
  Whether a path is a sub-folder (or equal) of a given location (path).

* Changed OpenGL.createShaderProgram(self, file_name: str) -> Optional[ShaderProgram]
  Now returns optional instead of 'just' a shader-program, since part of the code already expected that to be the case.

* Removed 'UpdateCheckerJob' class
  Now just handled withint the UpdateCheckerJob.

### DIFF
--- a/resources/bundled_packages/uranium.json
+++ b/resources/bundled_packages/uranium.json
@@ -5,8 +5,7 @@
             "package_type": "plugin",
             "display_name": "Console Logger",
             "description": "Outputs log information to the console.",
-            "package_version": "1.0.1",
-            "sdk_version": "7.6.0",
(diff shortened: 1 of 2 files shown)


Ghostkeeper (approved):
Not sure if it's really necessary to update that conftest. But it's fine.

PR #728, opened 2021-08-20
Retain the last saved filename after saving a file

konskarm:
Cura was not retaining the last saved filename selected by the user after they were saving their file. This PR fixes that by making sure that the filename selected by the user in the save file dialog will update the name of the workspace (project) all around Cura.

**Note:** The stored filename is forgotten when the user loads another project file, creates a new project (Ctrl+N) or clears the buildplate (Ctrl+D).

Original feature request: https://github.com/Ultimaker/Cura/issues/10061

Required by https://github.com/Ultimaker/Cura/pull/10328.

CURA-8358

### DIFF
--- a/plugins/LocalFileOutputDevice/LocalFileOutputDevice.py
+++ b/plugins/LocalFileOutputDevice/LocalFileOutputDevice.py
@@ -14,7 +14,6 @@ from UM.Logger import Logger
 from UM.Mesh.MeshWriter import MeshWriter
 from UM.Message import Message
 from UM.OutputDevice import OutputDeviceError
-from UM.OutputDevice.OutputDevice import OutputDevice
 from UM.OutputDevice.ProjectOutputDevice import ProjectOutputDevice
 from UM.i18n import i18nCatalog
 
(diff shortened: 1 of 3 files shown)


nallath (approved):

nallath:
I've found two weird things with this.

Clearing buildplate doesn't reset last used filename:
image

Changing machine doesn't reset last used filename:
image

plugins/LocalFileOutputDevice/LocalFileOutputDevice.py
```diff
@@ -140,6 +139,7 @@ def requestWrite(self, nodes, file_name = None, limit_mimetypes = None, file_han
                 raise OutputDeviceError.UserCanceledError()
 
         self.writeStarted.emit(self)
+        Application.getInstance().workspaceLoaded.emit(file_name)
```
nallath, line 142: I don't think that this is correct. This means that if you load two STL files that the second STL changes the file name.

nallath (changes requested):

plugins/LocalFileOutputDevice/LocalFileOutputDevice.py
konskarm: Nope. Loading STL files triggers the workspaceLoaded only if the buildplate is completely empty. Otherwise they are not workspaces so this isn't trigger again.

Ghostkeeper: The trigger being called is not correct either.

Basically this means that if any file is saved to local disk, we're pretending that we're "opening a workspace". No workspace has been opened, so anything that listens to this signal would expect that a workspace has been opened. But instead, a file has been loaded. Completely different situations.

Ghostkeeper (changes requested):

nallath (approved):

PR #734, opened 2021-08-30
CURA-8500: Safely convert message image sources to QUrls

konskarm:
Some messages may already provide a QUrl as the image source. In such cases, Cura would crash because the `QUrl.fromLocalFile()` expects a string and not a QUrl as input. 
This is now accounted for to make sure that in the image source will not be converted to QUrl again.

CURA-8500

### DIFF
--- a/UM/Qt/Bindings/VisibleMessagesModel.py
+++ b/UM/Qt/Bindings/VisibleMessagesModel.py
@@ -67,7 +67,7 @@ class VisibleMessagesModel(ListModel):
         message.progressChanged.connect(self._onMessageProgress)
 
     @staticmethod
-    def getImageSourceAsQUrl(image_source):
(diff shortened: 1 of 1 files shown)


UM/Qt/Bindings/VisibleMessagesModel.py
```diff
@@ -66,6 +66,14 @@ def addMessage(self, message):
         message.textChanged.connect(self._onMessageTextChanged)
         message.progressChanged.connect(self._onMessageProgress)
 
+    @staticmethod
+    def getImageSourceAsQUrl(image_source):
```
nallath, line 70:
```suggestion
    def getImageSourceAsQUrl(image_source: Union[QUrl, str]) -> QUrl:
```

nallath (dismissed):

PR #741, opened 2021-10-01
Add cloud message type

nallath:
CURA-8601
https://github.com/Ultimaker/Cura/pull/10547

### DIFF
--- a/UM/Message.py
+++ b/UM/Message.py
@@ -28,6 +28,7 @@ class Message(QObject):
         NEUTRAL = 1
         WARNING = 2
         ERROR = 3
+        CLOUD = 4
 
     def __init__(self, text: str = "", lifetime: int = 30, dismissable: bool = True, progress: float = None,
                  title: Optional[str] = None, parent=None, use_inactivity_timer: bool = True, image_source: str = "",
--- a/UM/Qt/qml/UM/MessageStack.qml
+++ b/UM/Qt/qml/UM/MessageStack.qml
@@ -103,6 +103,7 @@ ListView
                 {
                     switch (model.message_type)
                     {
+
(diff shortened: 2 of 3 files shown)


Ghostkeeper:
The failing automated test seems to be unrelated to these changes. Is it nondeterminate? :thinking:

UM/Message.py
```diff
@@ -28,6 +28,7 @@ class MessageType(IntEnum):
         NEUTRAL = 1
         WARNING = 2
         ERROR = 3
+        CLOUD = 4
```
Ghostkeeper, line 31: Instead of adding a message type for every sort of icon we want to show there, maybe it's more appropriate to have a separate field to override the icon for a particular message? I'm bringing this up for three reasons:

* We haven't been using these message types for long and there's already different icons they want to have shown. It'll happen more often.
* The "cloud" message type doesn't fit within the rest of the types in this list. It's not like an error level or a positive vs. negative sense.
* This adds cloud knowledge to Uranium. Adding an icon override field would allow the reference to the icon to be contained within Cura or even within the plug-in.

We can figure out the colour separately. It could be another separate field, or we could make the neutral message type have "primary" colour, which was the original idea in the first message designs anyway.

Ghostkeeper (changes requested):

UM/Message.py
nallath: Problem is; That is well beyond the scope of this feature. The way that the icon is handled right now doesn't support having custom icons at all. 

Unfortunately, we didn't see this during the refinement, otherwise we wouldn't have set it to the low amount of points that we did.

nallath:
Closed as we went for an easier solution.

Issue #744, opened 2021-10-26
ViewModel._update() does not support plugins with multiple views / views with PluginName_ViewName nomenclature


Tyronnosaurus:
I found this possible bug trying to solve another issue. More info in [this Cura forums thread](https://community.ultimaker.com/topic/38259-plugin-development-hide-a-new-view-from-the-view-selector-in-preview-stage/).

------------------------------------------------
First, some info we need to know:

- If we create a plugin with one View, its ID is the plugin's name.
- If we create a plugin with 2 or more Views, in order to differentiate between them we need to assign a self.name value to each. Then, their ID is **PluginName_ViewName**. See more info in [this Cura Forums thread](https://community.ultimaker.com/topic/34982-plugin-with-2-stages/?tab=comments#comment-278488) and in the method [PluginObject.getId()](https://github.com/Ultimaker/Uranium/blob/d8e70aa3cae26dda9b5df901f50ccbdb51a366f8/UM/PluginObject.py#L26)
This system ensures no two Views ever have the same name, no matter how many plugins we install.
-------------------------------------------------
The actual bug:

[ViewModel._update()](https://github.com/Ultimaker/Uranium/blob/d8e70aa3cae26dda9b5df901f50ccbdb51a366f8/UM/Qt/Bindings/ViewModel.py#L39) has no support for the PluginName_ViewName nomenclature:

```
        views = self._controller.getAllViews()
        ...
        for view_id in views:
            view_meta_data = PluginRegistry.getInstance().getMetaData(view_id).get("view", {})
```
 
**If we've got a view called MyPlugin_View1, then _getMetaData(view_id)_ will try to find a plugin called "MyPlugin_View1" instead of just "MyPlugin".**

Moreover, even if it found the plugin in question, get() would return a list instead of a dict, due to the fact that in the file plugin.json you can declare either one view (a dict) or multiple views (list of dicts). We'd need to filter every dict to find the one where key 'name' is equal to MyView.name (both are set manually, so the developer must ensure they're the same).

-------------------------------------------------
Possible solution:
```
        views = self._controller.getAllViews()
        ...
        for view_id in views:
          #Views may be named as "PluginName" or "PluginName_ViewName"
          plugin_name = view_id.split('_')[0]    #Get the part before '_', or all the string if there's no '_'
          view_meta_data = PluginRegistry.getInstance().getMetaData(plugin_name).get("view", {})
          if isinstance(view_meta_data,list):      #Multiple views in the plugin
[...]

PR #745, opened 2021-10-27
Repaired Views dropdown in Preview stage

Tyronnosaurus:
(for plugins with 2 views or more)

If a view has a _name, its ID is not PluginName but rather PluginName_ViewName. This is necessary to give unique IDs to each view in a plugin that has multiple views. However, it breaks this line:
```
   view_meta_data = PluginRegistry.getInstance().getMetaData(view_id).get("view", {})
```
because it tries to look for a plugin called PluginName_ViewName.

More info in [Cura forums thread](https://community.ultimaker.com/topic/38259-plugin-development-hide-a-new-view-from-the-view-selector-in-preview-stage/?tab=comments#comment-297251) & issue #744.

### DIFF
--- a/UM/Qt/Bindings/ViewModel.py
+++ b/UM/Qt/Bindings/ViewModel.py
@@ -5,7 +5,6 @@ from PyQt5.QtCore import Qt
 
 from UM.Qt.ListModel import ListModel
 from UM.Application import Application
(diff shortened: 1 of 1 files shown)


nallath:
Sorry it took me so long to re-review / test the changes! Somehow i missed the notification that you applied the changes.

PR #748, opened 2021-11-03
Cura 8558 get removed and to install packages

jellespijker:
Added some logic to obtain the **to be removed** and **to be installed** packages.

Which simplifies the logic in the Cura Marketplace plugin. See PR Ultimaker/Cura#10732

### DIFF
--- a/UM/PackageManager.py
+++ b/UM/PackageManager.py
@@ -37,8 +37,8 @@ class PackageManager(QObject):
         self._plugin_registry = self._application.getPluginRegistry()
 
         # JSON files that keep track of all installed packages.
-        self._user_package_management_file_path = None  # type: Optional[str]
-        self._bundled_package_management_file_paths = []  # type: List[str]
+        self._user_package_management_file_path: Optional[str] = None
+        self._bundled_package_management_file_paths: List[str] = []
         for search_path in Resources.getAllPathsForType(Resources.BundledPackages):
             if not os.path.isdir(search_path):
                 continue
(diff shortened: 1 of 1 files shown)


UM/PackageManager.py
```diff
@@ -64,19 +64,20 @@ def __init__(self, application: "QtApplication", parent: Optional[QObject] = Non
...
-        self._installation_dirs_dict = {"plugins": os.path.abspath(Resources.getStoragePath(Resources.Plugins))}  # type: Dict[str, str]
+        self._installation_dirs_dict: Dict[str, str] = {"plugins": os.path.abspath(Resources.getStoragePath(Resources.Plugins))}
 
-        self._bundled_package_dict = {}  # type: Dict[str, Dict[str, Any]] # A dict of all bundled packages
-        self._installed_package_dict = {}  # type: Dict[str, Dict[str, Any]] # A dict of all installed packages
-        self._to_remove_package_set = set()  # type: Set[str] # A set of packages that need to be removed at the next start
-        self._to_install_package_dict = {}  # type: Dict[str, Dict[str, Any]]  # A dict of packages that need to be installed at the next start
-        self._dismissed_packages = set()    # type: Set[str] # A set of packages that are dismissed by the user
-        self._installed_packages = {}  # type: Dict[str, Dict[str, Any]]  # A dict of packages that were installed during startup
+        self._bundled_package_dict: Dict[str, Dict[str, Any]] = {}  # A dict of all bundled packages
```
nallath, line 69: Nitpicking (maybe shouldn't be handled in this change, but hey, boyscouting rules); Maybe we should just define a package_data = Dict[str, Any] and then define bundled_package_dict type as Dict[str, package_data]. 

I think that would make it easier to read and also easier to refine the type if that's ever needed (eg; perhaps when we no longer allow Any but a subset thereof)

nallath (approved):

UM/PackageManager.py
jellespijker: I agree with you here, do we have out own typing module with custom types defined, such that these can be reused?I will change that.

jellespijker: I modernized most of the typings and used your proposed custom types, only changed the case of these types to be more inline with the Python  style. So they became: `PackageData` and `BundledPackageDict`

See commit 5d41cc6

rburema (approved):

PR #750, opened 2021-11-06
Round coordinates on getFaceIdAtPosition

eliadevito:
PyQt Pixel expects integer coordinates as parameters.

This fixes the crash on cura when using the
"Select face to align to the build plate" tool


nallath (approved):

nallath:
Thanks for the contribution!

PR #755, opened 2021-12-06
Shows a remove plugin message if the central storage location is missing

jellespijker:
This should warn the user that a backed up plugin has been corrupted and should be reinstalled

Contributes to: CURA-8666

### DIFF
--- a/UM/PluginRegistry.py
+++ b/UM/PluginRegistry.py
@@ -444,6 +444,7 @@ class PluginRegistry(QObject):
             to_register = plugin.register(self._application)  # type: ignore  # We catch AttributeError on this in case register() doesn't exist.
             if not to_register:
                 Logger.log("w", "Plugin %s did not return any objects to register", plugin_id)
+                self.removeCorruptedPluginMessage(plugin_id)
                 return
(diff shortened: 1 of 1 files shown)


UM/PluginRegistry.py
```diff
@@ -471,21 +472,25 @@ def loadPlugin(self, plugin_id: str) -> None:
...
-                                   button_align = Message.ActionButtonAlignment.ALIGN_RIGHT)
-
-            # Listen for the pyqt signal, since that one does support lambda's
-            unable_to_load_plugin_message.pyQtActionTriggered.connect(lambda message, action: (self.uninstallPlugin(plugin_id), message.hide()))
-
-            unable_to_load_plugin_message.show()
-            Logger.logException("e", "Error loading plugin %s:", plugin_id)
+            self.removeCorruptedPluginMessage(plugin_id)
+
+    def removeCorruptedPluginMessage(self, plugin_id):
```
nallath, line 477:
```suggestion
    def removeCorruptedPluginMessage(self, plugin_id: str) -> None:
```

nallath (changes requested):

nallath (approved):

PR #756, opened 2021-12-07
Cura 8587 disable update install and uninstall

jellespijker:
See also the Cura PR Ultimaker/Cura#11027

### DIFF
--- a/UM/PluginRegistry.py
+++ b/UM/PluginRegistry.py
@@ -234,9 +234,6 @@ class PluginRegistry(QObject):
             self.hasPluginsEnabledOrDisabledChanged.emit()
         self._savePluginData()
 
-    def getDisabledPlugins(self) -> List[str]:
-        return self._disabled_plugins
-
     #   Get a list of enabled plugins:
     def getActivePlugins(self) -> List[str]:
         plugin_list = []
--- a/UM/PluginRegistry.py
+++ b/UM/PluginRegistry.py
@@ -9,7 +9,7 @@ import stat  # For setting file permissions correctly;
 import time
 import types
 import zipfile
-from typing import Any, Callable, Dict, List, Optional, Tuple, TYPE_CHECKING
+from typing import Any, Callable, Dict, List, Optional, Set, Tuple, TYPE_CHECKING
 
 from PyQt5.QtCore import QCoreApplication
 from PyQt5.QtCore import QObject, pyqtSlot, QUrl, pyqtProperty, pyqtSignal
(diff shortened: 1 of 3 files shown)


UM/Qt/ListModel.py
```diff
@@ -161,14 +161,19 @@ def setProperty(self, index: int, property: str, value: Any) -> None:
...
 
         :param fun: The callable to use for determining the sort key.
+        :param key: Use the sorting function on the underlying data
+        :param reverse: reverse the sorted results
         """
 
         self.beginResetModel()
-        self._items.sort(key = fun)
+        if key:
+            self._items = [{key: item} for item in sorted([k[key] for k in self._items], key = fun, reverse = reverse)]
```
rburema, line 174: I'm not sure how I feel about data being changed (format, not just positionally altered) by a sorting function. Calling this twice (each time) with a different key will result in an error for example., since it won't be able to find `k[key]` the 2nd time around, since that will be in `k[<key used first time>][key]` instead then.

rburema (changes requested):
Yay for 3.8 style typing! :-)

I'll leave you to ponder these change-requests while I do the big chunk of the review over in Cura.

UM/PluginRegistry.py
```diff
@@ -52,41 +52,42 @@ def __init__(self, application: "Application", parent: QObject = None) -> None:
...
-        self._checked_plugin_ids = []     # type: List[str]
-        self._distrusted_plugin_ids = []  # type: List[str]
-        self._trust_checker = None  # type: Optional[Trust]
+        self._check_if_trusted: bool = False
+        self._clean_hierarchy_sent_messages: List[str] = []
+        self._debug_mode: bool = False
+        self._checked_plugin_ids: List[str] = []
+        self._distrusted_plugin_ids: List[str] = []
+        self._trust_checker: Optional[Trust] = None
+        self._plugins_enabled_or_disabled: bool = False  # Flag indicating if there were any plugins' en-/disabled this session
```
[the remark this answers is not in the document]
jellespijker: Changed this into a proper book keeping variable with a name containing current. We can now toggle en-/disabled plugins in Cura and only restart if actually needed. As per @lorenzoromagnoli his request.

UM/Qt/ListModel.py
jellespijker: Yes you are indeed correct. Will change this

rburema (approved):

nallath (approved):

PR #757, opened 2021-12-07
Uninstall package and plugin on error

nallath:
CURA-8666

### DIFF
--- a/UM/PluginRegistry.py
+++ b/UM/PluginRegistry.py
@@ -88,6 +88,8 @@ class PluginRegistry(QObject):
         self._distrusted_plugin_ids = []  # type: List[str]
         self._trust_checker = None  # type: Optional[Trust]
 
+    pluginRemoved = pyqtSignal(str)
+
     def setCheckIfTrusted(self, check_if_trusted: bool, debug_mode: bool = False) -> None:
         self._check_if_trusted = check_if_trusted
         if self._check_if_trusted:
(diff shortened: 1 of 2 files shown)


UM/PluginRegistry.py
```diff
@@ -499,9 +501,6 @@ def uninstallPlugin(self, plugin_id: str) -> Dict[str, str]:
         result = {"status": "error", "message": "", "id": plugin_id}
         success_message = i18n_catalog.i18nc("@info:status", "The plugin has been removed.\nPlease restart {0} to finish uninstall.", self._application.getApplicationName())
 
-        if plugin_id not in self._plugins_installed:
-            return result
```
casperlamboo, line 502: now that we don't return the result on "error" anymore it doesn't make alot of sense to define the result object above; i would remove it and only define it at the return point of the function

UM/PluginRegistry.py
```diff
@@ -515,8 +514,9 @@ def uninstallPlugin(self, plugin_id: str) -> Dict[str, str]:
 
             # Remove the plugin object from the Plugin Registry:
             self._plugins.pop(plugin_id, None)
-            self._plugins_installed.remove(plugin_id)
-
+            if plugin_id in self._plugins_installed:
+                self._plugins_installed.remove(plugin_id)
```
casperlamboo, line 518: this whole code block executes `if not in_to_install`. Might I suggest removing the else statement entirely and executing it in for all cases. 

While I cannot think of any cases where a plugin would be the `_plugins_installed` object while also being in the `_plugins_to_install` object, I think it's best to cover all these edge cases. The internal check prevent plugins from being removed if they are not present; so it won't break stuff if it gets executed "incorrectly".

casperlamboo: https://github.com/Ultimaker/Uranium/commit/68d3c1508e23fd79dab1ae4fefcb15cce1695909

UM/PluginRegistry.py
casperlamboo: https://github.com/Ultimaker/Uranium/commit/188cf82c996f6469e8cfd29278cb194937fe2f7e

PR #759, opened 2021-12-14
Use a direct copy instead of cleanup for signal connect

nallath:
This makes the signals significantly faster.

The results were most visible in creation of the container trees (where a lot of signals are connected). This made it ~5x as fast to populate the container tree.

### DIFF
--- a/UM/Signal.py
+++ b/UM/Signal.py
@@ -542,7 +542,7 @@ class WeakImmutablePairList(Generic[T, U], Iterable):
         """
 
         new_instance = WeakImmutablePairList()  # type: WeakImmutablePairList[T,U]
-        new_instance.__list = self.__cleanList()
+        new_instance.__list = self.__list.copy()
         new_instance.__list.append( (weakref.ref(left_item), weakref.ref(right_item)) )
         return new_instance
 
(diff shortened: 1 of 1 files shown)


UM/Signal.py
```diff
@@ -542,7 +542,7 @@ def append(self, left_item: T, right_item: U) -> "WeakImmutablePairList[T,U]":
         """
 
         new_instance = WeakImmutablePairList()  # type: WeakImmutablePairList[T,U]
-        new_instance.__list = self.__cleanList()
+        new_instance.__list = self.__list.copy()
```
rburema, line 545: This also happens in `WeakImmutableList` (as opposed to `WeakImmutablePairList`). I'll put the same change there then, with the same reasoning. (Unless that messes up my tests.)

rburema (approved):
So um...

Have you checked the speed of the old Marketplace after these changes? Or maybe it was the thing I noticed you missed?

Ha ha.
Ha ha ha.

rburema:
Never mind, turns out it's a little bit faster, but 4.13 has the same behaviour.

Jeeze this new laptop is a _beast_ :-D

PR #760, opened 2021-12-15
Prevent unneeded copy in container query

nallath:
By adding the values of the medatada dict to a list, a complete copy of the list is made. This isn't actually needed, since we just want to filter results (wich are then copied to a list). As unneeded copies tend to be on the expensive side, this makes profile loading a lot faster

### DIFF
--- a/UM/Settings/ContainerQuery.py
+++ b/UM/Settings/ContainerQuery.py
@@ -87,7 +87,7 @@ class ContainerQuery:
         # cache if it's not there yet.
         key_so_far = (self._ignore_case, )  # type: Tuple[Any, ...]
         if candidates is None:
-            filtered_candidates = list(self._registry.metadata.values())
+            filtered_candidates = self._registry.metadata.values()
         else:
             filtered_candidates = candidates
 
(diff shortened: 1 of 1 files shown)


UM/Settings/ContainerQuery.py
```diff
@@ -87,9 +87,9 @@ def execute(self, candidates: Optional[List[Any]] = None) -> None:
         # cache if it's not there yet.
         key_so_far = (self._ignore_case, )  # type: Tuple[Any, ...]
         if candidates is None:
-            filtered_candidates = list(self._registry.metadata.values())
+            filtered_candidates: Union[ValuesView[Dict[str, Any]], List[Dict[str, Any]]] = self._registry.metadata.values()
```
rburema, line 90: This _does_ mean that the type now changes during execution (well, technically it's always that Union... but _eh_). Not really 'clean', but it'll work.

( If this was C++ I'd sent it back I think :-p )

nallath: Well, it quacks like a duck and waggles like a duck ;)

PR #761, opened 2021-12-15
Minor shader speedups

nallath:
Some small stuff that I found when looking at the profile file provided in https://github.com/Ultimaker/Cura/issues/11003

Note that these changes don't solve the real problem.

### DIFF
--- a/UM/View/GL/ShaderProgram.py
+++ b/UM/View/GL/ShaderProgram.py
@@ -361,15 +361,17 @@ class ShaderProgram:
             value = cast(Color, value)
             cast(QOpenGLShaderProgram, self._shader_program).setUniformValue(uniform,
                 QColor(round(value.r * 255), round(value.g * 255), round(value.b * 255), round(value.a * 255)))
+        elif type(value) is list and type(cast(List[List[float]], value)[0]) is list and len(
+            cast(List[List[float]], value)[0]) == 4:
+            value = cast(List[List[float]], value)
+            cast(QOpenGLShaderProgram, self._shader_program).setUniformValue(uniform,
+                                                                             self._matrixToQMatrix4x4(Matrix(value)))
         elif type(value) is list and len(cast(List[float], value)) == 4:
             value = cast(List[float], value)
             cast(QOpenGLShaderProgram, self._shader_program).setUniformValue(uniform, QVector4D(value[0], value[1], value[2], value[3]))
         elif type(value) is list and type(cast(List[List[float]], value)[0]) is list and len(cast(List[List[float]], value)[0]) == 2:
             value = cast(List[List[float]], value)
             cast(QOpenGLShaderProgram, self._shader_program).setUniformValueArray(uniform, [QVector2D(i[0], i[1]) for i in value])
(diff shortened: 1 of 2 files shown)


UM/View/GL/ShaderProgram.py
```diff
@@ -348,18 +348,25 @@ def removeAttributeBinding(self, key: str) -> None:
         del self._attribute_bindings[key]
 
     def _matrixToQMatrix4x4(self, m):
-        return QMatrix4x4(m.getData().flatten())
+        return QMatrix4x4(m.getFlatData())
 
     def _setUniformValueDirect(self, uniform: int, value: Union[Vector, Matrix, Color, List[float], List[List[float]], float, int]) -> None:
-        if type(value) is Vector:
+        if type(value) is Matrix:
```
rburema, line 354: It's not _completely_ clear (to me) we set Vectors more often than Matrices (at least with the current state of our code). Take a look at all the `vec2`, `vec3` and `vec4` in our shaders and compare that to the occurrences of `mat4`. Then again the camera is the thing that _should_ change the most, if we set the others more often then a lot of those are probably spurious changes.

rburema (changes requested):
Have you tried the horizontal slider, or the play button next to that (simulation-view)?

~~~
Traceback (most recent call last):
  File "C:\dev\Uranium\UM\Qt\Bindings\MainWindow.py", line 255, in _render
    renderer.render()
  File "C:\dev\Uranium\UM\Qt\QtRenderer.py", line 163, in render
    render_pass.render()
  File "C:\dev\Cura\plugins\SimulationView\SimulationPass.py", line 189, in render
    layers_batch.render(self._scene.getActiveCamera())
  File "C:\dev\Uranium\UM\View\RenderBatch.py", line 179, in render
    self._shader.bind()
  File "C:\dev\Uranium\UM\View\GL\ShaderProgram.py", line 272, in bind
    self._setUniformValueDirect(uniform, self._uniform_values[uniform])
  File "C:\dev\Uranium\UM\View\GL\ShaderProgram.py", line 366, in _setUniformValueDirect
    cast(QOpenGLShaderProgram, self._shader_program).setUniformValue(uniform, QVector4D(value[0], value[1], value[2], value[3]))
TypeError: arguments did not match any overloaded call:
  QVector4D(): too many arguments
  QVector4D(float, float, float, float): argument 1 has unexpected type 'list'
  QVector4D(QPoint): argument 1 has unexpected type 'list'
  QVector4D(Union[QPointF, QPoint]): argument 1 has unexpected type 'list'
  QVector4D(QVector2D): argument 1 has unexpected type 'list'
  QVector4D(QVector2D, float, float): argument 1 has unexpected type 'list'
  QVector4D(QVector3D): argument 1 has unexpected type 'list'
  QVector4D(QVector3D, float): argument 1 has unexpected type 'list'
  QVector4D(QVector4D): argument 1 has unexpected type 'list'
~~~

UM/View/GL/ShaderProgram.py
nallath: Well, I just ran this through a profiler and that is what I found. But the effect of changing the order is probably quite small.

nallath:
Whoops. No. I did check the layerview (and the vertical slider) but not the horizontal one.

rburema (approved):

PR #763, opened 2021-12-22
Update uranium.po

MarongHappy:


Ghostkeeper:
(See https://github.com/Ultimaker/Cura/pull/11107.)

PR #768, opened 2022-01-04
Upgrade to Python 3.10

Ghostkeeper:
This makes Cura depend on Python 3.10 instead of 3.8. To upgrade to 3.10 we had to drop Shapely and replace it with Pyclipper.

The tests are currently failing because Pyclipper is not in the master build environment yet. It will be added by https://github.com/Ultimaker/cura-build-environment/pull/119.

Contributes to issue CURA-8592.


fieldOfView:
As far as I remember, parts of TriMesh depend on Shapely.

rburema:
@fieldOfView 

> As far as I remember, parts of TriMesh depend on Shapely.

Fortunately, it doesn't appear as if we're currently using those parts, and it's installable without:

> "Keeping trimesh easy to install is a core goal, thus the only hard dependency is numpy. Installing other packages adds functionality but is not required."

> "This includes things like [...], vector path handling (shapely and rtree), [...]"

(from the [documentation](https://trimsh.org/trimesh.html))

I don't think we use `trimesh.path` anywhere, so while a dependency might still be hidden somewhere, on the face of it, it seems we're safe.

fieldOfView:
That is good news. I'll have to test if all of MeshTools also still works. It doesn't use `trimesh.path` directly either, but perhaps it needs it indirectly somewhere.

jellespijker (approved):

PR #773, opened 2022-01-24
[Cura-8378] improve checkbox visibility

Joeydelarago:
The aim of this ticket is to use new checkbox designs throughout Cura. The CheckBox Component defined in Cura has been standardised and moved here. 

The setting visibility menu contains the only checkbox not defined in the Cura project so it has been updated to use this new CheckBox here.

### DIFF
--- a/UM/Qt/qml/UM/CheckBox.qml
+++ b/UM/Qt/qml/UM/CheckBox.qml
@@ -77,9 +77,9 @@ CheckBox
             {
                 switch (control.checkState)
                 {
-                    case Qt.Checked: return UM.Theme.getIcon("EmptyCheck", "low")
+                    case Qt.Checked: return UM.Theme.getIcon("CheckMark", "low")
                     case Qt.PartiallyChecked: return UM.Theme.getIcon("CheckBoxFill", "low")
-                    default: return UM.Theme.getIcon("EmptyCheck", "low")
+                    default: return UM.Theme.getIcon("CheckMark", "low")
                 }
             }
             opacity:
--- a/UM/Qt/qml/UM/CheckBox.qml
+++ b/UM/Qt/qml/UM/CheckBox.qml
@@ -23,13 +23,9 @@ CheckBox
         {
             if (!control.enabled)
             {
-                return UM.Theme.getColor("setting_control_disabled")
+                return UM.Theme.getColor("checkbox_disabled")
             }
-            if (control.hovered || control.activeFocus)
-            {
-                return UM.Theme.getColor("setting_control_highlight")
-            }
-            return UM.Theme.getColor("setting_control")
+            return UM.Theme.getColor("checkbox")
         }
 
         radius: UM.Theme.getSize("checkbox_radius").width
@@ -38,13 +34,13 @@ CheckBox
         {
             if (!enabled)
             {
-                return UM.Theme.getColor("setting_control_disabled_border")
+                return UM.Theme.getColor("checkbox_border")
             }
             if (control.hovered || control.activeFocus)
             {
-                return UM.Theme.getColor("setting_control_border_highlight")
+                return UM.Theme.getColor("checkbox_border_hover")
             }
-            return UM.Theme.getColor("setting_control_border")
+            return UM.Theme.getColor("checkbox_border")
         }
 
         UM.RecolorImage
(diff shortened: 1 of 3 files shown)


fieldOfView:
Congratulations on your new job!

UM/Qt/qml/UM/CheckBox.qml
```diff
@@ -0,0 +1,105 @@
+// Copyright (c) 2021 Ultimaker B.V.
```
nallath, line 1:
```suggestion
// Copyright (c) 2022 Ultimaker B.V.
```
Nitpicking :P

UM/Qt/qml/UM/CheckBox.qml
```diff
@@ -0,0 +1,105 @@
...
+
+CheckBox
+{
+    id: control
+
+    hoverEnabled: true
+
+    indicator: Rectangle
+    {
+        height: UM.Theme.getSize("checkbox").height
```
nallath, line 17:
```suggestion
        implicitHeight: UM.Theme.getSize("checkbox").height
```

UM/Qt/qml/UM/CheckBox.qml
```diff
@@ -0,0 +1,105 @@
...
+        width: UM.Theme.getSize("checkbox").width
```
nallath, line 18:
```suggestion
        implicitWidth: UM.Theme.getSize("checkbox").width
```
If you're creating re-usable components, it's generally better to set the implicit width / height. Impicit height/widht is the width/height that is used if whoever creates the item doesn't set a size (So you can see it as "default size")

UM/Qt/qml/UM/CheckBox.qml
```diff
@@ -0,0 +1,105 @@
...
+        }
+
+        UM.RecolorImage
+        {
+            anchors.verticalCenter: parent.verticalCenter
+            anchors.horizontalCenter: parent.horizontalCenter
+
+            height:
+            {
+                switch(control.checkState)
```
nallath, line 53: This is generally fine, but for the future you might want to have a look at states: https://doc.qt.io/qt-5/qml-qtquick-state.html

This will have the exact same result, but what i prefer about it is that you define all the changes that need to happen to a component given a certain state (eg; the check state) in a single place.

UM/Qt/qml/UM/qmldir
```diff
@@ -25,5 +25,6 @@ ProgressBar 1.3 ProgressBar.qml
 SimpleButton 1.1 SimpleButton.qml
 TabRow 1.3 TabRow.qml
 TabRowButton 1.3 TabRowButton.qml
+CheckBox 1.0 CheckBox.qml
```
nallath, line 28:
```suggestion
CheckBox 1.5 CheckBox.qml
```
Well, no? You can't suddenly add a new type to a lower semver. We haven't been consistent with this, I do admit that, but today is a great day to start :)

nallath (changes requested):
For the future, could you also make sure to tag the bodies of commit messages with the jira ticket nr? This makes it a lot easier to find the context of changes some time from now.

UM/Qt/qml/UM/CheckBox.qml
```diff
@@ -74,9 +76,10 @@ CheckBox
                 {
                     name: "partiallyChecked"
                     when: (control.checkState == Qt.PartiallyChecked)
-                    PropertyChanges {
+                    PropertyChanges
+                    {
                         target: checkIcon
-                        height: UM.Theme.getSize("checkbox_square").height
+                        height: parent.height - Math.round((parent.height - UM.Theme.getSize("checkbox_square").height) / 2) * 2 //Round border to pixels to properly center it.
```
Joeydelarago, line 82: The checkbox still appears 1 pixel misaligned on the vertical for me. I don't think this is noticeable though.

Screenshot 2022-02-01 at 17 23 10

Joeydelarago: I'm viewing this at a resolution of 1920x1080. System em size is 14px. On Ubuntu 20.04

Ghostkeeper: This is what I'm seeing:
image

Your picture is how it looked before my last commit in this PR! Something weird going on then.

PR #776, opened 2022-01-28
Update to sip 6 

nallath:
We were still using SIP 4.x, but it's on end of life for quite some time now.

https://github.com/Ultimaker/CuraEngine/pull/1565
https://github.com/Ultimaker/libArcus/pull/129


jellespijker:
Closing this in favor #804

PR #780, opened 2022-02-15
Set default priority of backend to normal on Windows

nallath:
CURA-8957
Fixes Ultimaker/Cura#11466


Ghostkeeper (changes requested):
Is it not better to just leave out this flag altogether then? Delete this line of code?

I imagine that would work better when people e.g. change the priority of the front-end process that spawns it. And it's less to maintain.

Ghostkeeper (approved):

PR #792, opened 2022-03-03
[CURA-8943] Update Tool Panel Icons style

Joeydelarago:
Rotate Tool & Scale tool image buttons swapped out for new style _ToolbarButton_ that is used in _Per Model Settings_

Moved Tooltip.qml and ToolbarButton.qml into Uranium so that the plugins/Tools can access these elements.

CURA-8943

### DIFF
--- a/UM/Qt/qml/UM/ToolTip.qml
+++ b/UM/Qt/qml/UM/ToolTip.qml
@@ -9,7 +9,7 @@ import UM 1.5 as UM
 ToolTip
 {
     // Defines the alignment of the content, by default to the left
-    property int contentAlignment: Text.AlignRight
+    property int contentAlignment: UM.Enums.ContentAlignment.AlignRight
 
     property alias tooltipText: tooltip.text
     property alias arrowSize: backgroundRect.arrowSize
@@ -35,7 +35,7 @@ ToolTip
 
     x:
     {
-        if (contentAlignment == Text.AlignLeft)
+        if (contentAlignment == UM.Enums.ContentAlignment.AlignLeft)
         {
             return (label.width + Math.round(UM.Theme.getSize("default_arrow").width * 1.2) + padding * 2) * -1
         }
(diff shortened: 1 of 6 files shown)


UM/Qt/qml/UM/ToolTip.qml
```diff
@@ -0,0 +1,75 @@
...
+
+import QtQuick 2.7
+import QtQuick.Controls 2.3
+
+import UM 1.5 as UM
+
+ToolTip
+{
+    // Defines the alignment of the content, by default to the left
+    property int contentAlignment: Text.AlignRight
```
nallath, line 12: See changes suggested in the Cura ticket.

nallath (changes requested):

UM/Qt/qml/UM/ToolTip.qml
```diff
@@ -0,0 +1,75 @@
...
+    // If the text is not set, just set the height to 0 to prevent it from showing
+    height: label.contentHeight + 2 * UM.Theme.getSize("thin_margin").width
+
+    x:
+    {
+        if (contentAlignment == UM.Enums.ContentAlignment.AlignLeft)
+        {
+            return (label.width + Math.round(UM.Theme.getSize("default_arrow").width * 1.2) + padding * 2) * -1
+        }
+        return parent.width + Math.round(UM.Theme.getSize("default_arrow").width * 1.2 + padding)
```
casperlamboo, line 42: The else case here is always the `ContentAlignment.AlignRight` case, no? I would make this explicit by using a switch statement
```suggestion
        switch (contentAlignment)
        {
            case UM.Enums.ContentAlignment.AlignLeft:
                return (label.width + Math.round(UM.Theme.getSize("default_arrow").width * 1.2) + padding * 2) * -1;
            case UM.Enums.ContentAlignment.AlignRight:
                return parent.width + Math.round(UM.Theme.getSize("default_arrow").width * 1.2 + padding);
        }
```

PR #793, opened 2022-03-04
[CURA-8991] Align input component styling

Joeydelarago:
Update checkboxes, radio and textinput to align with new designs.

CURA-8991

### DIFF
--- a/UM/Qt/qml/UM/CheckBox.qml
+++ b/UM/Qt/qml/UM/CheckBox.qml
@@ -50,13 +50,14 @@ CheckBox
             states: [
                 State {
                     name: "unchecked"
-                    when: control.enabled && control.checkState == Qt.Unchecked
+                    when: control.checkState == Qt.Unchecked
                     PropertyChanges
                     {
                         target: indicator_item
                         width: Math.round(parent.width / 1.5)
                         height: width
-                        opacity: false
+                        visible: false
+                        source: ""
                     }
                 },
                 State {
--- a/UM/Qt/qml/UM/CheckBox.qml
+++ b/UM/Qt/qml/UM/CheckBox.qml
@@ -67,7 +67,7 @@ CheckBox
                         target: indicator_item
                         width: Math.round(parent.width / 1.8)
                         height: width
-                        color: control.enabled ? UM.Theme.getColor("checkbox_mark") : UM.Theme.getColor("checkbox_mark_disabled")
+                        color: control.enabled ? UM.Theme.getColor("checkbox_square") : UM.Theme.getColor("checkbox_mark_disabled")
                         source: UM.Theme.getIcon("Solid")
                     }
                 },
--- a/UM/Qt/qml/UM/CheckBox.qml
+++ b/UM/Qt/qml/UM/CheckBox.qml
@@ -9,7 +9,7 @@ CheckBox
 {
     id: control
     property alias tooltip: tooltip.text
-    enabled: false
+
     hoverEnabled: true
     states: [
         State {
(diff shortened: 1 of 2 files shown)


casperlamboo (approved):

UM/Qt/qml/UM/CheckBox.qml
```diff
@@ -46,20 +51,37 @@ CheckBox
                 State {
                     name: "unchecked"
                     when: control.enabled && control.checkState == Qt.Unchecked
```
casperlamboo, line 53: I see you removed the `control.enabled &&` part everywhere else, why not here?

UM/Qt/qml/UM/CheckBox.qml
```diff
@@ -46,20 +51,37 @@ CheckBox
...
-                    PropertyChanges { target: indicator_item; width: Math.round(parent.width / 1.5); height: width }
-                    PropertyChanges { target: indicator_item; color: UM.Theme.getColor("checkbox_mark"); source: UM.Theme.getIcon("Check", "low"); opacity: false; }
+                    PropertyChanges
+                    {
+                        target: indicator_item
+                        width: Math.round(parent.width / 1.5)
+                        height: width
+                        opacity: false
```
casperlamboo, line 59: this was there before but this might be a good time to tackle this as well.

- why set width and height if the property is not visible
- its a bit weird to set opacity to a bool as the property is a real https://doc.qt.io/qt-5/qml-qtquick-item.html#opacity-prop. I have two suggetsions

```suggestion
                        visible: false
```

```suggestion
                        source: ""
```

casperlamboo (changes requested):

UM/Qt/qml/UM/CheckBox.qml
```diff
@@ -9,19 +9,23 @@ CheckBox
 {
     id: control
     property alias tooltip: tooltip.text
-
+    enabled: false
```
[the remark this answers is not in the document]
Joeydelarago: Accidentally missed some pushes. This was just for testing

UM/Qt/qml/UM/CheckBox.qml
```diff
@@ -45,21 +50,39 @@ CheckBox
...
                 State {
                     name: "unchecked"
-                    when: control.enabled && control.checkState == Qt.Unchecked
-                    PropertyChanges { target: indicator_item; width: Math.round(parent.width / 1.5); height: width }
-                    PropertyChanges { target: indicator_item; color: UM.Theme.getColor("checkbox_mark"); source: UM.Theme.getIcon("Check", "low"); opacity: false; }
+                    when: control.checkState == Qt.Unchecked
+                    PropertyChanges
+                    {
+                        target: indicator_item
+                        width: Math.round(parent.width / 1.5)
```
nallath, line 57: Why change the width if the indicator becomes unchecked?

UM/Qt/qml/UM/CheckBox.qml
```diff
@@ -45,21 +50,39 @@ CheckBox
...
                     name: "partially_checked"
-                    when: control.enabled && control.checkState == Qt.PartiallyChecked
-                    PropertyChanges { target: indicator_item; width: Math.round(parent.width / 1.8); height: width  }
-                    PropertyChanges { target: indicator_item; color: UM.Theme.getColor("checkbox_border"); source: UM.Theme.getIcon("Solid"); opacity: true; }
+                    when: control.checkState == Qt.PartiallyChecked
+                    PropertyChanges
+                    {
+                        target: indicator_item
+                        width: Math.round(parent.width / 1.8)
+                        height: width
```
nallath, line 70: I'm fairly sure that you don't need to set it again. When it's initially set, it creates a binding.

nallath (changes requested):

nallath (approved):

PR #798, opened 2022-03-16
[CURA-9044] Update Profiles page to show active profile in italics

Joeydelarago:
Update italics logic to work with active profile on profiles page.
Fix height binding loop error in Management page

https://github.com/Ultimaker/Cura/pull/11696
CURA-9044


nallath (changes requested):
I don't agree with the solution here. There are a number of issues with it
1. It introduces the Cura namespace into a Uranium component. 
2. It introduces logic about the quality name in a generic page (the Management page is used by other pages than the ProfilesPage)

The fix should just be that the activeId is correctly set in ProfilesPage in Cura

nallath (approved):

PR #809, opened 2022-04-21
CURA-9145 Small Black window appears on macOS

Joeydelarago:
gl_window is used to detect the opengl version on some MacOs versions. This seems to have been unassigned automagically in QT5. With QT6 it was staying open in the background.

The fix here is to assign it to null to remove it from memory.

CURA-9145


nallath (approved):

jellespijker:
The `gl_` was also giving us troubles with the rotate arrows. #811 Luckily @fieldOfView came to the rescue

PR #810, opened 2022-04-22
[5.0b1] Fix TooltipArea

fieldOfView:
This PR fixes the TooltipArea control, which is no longer working because it was using a private QtQuick Controls 1 element. The PR wraps the UM.Tooltip, so tooltips are now styled like the rest of the app. The tooltip does not have an arrow pointing towards the TooltipArea, so we are more free to place oversize tooltips in such a way that they don't pop up in a way that they need to immediately close because they obscure the TooltipArea.

### DIFF
--- a/UM/Qt/qml/UM/TooltipArea.qml
+++ b/UM/Qt/qml/UM/TooltipArea.qml
@@ -3,41 +3,36 @@
 
 import QtQuick 2.4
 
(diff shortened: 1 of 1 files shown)


MariskaMaas:
Thanks! For internal reference CURA-9219

nallath (approved):

PR #812, opened 2022-04-30
[5.0b1] Fix using "Fusion" as the base style for applications

fieldOfView:
This PR changes how applications force using the "fusion" style. Without this, Windows mixes in a "native" style which adds strange overlays on hover which negatively affect readability and break consistency across platforms.

The previous QApplication.setStyle() after initialising the QApplication class no longer seems to work in Qt6 (if it ever worked).

### DIFF
--- a/UM/Qt/QtApplication.py
+++ b/UM/Qt/QtApplication.py
@@ -100,6 +100,9 @@ class QtApplication(QApplication, Application):
         # use Qt Quick Scene Graph "basic" render loop
         os.environ["QSG_RENDER_LOOP"] = "basic"
 
+        # Force using Fusion style for consistency between Windows, mac OS and Linux
+        os.environ["QT_QUICK_CONTROLS_STYLE"] = "Fusion"
+
         super().__init__(sys.argv, **kwargs)
 
         self._qml_import_paths: List[str] = []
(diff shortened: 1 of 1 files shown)


jellespijker:
@fieldOfView might this be related to ultimaker/cura#11874

fieldOfView:
@jellespijker I don't really see how. But if it fixes that, all the better ;-)

fieldOfView:
This PR fixes mouse over effects like this, which make buttons illegible (2x magnified for clarity):
image

Checkboxes also have weird hover states, where a whole different checkmark is blended over the styled checkmark (and a dotted line is added) (4x magnified):
image

Further weirdness in the Marketplace with an illegible tertiary button, and see that tiny external link in the top right of the plugin "card"? Also miraculously fixed by setting "Fusion" as the base theme:
image

jellespijker:
On second thought I don't think they're related, maybe next time I should do more then only read the title, but also look at the code change ;-)

nallath (approved):

nallath:
I can't really test this, but the changes make sense.

fieldOfView:
You could check for adverse effects on Linux; I have not tested that myself.

nallath:
Well, no adverse effects on linux, so I'm just going to merge this.

PR #816, opened 2022-05-06
[CURA-8856] Don't spam the user already running the beta with a notification for said beta.

rburema:
Retrieved beta version would always be > local beta version.

Since we can't store the postfix (only up to 'revision' level) and major.minor.revision is loaded as 3 intergers from (online) latest.json -- the latest BETA version would always be (stoed online and) retrieved as a.b.c (withouth postfix info), but compared against the local a.b.c-beta+whatever (so, with postfix). The algorithm would then quite correctly deduce that a.b.c-beta+whatever was in fact a 'smaller' version than a.b.c (no postfix). Which then would lead to a message that a new beta was avialable _even if the user was already runnng said beta_.

This fixes that, but the underlying issue (not providing a postfix version) still remains. Problems this could cause are when we want to release and announce a second beta, without already incrementing the patch/revision nr. just because of that.

( Sadly this will only fix the issue for whenever we release 5.1.0-beta+1 of course... )

### DIFF
--- a/UM/Version.py
+++ b/UM/Version.py
@@ -111,6 +111,14 @@ class Version:
 
         return self._postfix_version
 
+    def getWithoutPostfix(self) -> "Version":
+        """Returns this as _only_ a major.minor.revision, without the postfix type/version.
+
+        The postfix is everything beyond the patch, like '-beta+1' in 5.0.0-beta+1 -- in this example 5.0.0 is returned.
+        """
+
+        return Version([self._major, self._minor, self._revision])
+
     def hasPostFix(self) -> bool:
         """Check if a version has a postfix."""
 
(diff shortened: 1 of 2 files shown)


fieldOfView:
Will this still notify users of "5.0.0-BETA+1" when "5.0.0" is released?
Or is that irrelevant because the beta notifies the user of all 5.0.0 releases regardless of postfix?

rburema:
> Will this still notify users of "5.0.0-BETA+1" when "5.0.0" is released?
Or is that irrelevant because the beta notifies the user of all 5.0.0 releases regardless of postfix?

Ah no, I think it won't at the moment. That last bit is what _should_ happen of course though.

Ghostkeeper:
Indeed, if latest.json contains 5.0.0 for both the beta and the stable version, it thinks it's already at the latest version even though it's currently in a beta.
I've put a test JSON file on http://dulek.net/work/latest.json

This will need to be fixed.

Ghostkeeper (approved):

Ghostkeeper:
Works correctly this time!

PR #817, opened 2022-05-09
Fix "Zoom Toward Mouse Direction" #6775

seaniepie:
Zoom toward mouse always referenced the mouse position as half the actual position. Mainly due to calculations based on the viewport centre rather than the dimensions and native mouse positions. This patch fixes it by not referencing the centre of the viewport, but to the mouse position directly.

### DIFF
--- a/plugins/Tools/CameraTool/CameraTool.py
+++ b/plugins/Tools/CameraTool/CameraTool.py
@@ -288,14 +288,12 @@ class CameraTool(Tool):
             move_vector = Vector(0.0, 0.0, 1.0)
 
             if event is not None and self._zoom_to_mouse:
-                viewport_center_x = QtApplication.getInstance().getRenderer().getViewportWidth() / 2
-                viewport_center_y = QtApplication.getInstance().getRenderer().getViewportHeight() / 2
+                viewport_width = QtApplication.getInstance().getRenderer().getViewportWidth()
+                viewport_height = QtApplication.getInstance().getRenderer().getViewportHeight()
+                
                 main_window = cast(MainWindow, QtApplication.getInstance().getMainWindow())
-                mouse_diff_center_x = viewport_center_x - main_window.mouseX
-                mouse_diff_center_y = viewport_center_y - main_window.mouseY
-
-                x_component = mouse_diff_center_x / QtApplication.getInstance().getRenderer().getViewportWidth()
-                y_component = mouse_diff_center_y / QtApplication.getInstance().getRenderer().getViewportHeight()
+                x_component = main_window.mouseX / viewport_width
+                y_component = main_window.mouseY / viewport_height
 
                 move_vector = Vector(x_component, -y_component, 1)
                 move_vector = move_vector.normalized()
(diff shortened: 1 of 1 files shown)


nallath:
This breaks zoom towards mouse direction on my machine (Ubuntu). The original code does work correctly on my device.

seaniepie:
@nallath, I think this latest commit allows for Mac/Win AND Linux in an elegant enough way. Do you have Win and Mac to test on also or only Linux?

nallath:
> @nallath, I think this latest commit allows for Mac/Win AND Linux in an elegant enough way. Do you have Win and Mac to test on also or only Linux?

I will ask one of the other developers :)

rburema:
Hi, designated Windows dev here. I initially observed that, on my machine, our method works, but yours didn't.

I have a theory, however, based on some tests with changing the scaling the OS applies globally (you can set that in the 'Display Settings').

Our method seems to work fine when the scaling of the OS is set to 100%. In fact, on 100% ours works, and your change doesn't seem to, at least on my machine.

Both our method and yours break when I set it to (say) 175% scaling. Your method seems to work around 125%, 150%, which is recommended for some (laptop) screens, so Windows auto-selects that.

So I don't think this is the right way to fix this, especially since we don't even know if mac works or not (behaviour is closer to linux usually). Also, we have standardized methods for Win/Mac/Linux look for `Platform.isLinux()` for example.

seaniepie:
I think it must be down to the resolution (for a Mac laptop this is often 2x (ish) due to their 'retina' screens). This is probably not being accounted for but only seems to affect zoom scaling when it comes to dealing with the mouse position. Probably because it is only there that the code is specifically interested in the mouse position. I'll do a bit more digging and then close this pull request.

seaniepie:
Part of Qt seems to help with this using [devicePixelRatio](https://doc.qt.io/qtforpython/PySide6/QtGui/QScreen.html#PySide6.QtGui.PySide6.QtGui.QScreen.devicePixelRatio)
If we instead use a simple multiplier on the mouse position derived from the `devicePixelRatio()`, we should be able to fix for any platform, display and resolution scaling.

seaniepie:
@nallath @rburema, Latest commit makes use of `devicePixelRatio`. Can you test on your OSs too? Can an independent Mac dev also test for us? Preferably someone with a Retina display that can switch between pixelRatios.

plugins/Tools/CameraTool/CameraTool.py
```diff
@@ -288,14 +288,18 @@ def _zoomCamera(self, zoom_range: float, event: Optional[Event] = None) -> None:
...
+                viewport_center_x = viewport_width / 2
+                viewport_center_y = viewport_height / 2
+                
                 main_window = cast(MainWindow, QtApplication.getInstance().getMainWindow())
-                mouse_diff_center_x = viewport_center_x - main_window.mouseX
-                mouse_diff_center_y = viewport_center_y - main_window.mouseY
-
-                x_component = mouse_diff_center_x / QtApplication.getInstance().getRenderer().getViewportWidth()
-                y_component = mouse_diff_center_y / QtApplication.getInstance().getRenderer().getViewportHeight()
+                window_PixelRatio = main_window.devicePixelRatio()
```
fieldOfView, line 297: Code style: `window_pixel_ratio`

seaniepie: Cool, thanks for the heads up

nallath:
The latest version seems to work correctly for me (Ubuntu 21)

rburema:
Ah sorry, this one fell between the cracks it seems. Yes it works now!

seaniepie:
Yay. Thanks

PR #818, opened 2022-05-10
Render labels using `Text.QtRendering` on OSX

casperlamboo:
Fonts were looking a bit to thick on when using `Text.NativeRendering`, so using `Text.QtRendering` instead. After this the font weight looks identical to figma (as far as I can see).

In this commit I also changed all `Label`'s to `UM.Label`'s and removed default properties where I could.

CURA-9154


fieldOfView:
> Fonts were looking a bit to thick

Are you sure that is not caused by some (stacked) non-whole-pixel alignment?

nallath (approved):

PR #822, opened 2022-05-20
[SEC-255] [CURA-8966] In secure version: disable loading themes from unbundled folders.

rburema:
This was a bit more tricky then it at first seemed, since the information whether this is a 'secure version' comes from the user-application and is not known in the Uranium library. This is not normally such a point, but both the theme and the preferences objects are loaded _very_ early in the process, and that information needs to be injected before then. (Well, in the case of the Theme object it's less important, since in the implementation chosen that is now security wise at least only in charge of whether or not to even show the theme as selectable in the interface, so that it only needs to be aware of the 'security' status any time before the user can see a preference screen, but not necessarily earlier.)

See also frontend PR.

### DIFF
--- a/UM/Preferences.py
+++ b/UM/Preferences.py
@@ -2,7 +2,7 @@
 # Uranium is released under the terms of the LGPLv3 or higher.
 
 import configparser
-from typing import Any, Dict, IO, Optional, Tuple, Union
+from typing import Any, Callable, Dict, IO, Optional, Tuple, Union
 
 from UM.Logger import Logger
 from UM.MimeTypeDatabase import MimeTypeDatabase, MimeType #To register the MIME type of the preference file.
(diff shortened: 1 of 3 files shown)


rburema:
@fieldOfView  Yes, a theme (as of 4.13 at least) can still contain QML, which can contain javascript, which can turn into code injection.

(Note that w.r.t. Cura, this check will only trigger when IsEnterpriseVersion is on.)

The reason the bundled themes are not a risk, is that we assume that anything that is in the install is properly secured by an administrator, but that this doesn't necessarily go for the user folders/configs.

Joeydelarago (approved):

fieldOfView:
> Yes, a theme (as of 4.13 at least) can still contain QML,

I already forgot all about that.

But the same is not true past Cura 5, so why there's also a PR to do this in master is still a mystery to me.

PR #826, opened 2022-05-27
[CURA-8968] Also check afterwards.

rburema:
Seems like we _may_ start the socket for listening for the engine first, depending on threading, which doesn't open the engine itself, but does mess with the order of the preference reading. Also rename settings -> preferences.

SEC-257 | CURA-8968

### DIFF
--- a/UM/Preferences.py
+++ b/UM/Preferences.py
@@ -36,7 +36,7 @@ class Preferences:
         self._preferences = {}  # type: Dict[str, Dict[str, _Preference]]
         self._untrusted_preferences: Dict[(str, str), Callable] = {}
 
-    def indicateUntrustedPreference(self, group: str, key: str, eval_func: Callable) -> None:
+    def indicateUntrustedPreference(self, group: str, key: str, eval_func: Callable[[str], bool]) -> None:
         """Indicates that the value of this setting should be evaluated before acceptance, and otherwise not loaded."""
         self._untrusted_preferences[(group, key)] = eval_func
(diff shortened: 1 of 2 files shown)


UM/Preferences.py
```diff
@@ -34,11 +34,17 @@ def __init__(self) -> None:
 
         self._parser = None  # type: Optional[configparser.ConfigParser]
         self._preferences = {}  # type: Dict[str, Dict[str, _Preference]]
-        self._untrusted_settings: Dict[(str, str), Callable] = {}
+        self._untrusted_preferences: Dict[(str, str), Callable] = {}
 
-    def indicateUntrustedSetting(self, group: str, key: str, eval_func: Callable) -> None:
+    def indicateUntrustedPreference(self, group: str, key: str, eval_func: Callable) -> None:
```
nallath, line 39:
```suggestion
    def indicateUntrustedPreference(self, group: str, key: str, eval_func: Callable[[str], bool]) -> None:
```
Although it could be that this trips up if you put a lambda in it. I'm not entirely sure of that.

nallath (changes requested):

UM/Preferences.py
rburema: Well it runs, and is furthermore a valid way of typing lambda's (at least after you've put them in a variable, the lambda can't be typed that way during declaration) according to this stackoverflow answer: https://stackoverflow.com/a/33833896

PR #829, opened 2022-06-03
 Use "required packages" metadata to load a project file with a third party material

casperlamboo:
see https://github.com/Ultimaker/Cura/pull/12434

CURA-6990

### DIFF
--- a/UM/Qt/qml/UM/Dialog.qml
+++ b/UM/Qt/qml/UM/Dialog.qml
@@ -6,8 +6,7 @@ import QtQuick.Window 2.2
 import QtQuick.Layouts 1.3
 import QtQuick.Dialogs
 
-import UM 1.0 as UM
-
+import UM 1.5 as UM
 
 Window
 {
@@ -35,6 +34,9 @@ Window
     property alias backgroundColor: background.color
 
     property real buttonSpacing: 0
+
+    property bool buttonWarning: false
+    property alias buttonWarningText: warningText.text
 
     signal accepted();
     signal rejected();
@@ -84,7 +86,7 @@ Window
                 rightMargin: base.margin;
                 top: parent.top;
                 topMargin: base.margin;
-                bottom: buttonRow.top;
+                bottom: footer.top;
                 bottomMargin: base.margin;
             }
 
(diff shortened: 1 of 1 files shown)


UM/Qt/qml/UM/Dialog.qml
```diff
@@ -96,29 +98,66 @@ Window
...
-                id: leftButtonRow
-                Layout.alignment: Qt.AlignLeft
-                spacing: base.buttonSpacing
-            }
-
-            RowLayout {
-                id: rightButtonRow
-                Layout.alignment: Qt.AlignRight
-                spacing: base.buttonSpacing
+        Rectangle
```
rburema, line 101: This is in the base class of UM Dialog itself? So everywhere we use UM.Dialog will have this specific footer in it, even making a reference to the icon size for 'extruders', a concept of which the Uranium library doesn't have any idea?

(As a side note: This also breaks the SDK for anyone currently using UM.Dialog. We can live with that of course as we're not in feature freeze, but it's something that you could keep in mind as well.)

I'd personally make a new base-class in Cura if this is needed ... or make this a little more abstract, so you can just either insert your own children into the footer (if you still think an API break is needed), or adjust the buttonrow to the users liking.

We should also notify the testers of this change, as some of the last bugs we (tried to) fix before 5.0 to-do with the size of dialogs not adjusting to the contents(-sizes) had to do with this class, and specifically the buttons as well, IIRC.

casperlamboo: I agree with your feedback, so I tried something else. I removed the warning square from this component, and added properties to set the header and footer components of the dialog. https://github.com/Ultimaker/Uranium/pull/829/commits/8582976f6aa3f9f2deabf6efb6aec33518d096b1. I think these terms are a little more generic then the warning square boolean and thus more suited in this generic component. 

Then I this warning area can be placed within this newly introduced footer https://github.com/Ultimaker/Cura/pull/12434/commits/2f22002fd59e2be44934c0ff5c904c93a03fd690#diff-fe2a08cf2edd34d1b8105969082c4b23e2186f533c15d8b1588bf9a8ca5030a5.

I tried to keep the functionality the same and not introduce a _breaking change_; but this would still introduce a minor increase SDK, and all dialog components should be tested again. However I do think this is the right approach as not changing the dialog component would force us to copy the footer-button functionality to the Workspace Component.

Let me know what you think.

rburema: Looks good! I'm ok with smaller changes like this. The important thing is that the dependency on Cura is gone, and that people can access the left- and right-buttons under the same name, etc.

rburema (approved):

PR #834, opened 2022-06-17
[CURA-9321] Can't open files with different encodings

Joeydelarago:
This PR fixes two issues

1. Characters inside obj comments in incorrect encoding causing the file not to be loaded
The solution is to simply ignore any characters we can’t parse. All the object information in these files is standard ASCII characters, leaving only the comments that can be these un-decodable characters.  We can safely ignore these characters in comments and still load the file this way.

2. Not using numpy-stl to load stl files
This issue was caused by numpy-stl being missing from our builds causing the backup method of reading to be used. This backup was added due to a bug on linux with numpy-stl (CURA-7154). In the comment of this ticket it says it is okay to revert this after we upgrade python to 3.9 (which we have). I’ve removed this backup method of reading stl and now all OS should work as expected. This should be tested on Linux with the models from this ticket to be sure though


plugins/FileHandlers/STLReader/STLReader.py
```diff
@@ -15,23 +14,8 @@
...
-use_numpystl = False
-
-try:
-    # Work around for CURA-7154. Some models crash with a segfault, but only when packed.
-    if platform.system() != "Linux":
-        import stl  # numpy-stl lib
-        import stl.mesh
-
-        # Increase max count. (100 million should be okay-ish)
-        stl.stl.MAX_COUNT = 100000000
```
Ghostkeeper, line 28: We should still have the increased maximum vertex count. Otherwise it can't open meshes with more than 10 million vertices.

plugins/FileHandlers/STLReader/STLReader.py
```diff
@@ -47,28 +31,11 @@ def __init__(self) -> None:
...
-            try:
-                self._loadWithNumpySTL(file_name, mesh_builder)
-                file_read = True
-            except:
-                Logger.logException("e", "Reading file failed with Numpy-STL!")
-
-        if not file_read:
-            Logger.log("i", "Using legacy code to load STL data.")
-            f = open(file_name, "rb")
-            if not self._loadBinary(mesh_builder, f):
```
Ghostkeeper, line 63: The `_loadBinary` and `_loadAscii` functions are no longer used now, and should be removed.

plugins/FileHandlers/STLReader/tests/TestStlReader.py
```diff
@@ -16,14 +16,13 @@ def test_readASCII():
         result = reader.read(ascii_path)
     assert result
 
-    if STLReader.use_numpystl:
-        # If the system the test runs on supports numpy stl, we should also check the non numpy stl option.
-        f = open(ascii_path, "rt", encoding = "utf-8")
-        mesh_builder = MeshBuilder()
-        reader._loadAscii(mesh_builder, f)
-        mesh_builder.calculateNormals(fast=True)
+    # If the system the test runs on supports numpy stl, we should also check the non numpy stl option.
```
Ghostkeeper, line 19: No, we don't. Even other systems don't use them any more now.

plugins/FileHandlers/STLReader/tests/TestStlReader.py
```diff
@@ -32,13 +31,12 @@ def test_readBinary():
     with patch("UM.Application.Application.getInstance"):
         result = reader.read(binary_path)
 
-    if STLReader.use_numpystl:
-        # If the system the test runs on supporst numpy stl, we should also check the non numpy stl option.
-        f = open(binary_path, "rb")
-        mesh_builder = MeshBuilder()
-        reader._loadBinary(mesh_builder, f)
-        mesh_builder.calculateNormals(fast=True)
+    # If the system the test runs on supporst numpy stl, we should also check the non numpy stl option.
```
Ghostkeeper, line 34: No, we don't. Even other systems don't use them any more now.

Ghostkeeper (changes requested):

Ghostkeeper (approved):

PR #835, opened 2022-06-28
[Cura 12625] Added decimals capability to the RotateToolHandle Hint

Fabbro03:
Added decimals capability to the RotateToolHandle Hint without modifing the integer displaying of snap rotation.

I think this change can improve the user experience when using rotation without snap, which with just displaying the unit could cause some issues (see https://github.com/Ultimaker/Cura/issues/12625).

Can it be integrated within the next version of Cura?


Fabbro03:
Tested on Cura 5.0.0

Rotation with snap:
Rotation with snap

Rotation without snap:
rotation without snap

Joeydelarago (approved):

PR #844, opened 2022-08-15
[CURA-9272] Read semver postfix from latest.json

Joeydelarago:
It will now be possible to notify the user of multiple beta releases, instead of just the first one by using the prerelease postfix and version number defined in semver.


plugins/UpdateChecker/UpdateChecker.py
```diff
@@ -32,7 +32,7 @@ class UpdateChecker(Extension):
     The plugin is currently only usable for applications maintained by Ultimaker. But it should be relatively easy
     to change it to work for other applications.
     """
-    url = "https://software.ultimaker.com/latest.json"
+    url = "https://raw.githubusercontent.com/Ultimaker/Uranium/CURA-9272_semver_postfix/tests/latest.json"
```
jellespijker, line 35: Why do we even have this hard coded? and why should Uranium be aware of Cura being able to update?
I would suggest to add this link to the https://github.com/Ultimaker/Cura/blob/main/cura/UltimakerCloud/UltimakerCloudConstants.py . You can also add it as an option to the conanfile.py of Cura. This would allow us to pass the endpoint dynamically, the same way we switch staging endpoints.

Make sure that the logic in this repo doesn't fail when an empty/default url exist here.

UM/Version.py
```diff
@@ -111,14 +111,6 @@ def getPostfixVersion(self) -> int:
 
         return self._postfix_version
 
-    def getWithoutPostfix(self) -> "Version":
```
jellespijker, line 114: Isn't this part of the SDK? If so why remove this, since that would introduce an SDK break. Rather mark it as deprecated then

plugins/UpdateChecker/UpdateChecker.py
```diff
@@ -78,6 +78,14 @@ def _extractVersionAndURLFromData(cls, data: Dict, application_name: str) -> Tup
         if os not in data[application_name]:
             return None, None
 
+        if "postfix_type" in data[application_name][os] and "postfix_version" in data[application_name][os]:
+            # Prerelease versions include the extra postfix_type and postfix_version keys
+            return Version([int(data[application_name][os]["major"]),
```
jellespijker, line 83: I would feel more saver when we apply a bit of defensive programming here, with a try-catch block guarding against weird input

tests/TestVersion.py
```diff
@@ -65,6 +65,20 @@ def test_versionPostfix():
     assert version <= Version("1.2.3-alpha.5")
     assert version < Version("1.2.3-alpha.5")
 
+def test_postfix_lastest_json_format():
```
jellespijker, line 68: - What happens if the pre-release-tag version doesn't contain and number `<pre_release_tag>.1` like it does now?
- What happens with the build meta data after the `+`

jellespijker:
We should update the workflow ( https://github.com/Ultimaker/Cura/blob/830adcea1bef4b4ff05273d22e3c0f1f27e978f9/.github/workflows/conan-recipe-version.yml#L74 ) for our versioning number to output the same `<major>.<minor>.<patch>-<pre-release-tag>.<postfix>+<build_metadata>`

I know the url was already hard coded in here but I think we should move it out of this repo and in to the cloud constants in Cura and provide it as an Conan option (same as we do with staging)

Also double check if removing that function doesn't break API, if so add it back and mark it as deprecated

jellespijker (changes requested):

plugins/UpdateChecker/UpdateChecker.py
Joeydelarago: I honestly have no idea, maybe this whole file should be moved into Cura? Otherwise, it will not be possible to fetch any configuration information without coupling Uranium to Cura.

casperlamboo (approved):

casperlamboo (approved):

PR #848, opened 2022-09-01
Add the option to drop down models individually 

Piezoid:
Counterpart of frontend PR  Ultimaker/Cura#13188
Should resolve Ultimaker/Cura#8117.

Few things to note:
* First commit refactors utility method used in TranslateTool. They provide access to SceneNodeSettings on selected object, and are reused for both `LockPosition` and the new `AutoDropDown` node settings,
* Unsure if SceneNodeSettings must be a string, this requires a [`bool -> str -> bool` round-trip](https://github.com/Piezoid/Cura/blob/111136cf493c7d2a567d68bfb9ceffe90f340453/cura/PlatformPhysics.py#L84) that may be unnecessary. I presume that strings are required for serialization? Anyway, I mimicked the type of the existing `LockPosition` setting,
* The default value is set from the Application preference `physics/automatic_drop_down` and thus requires an access to this setting from Uranium, I'm unsure if that's authorized.


nallath:
> The default value is set from the Application preference physics/automatic_drop_down and thus requires an access to this setting from Uranium, I'm unsure if that's authorized.

In the case that it's not set, the value will return "None", which resolves itself to False. So although you could argue that it's using something from Cura, I think it's fine in this case.

plugins/Tools/TranslateTool/TranslateTool.py
```diff
@@ -206,28 +208,22 @@ def setLockPosition(self, value: bool) -> None:
...
-        if total_size == false_state_counter:  # No locked positions
-            return False
-        elif total_size == true_state_counter:  # All selected objects are locked
-            return True
-        else:
-            return "partially"  # At least one, but not all are locked
+        model movement on the build plate.
+        :param value: The setting state.
+        """
+        self.setSettingToSelection(SceneNodeSettings.AutoDropDown, str(value))
```
nallath, line 222: I think that this should also forcefully trigger a "sceneChanged" event, so that if this is changed by the user that it is automatically re-calculated.

RIght now if you disable drop objects in preferences and drag an object up and then click it in the translate tool, it remains floating for a bit (until anything else in the scene is updated).

nallath (changes requested):

nallath:
For our reference CURA-9600

Piezoid:
- Added the sceneChanged event on `setSettingToSelection`, so it should work for locking too,
- Refactored the QML to use tri-states checkboxes instead of the custom logic,
- I ended up preferring native python types instead of  string for ScenNodeSettings. [Cura was already using LockPosition as a bool](https://github.com/Ultimaker/Cura/blob/c7d7dd1/cura/PlatformPhysics.py#L89), so I figured that avoiding strings is less confusing. `getSettingToSelection` now returns `None` when the selection's truth values don't agree (was `"partially"` previously).

PR #849, opened 2022-09-05
Dont show toast message for every time a message is spawned

nallath:
CURA-9471


UM/Qt/QtApplication.py
```diff
@@ -391,9 +391,6 @@ def showMessage(self, message: Message) -> None:
                 message.setInactivityTimer(QTimer())
                 self.visibleMessageAdded.emit(message)
 
-        # also show toast message when the main window is minimized
-        self.showToastMessage(self._app_name, message.getText())
```
Joeydelarago, line 395: This seems to be the only place showToastMessage is called. Should we add @deprecated to showToastMessage()?

casperlamboo (approved):

PR #851, opened 2022-09-14
[CURA-9535] Some trimesh operations require networkx

rburema:
Seems networkx itself is low on dependencies itself. It does require some optional ones for e.g. it's linalg sub-package, but those are numpy and scipy, which we already have.


jellespijker:
Disregard my comment on the Cura repo

PR #852, opened 2022-09-15
Prevent "files changed" popup when file did not change

casperlamboo:
Solved simply adding one additional check before showing the popup where we look at the “last modified” timestamp of the model. If this timestamp is too far in the past we don’t show the popup. This difference in time was around ~0.5ms when testing, but set to 10000ms (10s) just to be certain.

Fixes: https://github.com/Ultimaker/Cura/issues/12322

CURA-9417

### DIFF
--- a/UM/Scene/Scene.py
+++ b/UM/Scene/Scene.py
@@ -21,7 +21,7 @@ if Platform.isWindows():
     from PyQt6.QtCore import QEventLoop  # Windows fix for using file watcher on removable devices.
 
 i18n_catalog = i18nCatalog("uranium")
-
+from time import time
 
 @signalemitter
 class Scene:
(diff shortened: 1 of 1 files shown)


UM/Scene/Scene.py
```diff
@@ -175,10 +175,14 @@ def _onFileChanged(self, file_path: str) -> None:
         try:
             if os.path.getsize(file_path) == 0:  # File is empty.
                 return
-        except EnvironmentError:  # Or it doesn't exist any more, or we have no access any more.
+        except EnvironmentError:  # Or it doesn't exist anymore, or we have no access anymore.
+            return
+
+        # On Mac the reload file dialog would pop up for recently downloaded files, even if the file did not change
```
Joeydelarago, line 181:
```suggestion
        # Fixes CURA-9417
        # On Mac the reload file dialog would pop up for recently downloaded files, even if the file did not change
```

Thought it would be handy to mention the ticket here in case the comes up again.

Joeydelarago (approved):

PR #853, opened 2022-09-20
Fix RecursionError when pickling container to cache

Piezoid:
Currently containers for many printer fails to pickle.
Increasing the recursion limit to 3000 fixes this.
I looked at the disassembly with `python -m pickletools`, there doesn't seem to be anything suspicious. The depth of the object tree is simply to deep for the default recursion limit.

### DIFF
--- a/plugins/LocalContainerProvider/LocalContainerProvider.py
+++ b/plugins/LocalContainerProvider/LocalContainerProvider.py
@@ -1,6 +1,7 @@
 # Copyright (c) 2022 Ultimaker B.V.
 # Uranium is released under the terms of the LGPLv3 or higher.
 
+import sys # For setrecursionlimit
 import os  # For getting the IDs from a filename.
 import pickle  # For caching definitions.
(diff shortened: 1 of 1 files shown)


Joeydelarago:
Hey @Piezoid, can I ask how you hit the 1000 recursion limit when pickling? Could you provide an example? Otherwise, I don't see any downsides to this change. I would just like to test first before merging.

Piezoid:
I don't think I was doing anything special.
I was getting some warnings in the logs:
```
2022-10-01 16:00:25,286 - WARNING - [MainThread] LocalContainerProvider.LocalContainerProvider._saveCachedDefinition [308]: The definition cache for definition ultimaker_s5 failed to pickle.
2022-10-01 16:00:28,200 - WARNING - [MainThread] LocalContainerProvider.LocalContainerProvider._saveCachedDefinition [308]: The definition cache for definition fdmprinter failed to pickle.
...
```

I'll try to reproduce it again, with a recent versions and empty user directory.

Joeydelarago (approved):

PR #858, opened 2022-10-18
Build mo translation files when creating conan package

jellespijker:
Added a build step which uses the Conan package gettext as a tool_requires to convert the po files to mo files and store these in the resources folder

When on Windows the msys2 recipe is used to ensure that we have a bash environment to run gnu gettext

Creating the mo files as part of the conan package will ensures that we no longer need to store them in the cura-binary-data and generate them manually. This should result in less risk of human error during the release cycle

See also:
- https://github.com/Ultimaker/Cura/pull/13596
- https://github.com/Ultimaker/cura-binary-data/pull/26

### DIFF
--- a/.github/workflows/unit-test.yml
+++ b/.github/workflows/unit-test.yml
@@ -93,7 +93,11 @@ jobs:
 
             -   name: Install Linux system requirements
                 if: ${{ runner.os == 'Linux' }}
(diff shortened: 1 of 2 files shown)


Joeydelarago (approved):

jellespijker:
I think the tests are failing due to our versioning workflow and are unrelated to the changes in this PR. I'm gonna be super naughty and merge this to main.

If it ends up broken on main, I will fix it from there or move to Australia depending on the severity in which I broke it.

PR #862, opened 2022-12-05
[CURA-9793] Extend recommended print settings

Joeydelarago:
Added more generic components for use in Cura.

See https://github.com/Ultimaker/Cura/pull/13972 for more details


UM/Qt/qml/UM/ComponentWithIcon.qml
```diff
@@ -0,0 +1,59 @@
+// Copyright (c) 2018 Ultimaker B.V.
+// Cura is released under the terms of the LGPLv3 or higher.
```
jellespijker, line 2:
```suggestion
// Copyright (c) 2022 UltiMaker
// Cura is released under the terms of the LGPLv3 or higher.
```

UM/Qt/qml/UM/ComponentWithIcon.qml
```diff
@@ -0,0 +1,59 @@
...
+
+import QtQuick 2.7
+import QtQuick.Controls 2.0
+import QtQuick.Layouts 1.3
+
+import UM 1.5 as UM
+import Cura 1.0 as Cura
```
jellespijker, line 9:
```suggestion
```
Bad @Joeydelarago this made me :cry: a bit

UM/Qt/qml/UM/Slider.qml
```diff
@@ -0,0 +1,71 @@
...
+        width: parent.width - UM.Theme.getSize("print_setup_slider_handle").width
+        anchors.horizontalCenter: parent.horizontalCenter
+        anchors.verticalCenter: parent.verticalCenter
+        color: UM.Theme.getColor("lining")
+
+        Repeater
+        {
+            id: repeater
+            anchors.fill: parent
+            model: 11
```
jellespijker, line 23: What is `11` and will it always be `11` for a generically named component named `Slider`

UM/Qt/qml/UM/Slider.qml
```diff
@@ -0,0 +1,71 @@
...
+import UM 1.7 as UM
+
+Slider
+{
+    id: slider
+    //Draw line
+    background: Rectangle
+    {
+        id: backgroundLine
+        height: UM.Theme.getSize("print_setup_slider_groove").height
```
jellespijker, line 13: Maybe we should have a `slider_groove` theme key-value and `print_setup_slider_groove` inherits from that.

Same goes for the items entries below.

The naming feels just to specific for the UM component `Slider`

UM/Qt/qml/UM/Switch.qml
```diff
@@ -0,0 +1,50 @@
+// Copyright (c) 2022 UltiMaker
```
jellespijker, line 1: I think this component is the best written piece of code, I have ever seen! Good job @Joeydelarago

UM/Qt/qml/UM/TextField.qml
```diff
@@ -0,0 +1,97 @@
+//Copyright (c) 2022 UltiMaker B.V.
+//Cura is released under the terms of the LGPLv3 or higher.
```
jellespijker, line 2:
```suggestion
// Copyright (c) 2022 UltiMaker
// Uranium is released under the terms of the LGPLv3 or higher.
```

jellespijker (changes requested):

UM/Qt/qml/UM/Slider.qml
Joeydelarago: The key value only works for colors (I think). I agree the naming is not right, I'll fix that.

UM/Qt/qml/UM/Slider.qml
Joeydelarago: This is some QML magic. It creates 11 empty model items. These are used to generate the dots in the background of the slider, 1 for each model. 

I tried using an alias or property for this value but it does not work for some reason. If we want more/less ticks in the future I think we can figure it out then.

Joeydelarago: I was mistaken. I added an alias for this value at the top of the file

jellespijker (approved):

plugins/Tools/ScaleTool/ScaleTool.qml
```diff
@@ -1,9 +1,10 @@
 // Copyright (c) 2021 Ultimaker B.V.
```
casperlamboo, line 1:
```suggestion
// Copyright (c) 2022 UltiMaker
```

UM/Qt/qml/UM/Validators/IntValidator.qml
```diff
@@ -0,0 +1,11 @@
+// Copyright (c) 2022 UltiMaker
+// Cura is released under the terms of the LGPLv3 or higher.
+
+import QtQuick 2.15
+
+RegularExpressionValidator
+{
+    property int maxNumbers: 12
+
+    regularExpression: new RegExp("^-?[0-9]{0,%0}$".arg(maxNumbers))
```
casperlamboo, line 10: Now the empty string parses as a valid number, do we want this?
```suggestion
    regularExpression: new RegExp("^-?[0-9]{1,%0}$".arg(maxNumbers))
```

UM/Qt/qml/UM/Validators/IntListValidator.qml
```diff
@@ -0,0 +1,9 @@
+// Copyright (c) 2022 UltiMaker
+// Cura is released under the terms of the LGPLv3 or higher.
+
+import QtQuick 2.15
+
+RegularExpressionValidator
+{
+    regularExpression: new RegExp("^\[?(\s*-?[0-9]{0,11}\s*,)*(\s*-?[0-9]{0,11})\s*\]?$")
```
casperlamboo, line 8: The following would parse as a number
```
[, , , , , , , ]
```
```suggestion
    regularExpression: new RegExp("^\[?(\s*-?[0-9]+\s*,)*(\s*-?[0-9]+)\s*\]?$")
```

UM/Qt/qml/UM/Switch.qml
```diff
@@ -0,0 +1,50 @@
...
+Switch
+{
+    id: control
+    property double indicatorMargin: 2
+
+    indicator: Rectangle
+    {
+        id: switchBackground
+        implicitWidth: 32
+        implicitHeight: 16
```
casperlamboo, line 18: Will these scale as expected with different resolution screens?

UM/Qt/qml/UM/Slider.qml
```diff
@@ -0,0 +1,132 @@
+import QtQuick 2.7
```
casperlamboo, line 1: I don't mind this, but previously these kind of components did not live in uranium. I'm fine that this lives here, I think it is logical. But I think this should be a consciousness decision and that the rest of the team should be aware.

UM/Qt/qml/UM/TextField.qml
```diff
@@ -0,0 +1,97 @@
...
+    property alias unit: unitLabel.text
+    property alias liningColor: background.liningColor
+    property alias borderColor: background.borderColor
+    property alias unitText: unitLabel.text
+
+    // States added in this component take precedence over states defined in children. You can force use child states
+    // over these states by setting this value to true. Look at SingleSettingTextField states for an example.
+    property bool overrideState: false
+
+    renderType: Qt.platform.os == "osx" ? Text.QtRendering : Text.NativeRendering
```
casperlamboo, line 23:
```suggestion
    renderType: Qt.platform.os === "osx" ? Text.QtRendering : Text.NativeRendering
```

UM/Qt/qml/UM/TextField.qml
```diff
@@ -0,0 +1,97 @@
...
+    {
+        id: background
+
+        UM.Label
+        {
+            id: unitLabel
+            anchors.right: parent.right
+            anchors.rightMargin: Math.round(UM.Theme.getSize("setting_unit_margin").width)
+            anchors.verticalCenter: parent.verticalCenter
+            visible: text != ""
```
casperlamboo, line 43:
```suggestion
            visible: text !== ""
```

UM/Qt/qml/UM/TextField.qml
```diff
@@ -0,0 +1,97 @@
...
+            {
+                target: background
+                liningColor: UM.Theme.getColor("text_field_border_active")
+                borderColor: UM.Theme.getColor("text_field_border_active")
+            }
+        },
+        State
+        {
+            name: "hovered"
+            when: (control.hovered && !control.activeFocus) && !overrideState
```
casperlamboo, line 89: These are equivalent andI think in this case removing the parentesis make it more readable
```suggestion
            when: control.hovered && !control.activeFocus && !overrideState
```

casperlamboo (approved):

PR #874, opened 2023-05-20
Fix SimpleButton use of UM.ColorImage

fieldOfView:
This PR fixes loading Cura on some systems where Qt does not like using `ColorImage` without a specific prefix of the `UM` package namespace.

Fixes https://github.com/Ultimaker/Cura/issues/15524


nallath (approved):

jellespijker (approved):
LGTM;

Thanks for fixing it so quickly @fieldOfView

fieldOfView:
It still puzzles me how the official builds did not fail on this. But oh well, it is fixed now.

PR #876, opened 2023-05-30
PP-285 add index function to safe compile

pkuiper-ultimaker:
Allow "index" function in save compile. Needed for PP-285


jellespijker (approved):
LGTM

PR #879, opened 2023-06-09
Not all errors in the database are a 'DatabaseError' apparently.

rburema:
Discovered this when we all had 5.4-alpha cache DB's on windows that didn't work with a late pre-beta, that crashed instead of recreating the sqlite3 cache database on error.


jellespijker (approved):
LGTM

Issue #765, opened 2022-01-03
UM/PluginRegistry.py:4: DeprecationWarning: the imp module is deprecated...


hroncok:
Just a heads up that I see the following warning on Python 3.10 when I start Cura 4.12.1:

```
/usr/lib/python3.10/site-packages/UM/PluginRegistry.py:4: DeprecationWarning: the imp module is deprecated in favour of importlib and slated for removal in Python 3.12; see the module's documentation for alternative uses
  import imp
```

Ghostkeeper:
This warning has been there since Python 3.4 in fact. Since the start of Cura 2's development, when the plug-in system was introduced. They now give a targeted version when it will be removed in stead of just "in the future".

A long time ago we did try changing to importlib but didn't have a lot of success. It may be worth another try, but for the moment it doesn't give us any trouble yet. I think this'll be one of those things that we'll cross when we get there.

hroncok:
This was actually removed from Python 3.12.

hroncok:
Fixed in https://github.com/Ultimaker/Uranium/pull/915

PR #885, opened 2023-07-09
Fix asserts for called once in Python 3.12

hroncok:
E               AttributeError: 'called_once_with' is not a valid assertion. Use a spec for the mock if 'called_once_with' is meant to be an attribute.. Did you mean: 'assert_called_once_with'?
    
        FAILED tests/TestBackend.py::test__onSocketStateChanged_listening - Attribute...
        FAILED tests/TestBackend.py::test_onSocketStateChanged_connected - AttributeE...
        FAILED tests/Jobs/TestJob.py::test_start - AttributeError: 'called_once_with'...
        FAILED tests/Jobs/TestJob.py::test_cancel - AttributeError: 'called_once_with...

### DIFF
--- a/tests/Jobs/TestJob.py
+++ b/tests/Jobs/TestJob.py
@@ -32,7 +32,7 @@ def test_start():
     job_queue = MagicMock()
     with patch("UM.JobQueue.JobQueue.getInstance", MagicMock(return_value = job_queue)):
         job.start()
-    job_queue.add.called_once_with(job)
+    job_queue.add.assert_called_once_with(job)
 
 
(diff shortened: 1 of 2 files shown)


hroncok:
This was needed in Fedora to update to Python 3.12 (together with using https://pypi.org/project/zombie-imp/ to workaround https://github.com/Ultimaker/Uranium/issues/765). Unfortunately, one of the asserts actually failed, hence I needed to comment it out. I don't think this ever asserted anything, so perhaps it was a wrong assert... ?

PR #890, opened 2023-07-26
[CURA-10722] Add 'additional settings providers' (Engine Plugins)

rburema:
Plugins should now be able to add settings. (But not append enum-values to existing ones, that's another ticket.)

- (De)serialization works, including load/save.
- Keys are tagged/'namespaced' with provider type (just a plugin for now), unique id, version.
- Will be sent to engine (except you have to demangle the tagged/namespaced name a bit).
- Setting-formulas work (except for taking values of another plugin).
- Can add to an existing category, or create a new one.

See (now) also the accompanying front-end PR here: https://github.com/Ultimaker/Cura/pull/16305

### DIFF
--- a/UM/Settings/AdditionalSettingDefinitionAppender.py
+++ b/UM/Settings/AdditionalSettingDefinitionAppender.py
@@ -11,69 +11,22 @@ from UM.PluginObject import PluginObject
 from UM.Settings.SettingDefinition import DefinitionPropertyType, SettingDefinition
 
 
-def prependIdToSettings(tag_type: str, tag_id: str, tag_version: str, settings: Dict[str, Any]) -> Dict[str, Any]:
-    """ This takes the whole (extra) settings-map as defined by the provider, and returns a tag-renamed version.
-
-    Additional (appended) settings will need to be prepended with (an) extra identifier(s)/namespaces to not collide.
-    This is done for when there are multiple additional settings appenders that might not know about each other.
-    This includes any formulas, which will also be included in the renaming process.
-
-    Appended settings may not be the same as 'baseline' (so any 'non-appended' settings) settings.
-    (But may of course clash between different providers and versions, that's the whole point of this function...)
-    Furthermore, it's assumed that formulas within the appended settings will only use settings either;
-     - as defined within the baseline, or;
-     - any other settings defined _by the provider itself_.
-
-    For each key that is renamed, this results in a mapping <key> -> _<provider_type>__<id*>__<version>__<key>
-     where '<id*>' is the version of the provider, but converted from using points to using underscores.
-    Example: 'tapdance_factor' might become '_plugin__dancingprinter__1_2_99__tapdance_factor'
-    Also note that all the tag_... parameters will be forced to lower-case.
-
-    :param tag_type: Type of the additional settings appender, for example; "PLUGIN".
-    :param tag_id: ID of the provider. Should be unique.
-    :param tag_version: Version of the provider. Points will be replaced by underscores.
-    :param settings: The settings as originally provided.
-
-    :returns: Remapped settings, where each settings-name is properly tagged/'namespaced'.
-    """
-    tag_type = tag_type.lower()
-    tag_id = tag_id.lower()
-    tag_version = tag_version.lower().replace(".", "_")
-
-    # First get the mapping, so that both the 'headings' and formula's can be renamed at the same time later.
-    def _getMapping(values: Dict[str, Any]) -> Dict[str, str]:
-        result = {}
-        for key, value in values.items():
-            mapped_key = key
-            if isinstance(value, dict):
-                if "type" in value and value["type"] != "category":
-                    mapped_key = f"_{tag_type}__{tag_id}__{tag_version}__{key}"
-                result.update(_getMapping(value))
-            result[key] = mapped_key
-        return result
-    key_map = _getMapping(settings)
-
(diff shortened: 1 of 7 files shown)


UM/Settings/DefinitionContainer.py
```diff
@@ -332,16 +332,52 @@ def deserialize(self, serialized: str, file_name: Optional[str] = None) -> str:
...
-            definition = SettingDefinition(key, self, None, self._i18n_catalog)
+        self._deserializeDefinitions(parsed["settings"])
+        return serialized
+
+    def _deserializeDefinitions(self, settings_dict: Dict[str, Any], force_category: Optional[str] = None) -> None:
+
+        # When there is a forced category (= parent) present, find the category parent, create it if it doesn't exist.
+        category_parent = None
+        if force_category:
+            category_parent = self.findDefinitions(key=force_category)
```
jellespijker, line 343:
```suggestion
            category_parent = self.findDefinitions(key = force_category)
```

UM/Settings/DefinitionContainer.py
```diff
@@ -332,16 +332,52 @@ def deserialize(self, serialized: str, file_name: Optional[str] = None) -> str:
...
             self._updateRelations(definition)
 
-        return serialized
+    def appendAdditionalSettingDefinitions(self, additional_settings: Dict[str, Dict[str, Any]]) -> None:
+        try:
+            merge_with_existing_categories = {}
+            create_new_categories = {}
+
+            for category, values in additional_settings.items():
+                if len(self.findDefinitions(key=category)) > 0:
```
jellespijker, line 365:
```suggestion
                if len(self.findDefinitions(key = category)) > 0:
```

jellespijker (changes requested):
LGTM; As we discussed during the stand-up we will add functionality to the plugin system as a whole (both front-end and curaengine_plugins) which will allow to add settings by simply specify a path to a `json` file. Which will have setting entries such as they are defined in `fdmprinter.def.json`

jellespijker:
Also pinging @fieldOfView because I think he also could have some valuable insights on this PR and at the very least FYI

fieldOfView:
> Also pinging @fieldOfView 

'sup?

Hmm, I don't quite get why this would be necessary. Many of my plugins are adding settings that eventually also get sent to CuraEngine (which happily ignores them), then to be used in postprocessing. It is all a bit hacky (private method access), but has been working very well for years.

rburema:
@fieldOfView -- If you don't see the need to change for your own plugins, that's fine. We're basically using the same mechanism anyway, so we won't clash.

Since (other) people are (we hope) going to be writing engine-plugins (also), there is additional need to make a proper (non-hacky) way to define additional settings for engine plugins. -- We may trust _you_ with our internals, but that's not necessarily the case for _every_ developer out there.

Also to us these settings do not feel quite the same as plugin preferences, and we'd like the possibility of defining values for them in profiles (for ex.; a 'weird printer' plugin that comes with an engine part and a bunch of profiles for example).

Additionally, different plugins might have the same setting-definition name added. We tag/'sort-of-namespace' the settings internally so that they at least don't destabilize the system. A similar story for different versions of the same plugin that might have tweaked their setting-definition meanings.

UM/Settings/AdditionalSettingDefinitionAppender.py
```diff
@@ -0,0 +1,108 @@
...
+                # Replace key-name 'heading'.
+                key = key_map.get(key, key)
+                value = _doReplace(value)
+            result[key] = value
+        return result
+    return _doReplace(settings)
+
+
+class AdditionalSettingDefinitionsAppender(PluginObject):
+
```
nallath, line 77: What is an additional settings defintions appender? When should one use it? (aka: I think that this needs documentation)

UM/Settings/DefinitionContainer.py
```diff
@@ -332,16 +332,52 @@ def deserialize(self, serialized: str, file_name: Optional[str] = None) -> str:
...
+                # Forced category; these are then children of that category, instead of full categories on their own.
+                category_parent.children.append(definition)
+            else:
+                self._definitions.append(definition)
 
         for definition in self._definitions:
             self._updateRelations(definition)
 
-        return serialized
+    def appendAdditionalSettingDefinitions(self, additional_settings: Dict[str, Dict[str, Any]]) -> None:
```
nallath, line 359: Needs a bit of documentation. What is this doing? In what cases should one use this?

UM/Settings/AdditionalSettingDefinitionAppender.py
```diff
@@ -0,0 +1,108 @@
...
+from pathlib import Path
+import os.path
+from typing import Any, Dict, List
+
+from UM.Logger import Logger
+from UM.PluginObject import PluginObject
+from UM.Settings.SettingDefinition import DefinitionPropertyType, SettingDefinition
+
+
+def prependIdToSettings(tag_type: str, tag_id: str, tag_version: str, settings: Dict[str, Any]) -> Dict[str, Any]:
```
nallath, line 14: Consider making this a static function of the class. My main argument for it is that right now, the documentation (that is nice and extensive!) for this function takes away the attention of the documentation of the class. But the class should be the "meat" of what is going on (and thus, should probably be listed first)

nallath (changes requested):
Apart from my other review comments; We need unit tests for this.

PR #900, opened 2023-09-01
[CURA-10870] If last we crashed, don't restore the previous window.

rburema:
Should work around certain multiple-monitor crashes.

_Update_: Now this also a little bit of frontend magic, see Ultimaker/Cura#16683

### DIFF
--- a/UM/Application.py
+++ b/UM/Application.py
@@ -38,6 +38,11 @@ if TYPE_CHECKING:
 
 @signalemitter
 class Application:
+    class ShutdownSeverity:
+        HardCrash = "hard_crash"
+        CrashHandler = "crash_handler"
+        NormalShutdown = "normal_shutdown"
+
     """Central object responsible for running the main event loop and creating other central objects.
 
     The Application object is a central object for accessing other important objects. It is also
@@ -207,6 +212,7 @@ class Application:
         self._preferences.addPreference("general/visible_settings", "")
         self._preferences.addPreference("general/plugins_to_remove", "")
         self._preferences.addPreference("general/disabled_plugins", "")
+        self._preferences.addPreference("general/last_shutdown_severity", Application.ShutdownSeverity.NormalShutdown)
 
         self._controller = Controller(self)
         self._output_device_manager = OutputDeviceManager()
--- a/UM/Qt/Bindings/MainWindow.py
+++ b/UM/Qt/Bindings/MainWindow.py
@@ -57,7 +57,6 @@ class MainWindow(QQuickWindow):
         self._preferences.addPreference("general/window_top", self.DEFAULT_WINDOW_TOP)
         self._preferences.addPreference("general/window_state", Qt.WindowState.WindowNoState.value)
         self._preferences.addPreference("general/restore_window_geometry", True)
-        self._preferences.addPreference("general/last_shutdown_normal", True)
 
         restored_geometry = QRect(int(self._preferences.getValue("general/window_left")),
                                   int(self._preferences.getValue("general/window_top")),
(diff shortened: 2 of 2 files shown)


fieldOfView:
This will move/resize my Cura window after _any_ crash, not just multi-display-related crashes.

nallath:
We could check if the crash happened without an exception being caught as well right? That would decrease the amount of times that this would falsely trigger.

rburema:
I implemented @nallath 's suggestion.

UM/Application.py
```diff
@@ -38,6 +38,11 @@
 
 @signalemitter
 class Application:
+    class ShutdownSeverity:
+        HardCrash = "hard_crash"
+        CrashHandler = "crash_handler"
```
wawanbreton, line 43: Wouldn't the description "handled crash" be a bit more readable ?

UM/Qt/Bindings/MainWindow.py
```diff
@@ -63,14 +63,22 @@ def __init__(self, parent = None):
...
             restored_geometry = QRect(self.DEFAULT_WINDOW_LEFT,
                                       self.DEFAULT_WINDOW_TOP,
                                       self.DEFAULT_WINDOW_WIDTH,
                                       self.DEFAULT_WINDOW_HEIGHT)
 
+        # Now set the 'last shutdown normal' to False, then, when the app closes normally, set it back to True.
+        self._preferences.setValue("general/last_shutdown_severity", Application.ShutdownSeverity.HardCrash)
+        Application.getInstance().applicationShuttingDown.connect(
+            lambda: self._preferences.setValue("general/last_shutdown_severity",
+                                               Application.ShutdownSeverity.NormalShutdown))
```
wawanbreton, line 80: From what I could test, this doesn't work:
1. Calling a lambda as a slot doesn't seem to work at all
2. Even if you put the content in a member function, the `applicationShuttingDown` signal is called after the `Application.savePreferences()` is processed, so the `NormalShutdown` state is never saved to file.

UM/Qt/Bindings/MainWindow.py
```diff
@@ -63,14 +63,22 @@ def __init__(self, parent = None):
...
+        if (not self._preferences.getValue("general/restore_window_geometry") or
+                self._preferences.getValue("general/last_shutdown_severity") == Application.ShutdownSeverity.HardCrash):
             # Ignore whatever the preferences said.
-            Logger.log("i", "Not restoring window geometry from preferences because 'restore_window_geometry' is false")
+            Logger.log("i", "Window geometry not restored from preferences. "
+                            "See 'general/restore_window_geometry' and/or 'general/last_shutdown_severity' for cause.")
             restored_geometry = QRect(self.DEFAULT_WINDOW_LEFT,
                                       self.DEFAULT_WINDOW_TOP,
                                       self.DEFAULT_WINDOW_WIDTH,
                                       self.DEFAULT_WINDOW_HEIGHT)
```
wawanbreton, line 74: This is out of current subject, but why not use `QWidget::saveGeometry()` and `QWidget::restoreGeometry()` ? The stored content is not human-readable, but it's designed especially for this purpose.

wawanbreton (changes requested):

UM/Application.py
rburema: This isn't used as a description, it's used as values in the config-file. I'm not sure that can handle a space.

The CrashHandler is a specific part of our code, although on the other hand I do mean 'handled crash' in a more general sense (even though that's the only way we can handle crashes).

wawanbreton: Oh I didn't mean to remove the space of course..

But as you way, "handled crash" is a more generic description, whenever there is class called `CrashHandler` or a `SometingBadIsHappeningHandler` or whatever..

UM/Qt/Bindings/MainWindow.py
rburema: I think we want it to be human-editable as a last resort for our users. (It might also be easier for QA.)

UM/Qt/Bindings/MainWindow.py
```diff
@@ -63,14 +63,22 @@ def __init__(self, parent = None):
...
 
+        # Now set the 'last shutdown normal' to False, then, when the app closes normally, set it back to True.
+        self._preferences.setValue("general/last_shutdown_severity", Application.ShutdownSeverity.HardCrash)
```
wawanbreton, line 77: This won't work if Cura crashes within 10 seconds, which is the delay for saving the preferences file. But maybe that's acceptable (but I would say no, because many things happen at start and you really don't want to go in a crash/restart loop..)

wawanbreton (changes requested):

UM/Qt/Bindings/MainWindow.py
wawanbreton: As long as you don't use a QMainWindow, that's good enough

UM/Qt/Bindings/MainWindow.py
rburema: Ah right, it works for `pyqtSignal`, not for `Signal`.

As for the second point, you're (again) correct. Instead of relying on signals, it may be better to directly call the main window, as it's a local variable there (in the class that sends the signal) anyway.

UM/Application.py
rburema: True, ok, I'll change it.

UM/Qt/Bindings/MainWindow.py
rburema: I'll have to rethink this entire feature. I promise I'm not usually this bad... 😅

wawanbreton: Maybe the preferences should be saved "when appropriate" instead of "regularly". That's what I usually do.

UM/Qt/Bindings/MainWindow.py
wawanbreton: What's the point of using a custom `Signal` mechanism ?

rburema:
This design isn't working at all in it's current form. Multiple problems exist with it. As the time-box has run out, I'm closing this.

PR #909, opened 2023-09-28
CURA-10967-extention of setpropertyvalue function

saumyaj3:
set property value updated to take into account if resolve is present or not.
To be reviewed with: https://github.com/Ultimaker/Cura/pull/16845


saumyaj3:
This PR is not needed

PR #910, opened 2023-09-29
Cura 11050 switching printers for first time

wawanbreton:
Applied some minor code modifications to speedup settings processing. This does not change the behavior at all but only processes things in a more optimized way.

Although it does not hurt, the result is better but still not perfect. Deeper changes would be required to keep decreasing the loading time, by processing less things.

### DIFF
--- a/UM/Qt/qml/UM/MenuItem.qml
+++ b/UM/Qt/qml/UM/MenuItem.qml
@@ -25,7 +25,7 @@ MenuItem
         enabled: root.action == null && root.enabled
         onActivated: root.triggered()
         context: Qt.ApplicationShortcut
-        sequence: root.action != null ? root.action.shortcut: null
+        sequences: [root.action != null ? root.action.shortcut: null]
     }
 
     // Workaround for menu items in controls not supporting mnemonic shortcuts in all cases.
--- a/UM/Settings/DefinitionContainer.py
+++ b/UM/Settings/DefinitionContainer.py
@@ -581,3 +581,6 @@ class DefinitionContainer(QObject, DefinitionContainerInterface, PluginObject):
 
     def __repr__(self) -> str:
         return str(self)
+
+    def __hash__(self) -> int:
+        return hash(self.getId())
--- a/UM/Settings/SettingDefinition.py
+++ b/UM/Settings/SettingDefinition.py
@@ -153,7 +153,7 @@ class SettingDefinition:
     def __hash__(self):
         """Ensure that the SettingDefinition is hashable, so it can be used in a set."""
 
-        return hash(str(self))
+        return hash((id(self), self._key, self._container))
 
     def __getstate__(self):
         """For Pickle support.
--- a/UM/Settings/SettingInstance.py
+++ b/UM/Settings/SettingInstance.py
@@ -173,14 +173,14 @@ class SettingInstance:
 
                     self.updateRelations(container, emit_signals = emit_signals)
 
-                if self._validator and emit_signals:
-                    self.propertyChanged.emit(self._definition.key, "validationState")
-
                 if emit_signals:
+                    if self._validator:
+                        self.propertyChanged.emit(self._definition.key, "validationState")
+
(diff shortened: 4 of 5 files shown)


UM/Settings/DefinitionContainer.py
```diff
@@ -333,16 +333,21 @@ def deserialize(self, serialized: str, file_name: Optional[str] = None) -> str:
         self._metadata["container_type"] = DefinitionContainer
 
         self._deserializeDefinitions(parsed["settings"])
+
+        for definition in self._definitions:
+            self._updateRelations(definition)
```
casperlamboo, line 338: Are we here also just interested in the newly added definitions?
```suggestion
        added_definitions = self._deserializeDefinitions(parsed["settings"])

        for definition in added_definitions:
            self._updateRelations(definition)
```

UM/Settings/SettingDefinition.py
```diff
@@ -153,7 +153,7 @@ def __setattr__(self, name: str, value: Any) -> None:
     def __hash__(self):
         """Ensure that the SettingDefinition is hashable, so it can be used in a set."""
 
-        return hash(str(self))
+        return hash((id(self), self._key, self._container))
```
jellespijker, line 156: switching from `str` to `id` can have unforeseen consequence, it will probably be okay in this case, but remember that `id()` is the memory address for the CPython implementation. Speaking of personal experience best to double check how this behaves, when instance are assigned `=`, `copied` or `deepcopied`.

UM/Qt/qml/UM/MenuItem.qml
```diff
@@ -25,7 +25,7 @@ MenuItem
         enabled: root.action == null && root.enabled
         onActivated: root.triggered()
         context: Qt.ApplicationShortcut
-        sequence: root.action != null ? root.action.shortcut: null
+        sequences: [root.action != null ? root.action.shortcut: null]
```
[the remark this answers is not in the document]
wawanbreton: @saumyaj3 Can you please take a look at this one ?

UM/Settings/SettingDefinition.py
wawanbreton: I changed all the `__hash__` methods which had a `__repr__` method present, and took the very same used variables, so the result should be stricly similar, without the cost of computing the string.

UM/Settings/DefinitionContainer.py
wawanbreton: Well I've been thinking about that indeed. The result should be similar, but I preferred to keep the old code to avoid breaking something. Could you test this suggestion during component test ?

UM/Settings/SettingDefinition.py
jellespijker: `id()` https://docs.python.org/3/library/functions.html#id is not the same as the dunder `__repr__` https://docs.python.org/3/library/functions.html#repr

jellespijker: Btw, I think it should work as expected in this case, but it is my experience that using `id()` can have unwanted side effects because, certainly if assignment `=` (copy of reference) and/or `deepcopy` come into play, not sure if that happens here. I'm just warning that this should be a conscious choice.

wawanbreton: In general, I agree that using `id()` for the hash is not a good idea, but it depends on whether you want to compare for two objects being the same instance or having the same content.

In this very specific case, the hash was computed on `str(self)`, which actually calls `__repr__` to create the string, then computes the hash of the string. The `__repr__` method of `SettingDefinition` uses fixed characters and the variables `id(self)`, `self._key` and `self._container`, so calling them directly in the hash method is just a shortcut. The computed number will be different, but will evolve the same way, which is what we want for a hash, and gives a similar result to the original code.

Maybe we could still discuss this, because having the ID involved for the hash may be bad indeed. It is useful for debug purposes in the `__repr__` method but may not make sense for the hash. Splitting the string representation and the hash seems to be a good thing to me, because they have different purposes.

PR #916, opened 2023-10-20
Makerbot iso view renderer

casperlamboo:


UM/Qt/QtRenderer.py
```diff
@@ -210,10 +210,10 @@ def _initialize(self) -> None:
         supports_vao = OpenGLContext.supportsVertexArrayObjects()  # fill the OpenGLContext.properties
         Logger.log("d", "Support for Vertex Array Objects: %s", supports_vao)
 
-        OpenGL()
```
fieldOfView, line 213: Moving the initialisation of the OpenGL singleton *may* have an influence on compatibility with specific gpu drivers, eg in combination with multi-monitor setups. It could be that this change actually fixes some of the "Cura does not work with multiple displays" issues, or it could make them worse. What I am saying is that this probably has to be tested on many different systems.

jellespijker: good point, maybe we should create see if we can finish setting up the alpha nightlies release, such that we have a longer test program for this

rburema: Best keep an eye out indeed, although it _should_ not matter if the initialization is early enough and happens in the right thread ... and nothing messes with the state in-between.

I think I'd prefer it if `OpenGL` was a 'normal' singleton instead of something you _first_ have to explicitly initialize. (Which we could make, because it's under our control; an Uranium class.)

Because I just spend the better part of an hour figuring out if this can work in any case (at least from looking at the code), and yes it does, technically.

The only 'reason' this works at all is that `beginRendering` is the only method that calls the `_initialize()` private method in `QtRenderer`, which is fortunately (except for in `reRenderLast`) called by ... `MainWindow` itself.

Otherwise, we'd have the following situation, since `QtRenderer` might be initialized earlier than the `MainWindow`.

In `CuraApplication.runWithGUI()`, we first get:

~~~
        self.getRenderer().setBackgroundColor(QColor(245, 245, 245))
~~~

which gets the renderer, which initializes it. Then a bunch of lines later in the same function (so _afterwards_), we:

~~~
        # Initialize QML engine
        self.setMainQml(Resources.getPath(self.ResourceTypes.QmlFiles, "Cura.qml"))
~~~

(... but that doesn't matter because the first render doesn't happen there, so the internal init method of the renderer doesn't get called, and the main-window is the first to get to OpenGL initialization after all.)

rburema (approved):
See my comment in the discussion with fieldofview and jelle. TLDR; I'd prefer it if the OpenGL singleton was something that did the explicit initialization itself, so you'd only have to 'getOpenGL', and not worry about what/where the first init is.

However, that the way it's done now was already the case before, so up to you if you want to do that as part of this ticket.

PR #922, opened 2023-10-29
MakerBot plugin

jellespijker:
CURA-10561

### DIFF
--- a/UM/Application.py
+++ b/UM/Application.py
@@ -185,10 +185,11 @@ class Application:
         Resources.addSecureSearchPath(os.path.join(self._app_install_dir, "Resources", self._app_name, "resources"))
 
         if not hasattr(sys, "frozen"):
-            Resources.addSearchPath(str(Path(sys.executable).parent.joinpath("share", "cura", "resources")))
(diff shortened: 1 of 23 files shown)


UM/Application.py
```diff
@@ -185,6 +185,9 @@ def initialize(self) -> None:
         Resources.addSecureSearchPath(os.path.join(self._app_install_dir, "Resources", self._app_name, "resources"))
 
         if not hasattr(sys, "frozen"):
+            uranium_data_root = os.environ.get('URANIUM_DATA_ROOT', None)
```
wawanbreton, line 188: I don't see a place where this variable is set. Is this for development environment only ? CURA_DATA_ROOT is set by conanfile.py though.

wawanbreton (commented):
Just one minor remark/question. The rest is quite harmless :)

UM/Application.py
jellespijker: It would allow you to use this when fooling around on a personal environment.  maybe point towards an user specific path.

The CURA_DATA_ROOT is already filled in and taken when you do a Conan install.

wawanbreton: Ok, just wanted to make sure it was purposely not added to conanfile.py.

PR #933, opened 2024-01-09
CURA-7812-decrease-Qml-singletons

saumyaj3:
CURA-7812
Delete and refactor a number of singletons in QML to clean up the code. Document the refactor decisions made in the SDK documentation. 
Deleted: (Deprecated)
QualityManagementModel
MaterialManagementModel
QualityProfilesDropDownMenuModel 
SimpleModeSettingsManager
MachineActionManager
UM.Application 
CustomQualityProfilesDropDownMenuModel

Merged: (also deprecated)
WorkspaceFileHandler (merged with meshfile )
OperationStack (merge with actions)
ActiveTool (merge with Controller)

### DIFF
--- a/UM/Qt/Bindings/OperationStackProxy.py
+++ b/UM/Qt/Bindings/OperationStackProxy.py
@@ -4,8 +4,10 @@
 from PyQt6.QtCore import QObject, pyqtSlot, pyqtProperty, pyqtSignal
 
 from UM.Application import Application
+from UM.Decorators import deprecated
 
 
+@deprecated("OperationStackProxy is deprecated and will be removed in major SDK release")
 class OperationStackProxy(QObject):
     def __init__(self, parent = None):
         super().__init__(parent)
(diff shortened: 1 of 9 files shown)


fieldOfView:
Doesn't removing these singletons constitute a rather big API break?

saumyaj3:
> Doesn't removing these singletons constitute a rather big API break?

Hey @fieldOfView Thanks for pointing it out. The work is in progress to fix it. Converted the PR to draft

UM/Qt/Bindings/OperationStackProxy.py
```diff
@@ -6,6 +6,7 @@
 from UM.Application import Application
 
 
+#"OperationStackProxy is deprecated and will be removed in major SDK release"
```
wawanbreton, line 9: Could you add a `deprecated` decorator on the constructor method ? This comment is not very visible...

wawanbreton (commented):
Just a single suggestion to make the deprecations more visible

UM/Qt/Bindings/OperationStackProxy.py
saumyaj3: Yes I tried, It doesn't work here somehow. Cura start crashing when this class is called then. I provided the deprecated only for the wrapper, cause it shows there. Here, I added a message so that we remove it later by ourself.

wawanbreton: Ok, constructors are not normal methods indeed, so that's understandable. As long as warnings show up at some point, that's fine.

PR #936, opened 2024-01-31
Fix mac builds

casperlamboo:
See Cura PR: https://github.com/Ultimaker/Cura/pull/18188

CURA-11483


wawanbreton (approved):
Looks good, no risk or regression in this part

PR #938, opened 2024-02-05
Cura 11561 mockup pap

wawanbreton:
Small factorization to ease Printer Agnostic Project mockup.

Comes with https://github.com/Ultimaker/Cura/pull/18246

CURA-11561


UM/Qt/qml/UM/HelpIcon.qml
```diff
@@ -0,0 +1,39 @@
+// Copyright (c) 2024 Ultimaker B.V.
+// Uranium is released under the terms of the LGPLv3 or higher.
+
+import QtQuick 2.7
+import QtQuick.Controls 2.3
+
+import UM 1.8 as UM
+
+MouseArea
```
nallath, line 9: Why not use a UM.TooltipArea?

PR #939, opened 2024-02-20
Cura 11617 print manually

saumyaj3:
Duplicating a model (copy paste or multiply) breaks the ordering logic (print before/after) of UI as all the models generated from the original instance will take the same order index. As a result, you cannot print before/after if you use the multiply function. I added a simple solution to organize the models in the print sequence when copy paste or multiply operation is called when user defined sequence is switched on.

follow up of: https://github.com/Ultimaker/Cura/pull/17893

### DIFF
--- a/UM/Scene/Iterator/DepthFirstIterator.py
+++ b/UM/Scene/Iterator/DepthFirstIterator.py
@@ -10,7 +10,7 @@ class DepthFirstIterator(Iterator.Iterator):
 
     def _fillStack(self) -> None:
         self._addNodesToStack()
-        self._set_print_order_for_sliceable_nodes()
+        self._setPrintOrderForSliceableNodes()
 
     def _addNodesToStack(self) -> None:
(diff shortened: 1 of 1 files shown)


UM/Scene/Iterator/DepthFirstIterator.py
```diff
@@ -9,5 +9,16 @@ def __init__(self, scene_node: SceneNode) -> None:
         super().__init__(scene_node)
 
     def _fillStack(self) -> None:
+        self._addNodesToStack()
+        self._set_print_order_for_sliceable_nodes()
```
nallath, line 13:
```suggestion
        self._setPrintOrderForSliceableNodes()
```

nallath (changes requested):

saumyaj3:
This doesn't work . I am closing the PR. Solution made in Cura 👍

PR #940, opened 2024-02-21
Protobuf version mismatch caused tests,builds to fail.

rburema:


rburema:
Superseded by https://github.com/Ultimaker/Uranium/pull/941

PR #942, opened 2024-02-28
[CURA-10149] make 'Lay Flat By Face' work for groups & merges

rburema:
It wouldn't even let you select a face if the selected node was a group or a merged mesh.

Should fix https://github.com/Ultimaker/Cura/issues/14237

### DIFF
--- a/UM/View/SelectionPass.py
+++ b/UM/View/SelectionPass.py
@@ -188,8 +188,6 @@ class SelectionPass(RenderPass):
         """Get an unique identifier to any object currently selected for by-face manipulation at a pixel coordinate."""
         output = self.getOutput()
 
-        output.save("C:/tmp_/faceid.png")
-
         window_size = self._renderer.getWindowSize()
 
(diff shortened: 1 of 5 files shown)


UM/View/SelectionPass.py
```diff
@@ -152,6 +184,31 @@ def getIdAtPosition(self, x, y):
         pixel = output.pixel(px, py)
         return self._selection_map.get(Color.fromARGB(pixel), None)
 
+    def getIdAtPositionFaceMode(self, x, y):
+        """Get an unique identifier to any object currently selected for by-face manipulation at a pixel coordinate."""
+        output = self.getOutput()
+
+        output.save("C:/tmp_/faceid.png")
```
wawanbreton, line 191: Please remove this

wawanbreton (changes requested):
One small remark, a manifest debug instruction is still present

UM/View/SelectionPass.py
rburema: Done!

PR #948, opened 2024-03-18
Cura 11475 low performance multiple nozzles

wawanbreton:
Instead of getting all the settings and checking for their state, run only once through all the containers and gather only the keys having a User state.

Comes with https://github.com/Ultimaker/Cura/pull/18592

CURA-11475


UM/Settings/InstanceContainer.py
```diff
@@ -796,6 +797,17 @@ def sendPostponedEmits(self) -> None:
         while self._postponed_emits:
             signal, signal_arg = self._postponed_emits.pop(0)
             signal.emit(*signal_arg)
+            
+    def getAllKeysWithUserState(self):
```
casperlamboo, line 801:
```suggestion
    def getAllKeysWithUserState(self)-> Set[str]:
```

casperlamboo (commented):
small comment

casperlamboo (approved):

PR #949, opened 2024-04-03
Implement printer search feature in local printer selection

saumyaj3:
this condition fails and creates a crash when changed_row_count ==0

CURA-11003


wawanbreton (approved):
That does make sense :smiley:

nallath (approved):

PR #950, opened 2024-04-12
clearing memory stack

saumyaj3:
CURA-11658

### DIFF
--- a/UM/Operations/OperationStack.py
+++ b/UM/Operations/OperationStack.py
@@ -1,6 +1,6 @@
 # Copyright (c) 2019 Ultimaker B.V.
 # Uranium is released under the terms of the LGPLv3 or higher.
-
+import sys
 import threading
 import time
 
@@ -33,6 +33,13 @@ class OperationStack():
 
     def _onToolOperationStarted(self, tool):
         self._merge_operations = False
+
+    def controlStackSize(self):
+        if sys.getsizeof(self._operations)> 5000:
+            print("getting stack size", sys.getsizeof(self._operations), len(self._operations))
+            self._current_index -=1
+            self._operations.pop(0)
+            self.controlStackSize()
 
     def _onToolOperationStopped(self, tool):
         self._merge_operations = False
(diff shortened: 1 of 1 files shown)


UM/Operations/AddSceneNodeOperation.py
```diff
@@ -46,3 +46,8 @@ def redo(self) -> None:
         self._node.setParent(self._parent)
         if self._selected:  # It was selected while the operation was undone. We should restore that selection.
             Selection.add(self._node)
+
+    def delete(self) -> None:
```
wawanbreton, line 50: Maybe rename to be consistent with the other clearing methods ? Same for others
But this method may have a slightly different purpose, in which case this is fine.

UM/Operations/Operation.py
```diff
@@ -57,3 +57,11 @@ def push(self) -> None:
         # Because of circular dependency
         from UM.Application import Application
         Application.getInstance().getOperationStack().push(self)
+
+    def clear(self) -> None:
+        # Because of circular dependency
+        from UM.Application import Application
+        Application.getInstance().getOperationStack().clearStack()
```
wawanbreton, line 64: This is a bit weird that the operation object clears the stack, does that mean it will be done multiple times ? Wouldn't it be the job of the caller then ? Or maybe the operation should only remove itself instead.

wawanbreton (commented):
Same remark as for the other PR, I'm afraid of crashes, and I think methods naming is inconsistent.

UM/Operations/Operation.py
saumyaj3: I don't think it'll be cleared multiple times, its only when an object of operation asks to clear the stack. I took the example from the function push just below this function. The push in operation stack also happens from one of the operation.

UM/Operations/AddSceneNodeOperation.py
saumyaj3: I am on it. making it more clear to what clear-> clearingData or delete-> deletingObject mean

wawanbreton (approved):
Same remark as for https://github.com/Ultimaker/Cura/pull/18883, still some reserve about this but I let you move on.

PR #952, opened 2024-06-11
CURA-11939 upgrade to python 3.12

wawanbreton:
Make Uranium use Python 3.12, which also requires updating some dependencies.

CURA-11939


wawanbreton:
Implemented in https://github.com/Ultimaker/Uranium/pull/981

PR #956, opened 2024-07-11
CURA-11984 Fix some models not supporting "align to buildplate"

wawanbreton:
Instead of using a dynamic number of bits for the face index, store the object ID in the alpha channel, which was previously unused. This way we have the full 24 bits of the RGB channels to store the face index.

OpenGL blending is disabled for this pass, so the alpha channel is just stored as an extra channel but no calculation is performed based on it.

CURA-11984

### DIFF
--- a/UM/View/SelectionPass.py
+++ b/UM/View/SelectionPass.py
@@ -65,7 +65,6 @@ class SelectionPass(RenderPass):
 
         self._mode = SelectionPass.SelectionMode.OBJECTS
         Selection.selectedFaceChanged.connect(self._onSelectedFaceChanged)
-        self._face_mode_max_objects = 1  # Needed when selecting a face for (a) grouped or merged object(s).
 
         self._output = None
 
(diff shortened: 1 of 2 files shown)


rburema (approved):
Looks good! I don't fully remember why I initially did the alpha that way ... it might've been part of another shader first.

The only thing I do notice is that
`if current_model_id >= 128:`
can probably be replaced with
`if current_model_id >= 255:`
now.

Now I just need to test -- let's see if the grouped object stuff also still works.

wawanbreton:
> Looks good! I don't fully remember why I initially did the alpha that way ... it might've been part of another shader first.

Maybe because the alpha can be modified during the OpenGL pipeline processing, so it is a bit more sensitive.

> The only thing I do notice is that `if current_model_id >= 128:` can probably be replaced with `if current_model_id >= 255:` now.

Good catch, that is true :) I'll make a commit

PR #962, opened 2024-08-15
[CURA-12067] The json-decorator-scope didn't set the user-agent yet.

rburema:
Add the header-dictionaries from base (which should've got the User-Agent set, at the very least in case we wrap around something that inherits from DefaultUserAgentScope, which is the only use-case I encountered) with the json-specific settings that are already in there. In order to prevent breaking the API, don't directly change the inheritance of the JsonDecoratorScope, but instead check for the existence of the header-dict in the base.

### DIFF
--- a/UM/TaskManagement/HttpRequestScope.py
+++ b/UM/TaskManagement/HttpRequestScope.py
@@ -51,6 +51,10 @@ class JsonDecoratorScope(HttpRequestScope):
             "Content-Type": "application/json",
             "Accept": "application/json"
         }
+        # JsonDecorator probably _should_ have inherited from DefaultUserAgentScope in the first place.
+        # In order to prevent a break in the API by forcing that, just check if there are extra header specs needed.
+        if hasattr(self.base, "header_dict"):
+            self.header_dict.update(self.base.header_dict)
 
     def requestHook(self, request: QNetworkRequest) -> None:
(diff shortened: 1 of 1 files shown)


UM/TaskManagement/HttpRequestScope.py
```diff
@@ -51,6 +51,10 @@ def __init__(self, base: HttpRequestScope) -> None:
             "Content-Type": "application/json",
             "Accept": "application/json"
         }
+        # JsonDecorator probably _should_ have inherited from DefaultUserAgentScope in the first place.
+        # In order to prevent a break in the API by forcing that, just check if there are extra header specs needed.
```
wawanbreton, line 55: I'm not sure it should actually inherit from `DefaultUserAgentScope`. You may want to use this decorator without the user-agent, or with a different one. Those decorators are, as I understood them, designed to be chained.
I would suggest to either:
* Commit to your suggestion and force the inheritance
* Commit to this new behavior and don't mention the inheritence in the comment

On the other hand, I'm overall surprised by this fix. The `header_dict` values should be applied in the `requestHook` method, both the ones from the base and the instance, so I don't see what good this does. I'm probably missing something.

rburema: Ah, I missed that, sorry.
I think we should close this PR -- sorry 😓

PR #966, opened 2024-09-03
[CURA-11761] Cache results of class-member methods

rburema:
Be able to cache results from member-functions of classes (as well as just plain using `lru_cache` for class- and static- methods) and use that to cache a lot of the results from the Container/Settings stacks.

Depending on the situation (such as with a printer with multiple extruders in Cura for example) -- this can speed up the interface quite noticibly.

(There's also a small PR in the front-end to make extra sure that objects inheriting from the cached classes play fair. -- )

### DIFF
--- a/UM/Settings/ContainerStack.py
+++ b/UM/Settings/ContainerStack.py
@@ -409,7 +409,6 @@ class ContainerStack(QObject, ContainerInterface, PluginObject):
             Logger.log("d", "Could not get version from serialized: %s", e)
         return version
 
-    @cachePerInstance
     def deserialize(self, serialized: str, file_name: Optional[str] = None) -> str:
         """:copydoc ContainerInterface::deserialize
 
@@ -417,6 +416,8 @@ class ContainerStack(QObject, ContainerInterface, PluginObject):
 
         TODO: Expand documentation here, include the fact that this should _not_ include all containers
         """
+
+        CachedMemberFunctions.clearInstanceCache(self)
 
         # Update the serialized data first
         serialized = super().deserialize(serialized, file_name)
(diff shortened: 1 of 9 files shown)


UM/Settings/ContainerStack.py
```diff
@@ -605,6 +634,7 @@ def getSettingDefinition(self, key: str) -> Optional[SettingDefinition]:
             return None
 
     @UM.FlameProfiler.profile
+    #TODO: Find a way around keyword-arguments being passed as a dict, or live with not cacheing this particular func.
```
wawanbreton, line 637: Now the question is, how often is this method called in normal use conditions ?

UM/Settings/ContainerStack.py
```diff
@@ -832,6 +871,11 @@ def __str__(self) -> str:
     def __repr__(self) -> str:
         return str(self)
 
+    def __del__(self) -> None:
+        # None of the 'parents' seem to have __del__, so OK not to call `super.del` here.
```
wawanbreton, line 875: Should we not still do it, in case a parent someday gets a `__del__` ? In the meantime it should not hurt (applies multiple times)

wawanbreton (commented):
Some minor remarks/questions.
I am also a bit surprised that there is no "selective" clearing of the cache. I would have expected the container to clear only the relevant settings values when a source setting changes. This may not be necessary though.

UM/Settings/ContainerStack.py
rburema: There's 100+ matches on 'findContainer' and they bascially never call without arguments. Each of them would have to be converted to using criteria instead of the kwargs as much as possible, and that would have to be a frozendict or somesuch (which is a separate library, as opposed to frozenset), which would also break the API, unless we make it a separate method and don't use this one. 🫠 

Seeing as the rest of the caching already does its job, I don't think its worth it 😅

UM/Settings/ContainerStack.py
rburema: I tried that, but then it complains that `super` doesn't have a method `__del__`.

See also https://stackoverflow.com/questions/36722390/python-3-super-del

(... but there is a bit of a workaround if we want that, so I'll just commit that.)

wawanbreton: Oh, that sucks... You can keep it as is then.

wawanbreton (approved):

wawanbreton (approved):

PR #967, opened 2024-09-05
Minor optimization

wawanbreton:
Introduced while working on CURA-12093

It is just a slight optimization, and absolutely not necessary.

### DIFF
--- a/UM/Settings/SettingFunction.py
+++ b/UM/Settings/SettingFunction.py
@@ -104,8 +104,7 @@ class SettingFunction:
             local_variables[name] = value
 
         if additional_variables is not None:
-            for name, value in additional_variables.items():
-                local_variables[name] = value
+            local_variables = local_variables | additional_variables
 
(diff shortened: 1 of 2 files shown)


UM/Settings/SettingFunction.py
```diff
@@ -104,8 +104,7 @@ def __call__(self, value_provider: ContainerInterface, context: Optional[Propert
             local_variables[name] = value
 
         if additional_variables is not None:
-            for name, value in additional_variables.items():
-                local_variables[name] = value
+            local_variables = local_variables | additional_variables
```
rburema, line 107: Or even:

```suggestion
            local_variables |= additional_variables
```

rburema (approved):
Nice catch. I'll go review the main thing now.

UM/Settings/SettingFunction.py
wawanbreton: Even better indeed.. I'll give it a try

PR #970, opened 2024-09-27
CURA-12138 export for support option

wawanbreton:
CURA-12138

* Allow saving a project file without setting the name as the project name
* Allow giving extra arguments to a file writer
* Allow FileLogger to flush the current file and returns the list of used files paths

### DIFF
--- a/UM/FileHandler/WriteFileJob.py
+++ b/UM/FileHandler/WriteFileJob.py
@@ -10,7 +10,7 @@ from UM.Message import Message
 import io
 import time
 
-from typing import Any, Optional, Union
+from typing import Any, Optional, Union, Dict
 
 
 class WriteFileJob(Job):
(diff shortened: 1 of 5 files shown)


UM/FileHandler/WriteFileJob.py
```diff
@@ -19,7 +19,7 @@ class WriteFileJob(Job):
     The writer defines what the result of this job is.
     """
 
-    def __init__(self, writer: Optional[FileWriter], stream: Union[io.BytesIO, io.StringIO], data: Any, mode: int) -> None:
+    def __init__(self, writer: Optional[FileWriter], stream: Union[io.BytesIO, io.StringIO], data: Any, mode: int, writer_args: dict = {}) -> None:
```
nallath, line 22: You shouldn't use a dict like this! It's a bit of a weird thing that python does. So basically what happens here is that it generates a _single_ default dict that is static. So every time the default value is used, it references the same dict. 

Instead, you should do something like:

```suggestion
    def __init__(self, writer: Optional[FileWriter], stream: Union[io.BytesIO, io.StringIO], data: Any, mode: int, writer_args: Optional[Dict] ) -> None:
        if writer_args is None:
            writer_args = {}
```
Also note that i've changed the type of the writer_args to Dict with a capital letter, as that's the one used by typing.

nallath (changes requested):

UM/FileHandler/WriteFileJob.py
wawanbreton: > So basically what happens here is that it generates a _single_ default dict that is static. So every time the default value is used, it references the same dict.

Waw, that is something I didn't expect indeed. I'll try to remember.

PR #974, opened 2024-10-15
[CURA-12190] Expand the circumstances under which we expect the C++ object to be gone.

rburema:
This was uncovered after caching, but might've existed before; that extra update function that's now within the try could _also_ call the emitters, causing the same problem for which the existing try-catches where made.


wawanbreton (approved):
A bit scary, but ok :grimacing:

rburema:
> A bit scary, but ok 😬

I think that's mostly a result from coupling a garbage collected language so tightly with a self-managed one, and then try to have objects that correspond to each other on both sides of that barrier managed by yet another system that has been known to have some bugs -- and then I come along and try to do caching 😓

PR #978, opened 2024-11-25
Rotation by input & snap angle input

HellAholic:
Add rotation by input fields
- Fields only active when snap rotation is disabled
- Fields take user input, rotate the model and return to value 0
- Getting model rotation or storing the rotation values lead to unexpected behavior (read leave Quaternions alone)
- Includes the snap rotation PR (https://github.com/Ultimaker/Uranium/pull/979)
- Adds setter and getter for RX,RY,RZ (rotations in each axis)
- Applies the rotations for single or multiple model(s)

Rotate by input
rotation_by_input1

Snap angle update
Based on the idea of GregValiant
snap_angle

### DIFF
--- a/plugins/Tools/RotateTool/RotateTool.qml
+++ b/plugins/Tools/RotateTool/RotateTool.qml
@@ -125,12 +125,12 @@ Item
             width: Math.ceil(contentWidth) //Make sure that the grid cells have an integer width.
         }
 
-		UM.TextFieldWithUnit
+        UM.TextFieldWithUnit
         {
             id: angleTextField
             width: UM.Theme.getSize("setting_control").width
             height: UM.Theme.getSize("setting_control").height
-            unit: "degrees"
+            unit: "°"
             text: snapText
 
             validator: UM.FloatValidator
@@ -141,14 +141,14 @@ Item
             onEditingFinished:
             {
                 var modified_text = text.replace(",", ".") // User convenience. We use dots for decimal values
-                if(text !="")
+                if(text != "")
                 {
                     UM.Controller.setProperty("RotationSnapAngle", modified_text)
                 }
             }
             onActiveFocusChanged:
             {
-                if(!activeFocus && text =="")
+                if(!activeFocus && text == "")
                 {
                     snapText = 0.1; // Need to change it to something else so we can force it to getvalue
                     snapText = UM.Controller.properties.getValue("RotationSnapAngle")
@@ -206,12 +206,12 @@ Item
             width: Math.ceil(contentWidth) //Make sure that the grid cells have an integer width.
         }
 
-		UM.TextFieldWithUnit
+        UM.TextFieldWithUnit
         {
             id: xangleTextField
             width: UM.Theme.getSize("setting_control").width
             height: UM.Theme.getSize("setting_control").height
-            unit: "degrees"
+            unit: "°"
             text: xText
 
             validator: UM.FloatValidator
@@ -222,15 +222,15 @@ Item
             onEditingFinished:
             {
                 var modified_text = text.replace(",", ".") // User convenience. We use dots for decimal values
-                if(text !="")
+                if(text != "")
                 {
-                    UM.Controller.setProperty("RX", modified_text)
+                    UM.Controller.setProperty("RotationX", modified_text)
                     text = "0"
                 }
             }
             onActiveFocusChanged:
             {
-                if(!activeFocus && text =="")
+                if(!activeFocus && text == "")
                 {
                     xText = 0.1; // Need to change it to something else so we can force it to getvalue
                     xText = 0
@@ -243,7 +243,7 @@ Item
             id: yangleTextField
             width: UM.Theme.getSize("setting_control").width
             height: UM.Theme.getSize("setting_control").height
-            unit: "degrees"
+            unit: "°"
             text: yText
 
             validator: UM.FloatValidator
@@ -254,16 +254,16 @@ Item
             onEditingFinished:
             {
                 var modified_text = text.replace(",", ".") // User convenience. We use dots for decimal values
-                if(text !="")
+                if(text != "")
                 {
                     // Yes this is intentional. Y & Z are flipped between model axes and build plate axes
-                    UM.Controller.setProperty("RZ", modified_text)
+                    UM.Controller.setProperty("RotationZ", modified_text)
                     text = "0"
                 }
             }
             onActiveFocusChanged:
             {
-                if(!activeFocus && text =="")
+                if(!activeFocus && text == "")
                 {
                     yText = 0.1; // Need to change it to something else so we can force it to getvalue
                     // Yes this is intentional. Y & Z are flipped between model axes and build plate axes
@@ -277,7 +277,7 @@ Item
             id: zangleTextField
             width: UM.Theme.getSize("setting_control").width
             height: UM.Theme.getSize("setting_control").height
-            unit: "degrees"
+            unit: "°"
             text: zText
 
             validator: UM.FloatValidator
@@ -288,16 +288,16 @@ Item
             onEditingFinished:
             {
                 var modified_text = text.replace(",", ".") // User convenience. We use dots for decimal values
-                if(text !="")
+                if(text != "")
                 {
                     // Yes this is intentional. Y & Z are flipped between model axes and build plate axes
-                    UM.Controller.setProperty("RY", modified_text)
+                    UM.Controller.setProperty("RotationY", modified_text)
                     text = "0"
                 }
             }
             onActiveFocusChanged:
             {
-                if(!activeFocus && text =="")
+                if(!activeFocus && text == "")
                 {
                     zText = 0.1; // Need to change it to something else so we can force it to getvalue
                     // Yes this is intentional. Y & Z are flipped between model axes and build plate axes
@@ -333,20 +333,20 @@ Item
     {
         target: base
         property: "xText"
-        value: base.roundFloat(UM.Controller.properties.getValue("RX"), 2)
+        value: base.roundFloat(UM.Controller.properties.getValue("RotationX"), 2)
     }
 
     Binding
     {
         target: base
         property: "zText"
-        value: base.roundFloat(UM.Controller.properties.getValue("RY"), 2)
+        value: base.roundFloat(UM.Controller.properties.getValue("RotationY"), 2)
     }
 
     Binding
     {
         target: base
         property: "yText"
-        value: base.roundFloat(UM.Controller.properties.getValue("RZ"), 2)
+        value: base.roundFloat(UM.Controller.properties.getValue("RotationZ"), 2)
     }
 }
--- a/plugins/Tools/RotateTool/RotateTool.qml
+++ b/plugins/Tools/RotateTool/RotateTool.qml
@@ -150,12 +150,13 @@ Item
             {
                 if(!activeFocus && text =="")
                 {
-                    snapText = 0.1; // Yeaaah i know. We need to change it to something else so we can force it to 0
+                    snapText = 0.1; // Need to change it to something else so we can force it to getvalue
                     snapText = UM.Controller.properties.getValue("RotationSnapAngle")
                 }
             }
         }
     }
+
     UM.CheckBox
     {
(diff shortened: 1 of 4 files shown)


GregValiant:
@nallath Any thoughts on this one?  It sort of goes hand-in-hand with the "Y" axis Translate tool.

GregValiant:
It looks really good.  It will rotate to any custom angle to two decimal places.

Here I've adjusted the angle to 11.25 so it will hit 22.5 right on the money.
image

rburema:
I know we had something like this previously, but purposefully took them out because they'd mess up scaling, mirroring, etc.

For example:

https://github.com/user-attachments/assets/8d17cea5-a4b4-418d-8dc6-fa1c9534638c

HellAholic:
Thanks for the check and feedback. Did not check that part. I'll try to see if I can figure out what goes wrong where and either:
- Take out rotation by input and just leave the snap rotation in (if that doesn't cause the same issue)
- Fix the scaling??? (wishful thinking but a magical time is on approach)

GregValiant:
That looks like how it always acted because the bounding box doesn't rotate but rather stretches to fit the model.  Any rotation must be the last transformation.
If you rotate the model first and then stretch it, it gets skewed because that's how it is in the bounding box.
This is in 4.13.1
https://github.com/user-attachments/assets/71112f15-5efb-4b56-9fb1-60ac625b05d7

GregValiant:
I think the only way around the weirdness that occurs when scaling a rotated body is to rotate the bounding box and have the part coordinate system tied to it.  When you rotate the model, the local coordinate system rotates with it.  I don't see that happening.
I have Cura 2.3 loaded on a desktop and the behavior is the same.  Rotation must always be last.

rburema:
You're right, it does already happen -- woke up way to early with the thought to check this, but you'd beaten me to it. -- That'll teach me to review stuff when I'm tired.

Anyway, I think this works because the rotation fields get set back to `0` here, we previously didn't have that (_also_ tried to keep the rotation OK), and I think _that_ may have caused the actual issue.

So the concept looks good (though I haven't looked at the code properly yet) -- I've only taken off a few days this year, so now I'm really going into my month-long holiday == see you on the other side! 👋🏻

GregValiant:
I'm sure this will still be sitting here when you get back.

I'd been using my bare bones version of the Rotation change (along with the Translation Tool "Y" flip) for a while, and I haven't seen any adverse behavior.  I have the latest versions of both installed and it's all good.

Those two changes did require some head scratching (does Hellaholic have hair?) _BUT_ altering "Quaternions" has been avoided.  That's gotta be worth something right there.

The CAD packages I'm familiar with (Inventor and SolidWorks) do not store "current rotation angles" anywhere.  The only way to get the current angle is to adjust the coordinate system the way you need it, and then to add a dimension to the workspace.  The angle is always subjective to the view rather than being absolute.
The reason for that is (drum roll please) ... _the Bounding Box doesn't rotate_ .  It simply stretches square to the World Coordinate System to contain the extents of the model whatever they happen to be.

HellAholic:
Well, turns out the issue happens even without the changes to the rotation xD
I'll add a separate bug ticket since I can reproduce the issue in 5.8.1 as well. So we're not breaking anything but as it is customary for any QA action (even when it's a CC PR) a bug spawns out of nothingness :P

GregValiant:
You need a model that is more fun.

https://github.com/user-attachments/assets/d9472b44-8328-4376-85e2-01262266c9e6

I think it just needs a tooltip change.
_"Any model rotation must be the last transformation or the model can be subject to random wonkiness."_

HellAholic:
Had an idea to get around the issue, but not sure how practical it would be. It involves Quaternions. So the outline is:

1. Get current model's Quaternion and store it (temp)
2. Invert the rotation (reset rotation)
3. Apply scale
4. re-apply rotation through the stored Quaternion or matrix

First need to get things working with Conan 2 xD

GregValiant:
_"Re-apply rotation through the stored Quaternion or matrix"_
What if the model is at a 2 or 3 angle compound rotation?

I'm catching up on my reviews.  It's been about 3 weeks now and I haven't had any problems with this at all.  The changes to the Translate Tool have likewise been fine.

HellAholic:
I'm trying to understand the scale behavior, but it doesn't seem to be as "Simple" as the rotation. The compound rotation is not the issue, at any given time, there is only one Quaternion that describes the model. It will get messy with the operations a bit since it adds a step but that wasn't the issue.
There are a lot of rounding and loss-prone calculations that are done during the scale, as far as I could figure out, so that makes it the same level of tricky as getting the value of rotation back (which I gave up on since it wasn't possible - infinite solutions) the scale goes  the other way and stores the scale factor in x, y, z, but when you rotate the model, those values are no longer valid.

So a very long story short, it's a big project to refactor the scale, and cannot be a simple adjustment by me (QA) :P
But since it's separate from this PR, I'll leave it be for now, if I have a spark of inspiration or suddenly develop an urge to code, I'll pick it up again.

GregValiant:
It would have to be a heck of an inspiration.

For now I'll stick with the thought that the bounding box has to rotate, and the Model Coordinate System must rotate with it.  That would keep the Model Coordinate System always at right angles to the faces of the bounding box.

Everything else leads to **_"Quaternion Madness"_**. which is un**Cura**ble.

(Yes, it's a horrible pun.  No, I will not apologize.)

GregValiant:
Because QT doesn't have both an "enabled" state and a "visible" state for controls, this is pretty clumsy.

I have noticed that the odd scaling behavior of a rotated model only occurs when "Uniform Scaling" is False.

- I rotate a model
- That sets a variable "has_been_rotated" to True.
  - When "has_been_rotated" is emitted - Uniform Scaling is set to True and both the checkbox and it's label are disabled.
  - That enables another checkbox and label which are dummies and are in the exact same location as the real ones.  The checkbox is locked on True and the label explains that due to the rotation, scaling can only be "Uniform".
- When I select the "Reset Rotation" button:
  - "has_been_rotated" becomes False
  - The dummy checkbox and label disappear and the real ones show up again allowing the user to de-select Uniform Scaling.

If it was possible to "lock" the Uniform Scaling checkbox and change the label text - the dummies wouldn't be needed.

This sort of thing is the reason I call myself the _King of Workarounds_.

plugins/Tools/RotateTool/RotateTool.qml
```diff
@@ -77,6 +106,57 @@ Item
...
+            text: "Snap Angle"
+            width: Math.ceil(contentWidth) //Make sure that the grid cells have an integer width.
+        }
+
+		UM.TextFieldWithUnit
+        {
+            id: angleTextField
+            width: UM.Theme.getSize("setting_control").width
+            height: UM.Theme.getSize("setting_control").height
+            unit: "degrees"
```
wawanbreton, line 133: Maybe use `°` ?

plugins/Tools/RotateTool/RotateTool.qml
```diff
@@ -77,6 +106,57 @@ Item
...
+
+            validator: UM.FloatValidator
+            {
+                maxBeforeDecimal: 3
+                maxAfterDecimal: 2
+            }
+            onEditingFinished:
+            {
+                var modified_text = text.replace(",", ".") // User convenience. We use dots for decimal values
+                if(text !="")
```
wawanbreton, line 144: I offer you a space
```suggestion
                if(text != "")
```

plugins/Tools/RotateTool/RotateTool.qml
```diff
@@ -77,6 +106,57 @@ Item
...
+                {
+                    UM.Controller.setProperty("RotationSnapAngle", modified_text)
+                }
+            }
+            onActiveFocusChanged:
+            {
+                if(!activeFocus && text =="")
```
wawanbreton, line 151: Sales on spaces
```suggestion
                if(!activeFocus && text == "")
```

plugins/Tools/RotateTool/RotateTool.qml
```diff
@@ -90,6 +170,144 @@ Item
...
+
+        UM.Label
+        {
+            height: UM.Theme.getSize("setting_control").height
+            text: "Z"
+            color: UM.Theme.getColor("y_axis"); // This is intentional. The internal axis are switched.
+            width: Math.ceil(contentWidth) //Make sure that the grid cells have an integer width.
+        }
+
+		UM.TextFieldWithUnit
```
wawanbreton, line 209: Wrong indentation ?

plugins/Tools/RotateTool/RotateTool.qml
```diff
@@ -90,6 +170,144 @@ Item
...
+                }
+            }
+            onActiveFocusChanged:
+            {
+                if(!activeFocus && text =="")
+                {
+                    xText = 0.1; // Need to change it to something else so we can force it to getvalue
+                    xText = 0
+                }
+            }
```
wawanbreton, line 238: The whole content of this item is duplicated a few times with slight differences, would it be easily possible to make a common element ?

wawanbreton (commented):
Looks goods ! Just a few remarks, mostly for consistency

GregValiant:
@HellAholic you are definitely better suited to handle the suggested changes.

plugins/Tools/RotateTool/RotateTool.qml
HellAholic: Tried a common element approach but the result was not as I expected, probably missing something in the implementation but only 1 field would show up.
image

GregValiant:
I still think that any rotation should lock the Scale Tool to "Uniform Scaling".  It should remain locked until "Reset all model transformations" is selected.

plugins/Tools/RotateTool/RotateTool.qml
wawanbreton: Ok, well there is probably a way, but if you at least tried, it is good enough :slightly_smiling_face:

wawanbreton:
> If someone were to ask - I would vote for the variable name changes and changing the "units" to " ° ". 

This is what @HellAholic did, and I also like it much more

> My files did not show the "incorrect indentations". Maybe I had fixed them. Would it even load if there were bad indentations?

This is a QML file, which is based on Javascript, not Python, so yes

> As for trying to make a common bit of code for the others; there are the 4 variables and 4 different textfields involved. At  first glance I didn't think it would be possible to squash them into a single bit of code.

It is always possible, but not always worth it indeed :slightly_smiling_face: in this case the pieces of code are grouped with each other, so it is acceptable IMHO.

wawanbreton (approved):

GregValiant:
I think this is as good as it can get and stay within making fairly simple changes.

I'm still not happy in regards to "non-uniform scaling of a rotated model" but it quickly becomes complicated and I don't see a fix.

- If an imported model is not "square to the world", there will always be issues with "Non-Uniform Scaling".
- If a model is square to the world, and is rotated by an exact multiple of 90°, then there usually isn't a problem with Non-Uniform Scaling.
- If there is a single model on the build plate, and the rotation is other than a multiple of 90°, then maybe the Scale Tool "uniformScalingCheckbox" could be set to "checked" and locked until the transformations are reset.
- If there are multiple models on the build plate then each model would seem to need a "has_been_rotated" property that would affect the "uniformScalingCheckbox" and the "NonUniformScaling" variable when that particular model is selected and the model has been rotated to an angle that is not a multiple of 90° and is square to the world.

That seems to get very messy very fast.

As an alternative, I am still in favor of adding a tooltip to the "uniformScalingCheckbox".
Something like:
"If a model has been rotated then 'Non-Uniform Scaling' might result in skewing of the model."

That is the tooltip I've added to this "ScaleTool.qml" file.
[ScaleTool.zip](https://github.com/user-attachments/files/18558920/ScaleTool.zip)

HellAholic:
Since this PR is reviewed I would keep it contained to the rotation. But I think your change is only a tooltip, in that case let's see :)
I will pay some cookie tributes to include it.
Need to fix the size change, and if I manage to do that then I'll include the reusable bit.

GregValiant:
Here's a dozen.  They're fresh.
​🍪​​🍪​​🍪​​🍪​​🍪​​🍪​​🍪​​🍪​​🍪​​🍪​​🍪​​🍪​

PR #979, opened 2024-11-26
Add rotation snap amount to UI

HellAholic:
Based on the idea of @GregValiant

image

A text field is added to the UI for snap angle
Using the setter and getter to update the behavior
Some modification to the setter and getter so the output/input can be consumed and stored as previous
The value is used to change the snap to angle while rotating the model
Works with all 3 rotation Axes


truegret:
I like it ..... !

GregValiant:
Simple, and easy to understand.
Nice job kid!

PR #980, opened 2024-11-26
Update TranslateToolHandle.py

GregValiant:
Flip the Y axis of the model coordinate system so that +Y is towards the back of the build plate.  This brings it in line with the build plate axis, and conventional coordinate systems.

This has been a pet peeve of mine for years.  Apparently it's not just me.  This goes back to at least 2017 and [#2152 ](https://github.com/Ultimaker/Cura/issues/2152).

North should be north.
image

### DIFF
--- a/plugins/Tools/TranslateTool/TranslateToolHandle.py
+++ b/plugins/Tools/TranslateTool/TranslateToolHandle.py
@@ -55,7 +55,7 @@ class TranslateToolHandle(ToolHandle):
                 width = self._line_width,
                 height = self._line_width,
                 depth = self._line_length,
-                center = Vector(0, 0, self._handle_position / 2),
+                center = Vector(0, 0, -(self._handle_position / 2)),
                 color = self._z_axis_color
             )
 
@@ -85,10 +85,10 @@ class TranslateToolHandle(ToolHandle):
                 width = self._handle_width,
                 height = self._handle_height,
                 depth = self._handle_width,
-                center = Vector(0, 0, self._handle_position),
+                center = Vector(0, 0, -(self._handle_position)),
                 color = self._z_axis_color,
                 axis = Vector.Unit_X,
-                angle = -90
+                angle = 90
             )
 
         self.setSolidMesh(mb.build())
@@ -117,7 +117,7 @@ class TranslateToolHandle(ToolHandle):
                 width = self._active_line_width,
                 height = self._active_line_width,
                 depth = self._active_line_length,
-                center = Vector(0, 0, self._active_handle_position / 2),
+                center = Vector(0, 0, -(self._active_handle_position / 2)),
                 color = self._z_axis_color
             )
 
@@ -142,7 +142,7 @@ class TranslateToolHandle(ToolHandle):
             width = self._active_handle_width,
             height = self._active_handle_width,
             depth = self._active_handle_width,
-            center = Vector(0, 0, self._active_handle_position),
+            center = Vector(0, 0, -(self._active_handle_position)),
             color = ToolHandle.ZAxisSelectionColor
         )
         self.setSelectionMesh(mb.build())


GregValiant:
@HellAholic would you be so kind as to take a look?  I haven't run into any problems with this and I've had it in place for a while.

GregValiant:
Go fishing or dancing or something.  This will all be here when you get back.

nallath:
I have changed the behavior a bit; It's now also changeable in a preference.

GregValiant:
@nallath 
I just looked at the "Print Head Polygon" dimensions and the "Y min" and "Y max" need to be flipped.  The view on the build plate is the same as it was before flipping the +Y.

This is with Y max = 70 and Y min = -5.  The big part should be to the rear of the build plate.
image

GregValiant:
Locally I have the "TranslateToolHandle.py" that was in the original PR.
I rapidly get out of my depth here, but I'll try to explain again.

If "Invert Y Axis" is enabled (what the PR does, and what it sounds like Jaime's switch makes an option) the defined polygon is inverted.  I will assume that it is an option and the user can turn it on or off.

The way the file loads the print head polygon from the definition file.
image
and it would display like this when "Invert Y axis"" is True.  The polygon is "mirrored about the X" in regards to the new Translate Tool Handle Y axis arrow direction.
image

These settings result in a correct display of the printhead polygon area
image
image

I think two things need to happen and I don't know how to do either.

1. When a new printer is added - the "Invert Y axis" option must be checked.  If it is "False" then all is good.  If it is "True", then the Ymax text box should be changed to the + value of the Ymin textbox, and the Ymin textbox should be changed to the negative value of the Ymax textbox.
Sort of like:
```
if invert_y_axis_setting:    
    ymIN_temp = abs(machineYMinField)
    ymAX_temp = -abs(machineYMaxField)
    machineYMinField = ymAX_temp
    machineYMaxField = ymIN_temp
```
2. When a user changes the "Invert Y axis" option, then similar code needs to run.
```
if invert_y_axis_setting CHANGES:    
    ymIN_temp = abs(machineYMinField)
    ymAX_temp = -abs(machineYMaxField)
    machineYMinField = ymAX_temp
    machineYMaxField = ymIN_temp
```

I think that is sufficient.
When a definition file loads then the box values get flipped if "Invert Y Axis" is true.
Anytime "Invert Y Axis" is changed, then the numbers get flipped.
The numbers get saved in the printer local definition, and Cura has to remember the current state of "Invert Y Axis".  As long as those two things happen it should all be good.  Project files from previous versions might be another story.

(As a side note, this took way to long to type even without mentioning Quaternion in it.)

GregValiant:
One more thing...I downloaded the files with the "Flip Y" checkbox in the preferences.  I would suggest changing the label to
"Flip tool handle Y axis **(restart required)**

GregValiant:
The Translate Tool text box for the Y is showing inverted values.  They are negative when they should be positive.  I made some progress on that.
In TranslateTool.py - I added an IF statement in "getY" (line 87) and that updates the box correctly when dragging the model with the tool.

    def getY(self) -> float:
        """Get the y-location of the selection bounding box center.

        :return: Y location in mm.
        """
        if Selection.hasSelection():
            # Note; The switching of z & y is intentional. We display z as up for the user,
            # But store the data in openGL space.
            if not bool(Application.getInstance().getPreferences().getValue("tool/flip_y_axis_tool_handle")):
                return float(Selection.getBoundingBox().center.z)
            else:
                return float(Selection.getBoundingBox().center.z) * -1
        return 0.0

A remaining problem is text input.  Typing a number into the box still results in backwards movement.  When the tool is flipped a negative input will still move the model towards the back of the build surface in the (now) positive Y.

GregValiant:
This is kind of weird.  With that "if" statement in place - by alternately clicking on the Y box and Z box, the +/- of the Y axis alternates and the model dances.  (Admittedly, not a very good dance.)
[Toggle Y location.zip](https://github.com/user-attachments/files/18036846/Toggle.Y.location.zip)

GregValiant:
I downloaded the raw TranslateTool.py file and applied your changes.
It looks good.  No jumping from + to - .  The number in the "Y" axis text box is correct per the "Y" orientation.
Changing back to the default "Y" axis it also looks good.  

I'm not sure if those were the only changes on my system.  I think I got all the files right, but I was jumping around working on this and a couple of other things (as thoughts occurred to me) and may have left some vestigial changes.

I checked MachineSettingsPrinterTab.qml and it appears to be the correct Cura file.  My changes to "Ymin" and "Ymax" "maximum and minimum values" did not appear to work correctly.  That might need a revisit because the way it looks: the "+/-" are incorrect in the Ymin and Ymax boxes when the Y is flipped.  Very minor thing as the polygon does display correctly.  I think if an adjustment is made there then there will be a cascade back through everything else.  It may be better to leave that one alone.  People should notice the inconsistency if they make manual changes to the values.

Before I forget...
Line 677 in GeneralPage.qml:
"Flip **translate** toolhandle Y axis **(restart required)**")

GregValiant:
That looks good.  I think it's a "set and forget" anyway.  Most users will only see it once and then acclimate to the change.

A favorite saying of mine is that "Everything affects everything" and this was another good example of that.

Do you have a list of all the files that have been changed for this?  I'd like to pull down the final versions and use them for a couple of weeks to see if anything else comes to light.  I'm a little sheepish over missing some of the effects of the "Y Axis" change.  It all seemed so simple.  (Avoiding the Quaternion thing was huge.)

GregValiant:
When the "Y" is flipped the PrintHead settings "Y min" and "Y max" can be confusing because the +/- doesn't doesn't change to the actual orientation of the tool handle.

In "MachineSettingsPrinterTab.qml" I would suggest changing:
line 217:
labelText: catalog.i18nc("@label", "Y min **( '-' towards back)**")
and line 257:
labelText: catalog.i18nc("@label", "Y max **( '+' towards front)**")

In fdmprinter.def.json.pot:
msgctxt "machine_head_with_fans_polygon description"
msgid "The shape of the print head. These are coordinates relative to the position of the print head, which is usually the position of its first extruder. The dimensions left and in front of the print head must be negative coordinates **(that is not affected by 'Flip 'Y' axis')**."
msgstr ""

Would those changes require changing all the "*.po" files?

HellAholic:
Any new string (label, description, etc) will require change to the *.po files. There is a process in place for updating translations for each release so couple more strings shouldn't be a problem.
So based on the Uranium/Cura/fdmprinter code (qml, python, ...) the strings inside the catalog.i18nc("String") are extracted into the *.po files then the files are sent for translations (either entirely or only the diff), and then you get the .po files for different languages.
I'm not sure the current description is correct, "dimensions left and in front of the print head must be negative"
image
The left one (x min) is correct but front of the print head is Y max which is currently positive. 🤔

GregValiant:
Here are "fdmprinter.def.json" with my proposed Tool Tip change for "machine_head_with_fans_polygon" and "MachineSettingsPrinterTab.qml" with the changes to the labels for "Y Min" and "Y Max".
They are on the Cura side but are relevant to this PR.

PR #987, opened 2025-02-19
CURA-12400 improve open source referencing

wawanbreton:
Add Qt as an extra dependency for which we want to display the license

CURA-12400

Requires https://github.com/Ultimaker/Cura/pull/20291

### DIFF
--- a/conandata.yml
+++ b/conandata.yml
@@ -3,10 +3,18 @@ version: "5.10.0-alpha.0"
 requirements:
   - "pyarcus/5.10.0@ultimaker/stable"
 
+extra_dependencies:
+  Qt:
+    version: "6.6.0"
+    sources_url: https://code.qt.io/cgit/qt/qtbase.git/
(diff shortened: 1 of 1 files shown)


conandata.yml
```diff
@@ -3,10 +3,19 @@ version: "5.10.0-alpha.0"
 requirements:
   - "pyarcus/5.10.0@ultimaker/stable"
 
+extra_dependencies:
+  Qt:
+    version: "6.6.0"
+    sources_url: https://code.qt.io/cgit/qt/qtbase.git
+    home_url: https://github.com/Ultimaker/Cura/blob/main/licenses_thirdparty/qt.md
```
rburema, line 10: I'm not sure you changed the right one here? Do we still want to link to the original qtbase repo in the `sources_url`?

rburema (approved):
I think it's fine, but please take a look at the 1 comment.

PR #989, opened 2025-03-12
CURA-12403 Fix wrongly switching project name

wawanbreton:
CURA-12403

:warning: This is technically an API change, but we already made the annoucement for the API change. Should we still update the wiki silently ?

Comes with https://github.com/Ultimaker/Cura/pull/20377


rburema:
Technically it's even a major change :-p

What I'd do is keep the old method, mark it as deprecated, and don't use it ourselves. Then we can remove it in the next major.

rburema (changes requested):
I can re-add the (deprecated) project myself as well, then after I'm done testing, send it to QA?

wawanbreton:
> I can re-add the (deprecated) project myself as well, then after I'm done testing, send it to QA?

That would make sense, yes please

PR #991, opened 2025-03-14
CURA-12309 hard crash when modifying a post processing setting value with auto slice

wawanbreton:
This could happen when an other Cura window was open, apparently the QML rendering then creates a different OpenGL context and we need to switch back to our own context before starting to draw

CURA-12309
Fixes https://github.com/Ultimaker/Cura/issues/19904


HellAholic (approved):
No more crashing

PR #992, opened 2025-03-16
Message ID is set to were but these translations still referenced old…

HellAholic:
CURA-12470

Typo fix in .po files
The reference was incorrect as the source string has been modified see:
https://github.com/Ultimaker/Uranium/blob/2ce596e1f186582754a09945efac1c0ce083dca2/UM/FileHandler/ReadFileJob.py#L91


wawanbreton (approved):
That is weird, I have no idea how the source text can have a typo, it is supposed to be automatically extracted from source code... Maybe some made a global replace at some point.

PR #995, opened 2025-04-28
CURA-12099 export and import to bambu 3mf

wawanbreton:
Updates the mesh writers to handle the `mime_type` argument

Comes with https://github.com/Ultimaker/Cura/pull/20518
CURA-12099

### DIFF
--- a/UM/Mesh/MeshWriter.py
+++ b/UM/Mesh/MeshWriter.py
@@ -12,7 +12,7 @@ class MeshWriter(FileWriter):
     def __init__(self, *args, **kwargs):
         super().__init__(*args, **kwargs)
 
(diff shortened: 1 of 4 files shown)


fieldOfView:
Would this require a major version increase? From a cursory look, all plugins implementing MeshWriter would need to be updated.

wawanbreton:
> Would this require a major version increase? From a cursory look, all plugins implementing MeshWriter would need to be updated.

That is true, yes. I have been trying to find a way to avoid it, but couldn't. That would be a version increase of the API though. I will see with the team whether this is acceptable as is. The addition of the `**kwargs` should be future-proof though, so that if we want to add new specific arguments later, there would be no API change.

PR #999, opened 2025-05-28
[CURA-12543] Backend support of Painting UI

rburema:
- Be able to update sub-images within the wrapper for OpenGL textures
- Add convenience method(s) such as for retrieving face-normals.
- Expose some things that previously weren't exposed, so we can render some very specific things we now need in the painting tool more efficiently -- the methods exposed don't require (much) knowledge of the internal workings, so it should be OK.

Frontend PR: https://github.com/Ultimaker/Cura/pull/20640

### DIFF
--- a/UM/Math/Plane.py
+++ b/UM/Math/Plane.py
@@ -22,7 +22,7 @@ class Plane:
     def distance(self):
         return self._distance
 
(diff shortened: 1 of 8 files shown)


wawanbreton (approved):
Very nice :slightly_smiling_face:

PR #1003, opened 2025-07-21
CURA-12528 unwrap uvs

wawanbreton:
Use libUvula to properly unwrap UV coordinates before we start painting on an object
Allow exporting UV coordinates into OBJ file

CURA-12528
Requires [libUvula](https://github.com/Ultimaker/libUvula)


UM/Mesh/MeshData.py
```diff
@@ -405,6 +409,24 @@ def invertNormals(self) -> None:
...
+        # remove_indices = False
+        if self._indices is None:
+            if self._vertices is None or self._vertex_count < 3:
+                return None
+            # remove_indices = True
+            indices = numpy.arange(self._vertex_count, dtype=numpy.int32).reshape(-1, 3)  # 3 verts per sub-array.
+        else:
+            indices = numpy.array()
+
+        self._uvs, texture_width, texture_height = uvula.unwrap(self._vertices, indices)
```
rburema, line 423: The `unwrap` function can throw
(see;
https://github.com/Ultimaker/libUvula/blob/4eb68c51af2fcaa10d4d70b8698b8fd31c729d93/pyUvula/pyUvula.cpp#L20
and;
https://github.com/Ultimaker/libUvula/blob/4eb68c51af2fcaa10d4d70b8698b8fd31c729d93/pyUvula/pyUvula.cpp#L38
).

Handling that would already pull one thing off of QA's list (the defensive programming thing) 😉

rburema (changes requested):
Looks good, some small remarks.

PR #1004, opened 2025-07-25
Cura 11810 fix mac dialogs

wawanbreton:
The Marketplace and the Preferences dialog are having display issues in Mac environment. In order to fix this, I have applied some sanity changes:
* Set the proper `transientParent` to dialogs created by Python code
* Set them as modal so that they sould always appear in front of the main window
* Instanciate them only when required, then delete them as soon as we don't need them, then create a new one if we need to re-display

As a bonus, I also removed some unused QML components and simplified the creation logic of the preferences dialog.

CURA-11810
Comes with https://github.com/Ultimaker/Cura/pull/20797

### DIFF
--- a/UM/Qt/QtApplication.py
+++ b/UM/Qt/QtApplication.py
@@ -605,10 +605,11 @@ class QtApplication(QApplication, Application):
 
     def createQmlSubWindow(self, qml_file_path: str, context_properties: Dict[str, "QObject"] = None) -> Optional["QQuickWindow"]:
         """
-        Create a QML window from a QML file. This is different from createQmlComponent in 3 different ways:
+        Create a QML window from a QML file. This method uses createQmlComponent internally, but adds a few specific
+        features for windows management:
             * The created object is a QQuickWindow instance
             * The transient parent of the window is explicitly set to be the main window
-            * The window ownership is handled and it will be destroyed as soon as it is no more visible,
+            * The window ownership is handled and it will be destroyed as soon as it's no longer visible,
               so the caller should not keep a reference to it
         :param qml_file_path: The absolute file path to the root qml file.
         :param context_properties: Optional dictionary containing the properties that will be set on the context of the
--- a/UM/Qt/qml/UM/Dialog.qml
+++ b/UM/Qt/qml/UM/Dialog.qml
@@ -103,7 +103,7 @@ Window
     {
         if(selfDestroy && !visible)
         {
-            destroy()
+            destroy();
         }
     }
 


UM/Qt/QtApplication.py
```diff
@@ -598,10 +599,38 @@ def closeSplash(self) -> None:
...
+    def _onWindowVisibleChange(self, visible: bool) -> None:
+        if not visible:
+            self._sub_windows.remove(self.sender())
+
+    def createQmlSubWindow(self, qml_file_path: str, context_properties: Dict[str, "QObject"] = None) -> Optional["QQuickWindow"]:
+        """
+        Create a QML window from a QML file. This is different from createQmlComponent in 3 different ways:
+            * The created object is a QQuickWindow instance
+            * The transient parent of the window is explicitly set to be the main window
+            * The window ownership is handled and it will be destroyed as soon as it is no more visible,
```
rburema, line 611: >  as soon as it is no more visible,

Should probably be: 'as soon as it's no longer visible,'

UM/Qt/QtApplication.py
```diff
@@ -598,10 +599,38 @@ def closeSplash(self) -> None:
             QtApplication.splash.close()
             QtApplication.splash = None
 
+    def _onWindowVisibleChange(self, visible: bool) -> None:
+        if not visible:
+            self._sub_windows.remove(self.sender())
+
+    def createQmlSubWindow(self, qml_file_path: str, context_properties: Dict[str, "QObject"] = None) -> Optional["QQuickWindow"]:
+        """
+        Create a QML window from a QML file. This is different from createQmlComponent in 3 different ways:
```
rburema, line 608: Since you're still using that method _within_ this one, it's a bit confusing (though technically correct) to say that it's 'different' -- maybe just state that this is a wrapper that handles some additional tasks instead.

UM/Qt/qml/UM/Dialog.qml
```diff
@@ -97,6 +99,14 @@ Window
         base.visible = true;
     }
 
+    onVisibleChanged:
+    {
+        if(selfDestroy && !visible)
+        {
+            destroy()
```
rburema, line 106: We're not too consistent with it, but since this is more than a one-liner and thus 'in' javascript, we should add the `;` after `destroy()` maybe.

wawanbreton:
> I just have the smallest, least consequential review comments ever 😅

I thought you were exaggerating, but it turns out you are right :laughing:

UM/Qt/qml/UM/Dialog.qml
wawanbreton: Right, I do prefer to add it but since our QML part comes with Python code, I often forget :smile:

PR #1005, opened 2025-07-29
CURA-12016 improve loading time with many printers

wawanbreton:
This PR provides 2 changes:
* Disable the garbage collector while loading cached containers from pickled files. This makes the loading twice as fast, which is significant because we do it multiple times with large containers
* Set the proper container sub-type in the metadata so that we can properly filter on sub-types (necessary for https://github.com/Ultimaker/Cura/pull/20813)

CURA-12016


rburema (approved):
I wonder if we can (temporarily) disable the gc 'more generally' than just that piece of code -- there must be other tasks where that's in the way.

PR #1007, opened 2025-08-04
Cura 12660 painting UI improvements

wawanbreton:
The main change in this PR concerns the way we use offscreen rendering passes. Previously they were just called when needed, but that caused some troubles e.g. when a QML message box was displayed at the same time. Since we still have not found a proper fix for this, I have now changed the behavior to enabling/disabling the render passes, but have them rendered always in the main Qt rendering loop, which always works. So now when a tool requires e.g. a selection-by-faces rendering pass, it just activates it and gets the output when needed.
As the bug was also present for rotation-by-face-selection and support-blocker-placement, I also updated those tools so that they use the new method and don't suffer this bug anymore.
And finally, since the rotation-by-face-selection tool was the only one using the face selection mechanism, and now it doesn't, I deprecated the whole behavior (which was actually partially implemented).

CURA-12660


rburema (changes requested):
Nothing major -- I'm going to lightly change a few things w.r.t. the API/SDK backwards compatibility for plugins though. (Since you're on holiday.)

PR #1024, opened 2026-02-11
Change event propagation order

wawanbreton:
The active tool is now the first to receive and possibly catch the event. It does seem quite logical to have it before the selection tool because the active tool is the one that should be actually interacting with the user, and the selection is somehow a fallback behavior. This change is required to allow painting on a model and using the shift key without having the selection tool to catch this event and change the selection.

:warning: Although I tested it thoroughly and saw no issues, there is a risk that some tools-related behavior have changed or are broken, especially for tools provided by external plugins.

Required by https://github.com/Ultimaker/Cura/pull/21393


rburema (approved):
Code seems OK, so I tried just these changes (not the painting ones) with a 3rd party plugin that is a tool (the banana split one), and that seems to work without trouble.

wawanbreton:
> Code seems OK, so I tried just these changes (not the painting ones) with a 3rd party plugin that is a tool (the banana split one), and that seems to work without trouble.

Cool, that is good news :smiley:

PR #1039, opened 2026-04-28
Only send the required inputs

HellAholic:
Only pass the required data rather than the entire run

Requires:
https://github.com/Ultimaker/cura-workflows/pull/53

[CURA-13126](https://ultimaker.atlassian.net/browse/CURA-13126)

[CURA-13126]: https://ultimaker.atlassian.net/browse/CURA-13126?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/tests/benchmarks/conftest.py
+++ b/tests/benchmarks/conftest.py
@@ -7,7 +7,7 @@ import warnings
 warn = True
 
 @pytest.hookimpl
-def pytest_ignore_collect(path, config):
+def pytest_ignore_collect(collection_path, config):
     if config.pluginmanager.hasplugin("pytest-benchmark"):
         return False
     else:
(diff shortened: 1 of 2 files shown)


tests/benchmarks/conftest.py
```diff
@@ -7,7 +7,7 @@
 warn = True
 
 @pytest.hookimpl
-def pytest_ignore_collect(path, config):
+def pytest_ignore_collect(collection_path, config):
```
wawanbreton, line 10: Argument name has changed but nothing else ? Is this argument unused ?

wawanbreton (approved):

tests/benchmarks/conftest.py
HellAholic: It was a deprecation change, `py.path.local arguments for hooks replaced with pathlib.Path` 
according to the documentation everything else is mostly the same, we just need to use collection_path instead of path.
Reference 👇 
https://docs.pytest.org/en/latest/deprecations.html#py-path-local-arguments-for-hooks-replaced-with-pathlib-path

PR #1044, opened 2026-06-17
[CURA-13191] Fix mesh-data related issues.

rburema:
- getFaceCount needs to return an integer, this wouldn't happen (whenever it hasn't any indices), and then the 'float'ness would propagate so that indexing wouldn't work in some cases
- the face normal returned wasn't actually well, normalized (now that it is, we can also remove the spurious normalization in the one place we where already using it)

done as part of CURA-13191

Backend thing for: https://github.com/Ultimaker/Cura/pull/21664


wawanbreton (approved):
Thank you for teaching me the `//` operator 😄

PR #1046, opened 2026-08-19
CURA-13243 center dialogs on linux

wawanbreton:
Since the update to Qt 6.10, dialogs don't appear centered on Linux environment. Since this should really happen natively, I found no other way but to actually set the window position manually.

An other solution would be to have `UM.Dialog` use a `QtQuick.Controls.Dialog` instead of a `QtQuick.Controls.Window`, which fixes the centering, but that leads to other issues such at not being able to set a minimum size.

CURA-13243


rburema (approved):
LGTM!

PR #1047, opened 2026-08-24
CURA-9839 Use updated pyArcus with support of repeated properties

wawanbreton:
Contributes to https://github.com/Ultimaker/Cura/pull/21766
CURA-9839


rburema (approved):
LGTM.

PR #1048, opened 2026-09-01
Cura 13267 file menu rework

wawanbreton:
This PR simplifies the logic behind the "File" menu. Since the "Save project" and "Load project" menus are dynamic depending on the loaded plugins and their enabled states, it is quite difficult to make it work reliably on all platforms.
* Change the `FileProvider` and `ProjectOutputDevice` classes so that their properties can be used directly from within QML
* Change the `FileProviderModel` and `ProjectOutputDevicesModel` classes so that they directly handle a list of objects to the QML part
* Change the `OutputDeviceManager` so that we can just add/remove the output devices instead of re-creating the full model every time something changes

CURA-13267

### DIFF
--- a/UM/FileProvider.py
+++ b/UM/FileProvider.py
@@ -1,7 +1,7 @@
 # Copyright (c) 2022 Ultimaker B.V.
 # Uranium is released under the terms of the LGPLv3 or higher.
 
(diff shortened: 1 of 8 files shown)


rburema (commented):
Skimmed it several times (since already merged), but looks good to me.

